PR #3, opened 2022-03-16
Use GlobalLock to prevent multiple prompts.

goagain:
To avoid multiple prompts, a named mutex is used, which can prevent in both single process and multiple processes.

Testcase:
Run the following command parallel but in one terminal server sessions.
`dotnet run --project ./src/AzureAuth --debug --resource={resource_id} --client={client_id} --tenant={tenant_id} -m web`

See [Mutex: Local vs Global](https://docs.microsoft.com/en-us/dotnet/api/system.threading.mutex?view=net-6.0#remarks)


src/MSALWrapper.Test/GlobalLockTest.cs
```diff
@@ -0,0 +1,120 @@
...
+            Assert.AreEqual(1, sum);
+        }
+
+        /// <summary>
+        /// Test mutex in tasks.
+        /// </summary>
+        [Test]
+        public void TestMutexInTasks()
+        {
+            int cas = 0;
```
kyle-rader, line 60: It's not clear to me on first sight, what the variable name `cas` means. Consider a more description name.

src/MSALWrapper.Test/GlobalLockTest.cs
```diff
@@ -0,0 +1,120 @@
...
+            int sum = 0;
+            var tasks = new List<Task>();
+            for (int i = 0; i < NumberOfThreads; i++)
+            {
+                var task = new Task(() =>
+                {
+                    using (new GlobalLock(this.lockName))
+                    {
+                        if (Interlocked.CompareExchange(ref cas, 1, 0) == 1)
```
kyle-rader, line 69: This is a pretty non-standard API to be using (unless you are someone who is used to writing parallel or concurrent code - which generally most people are not). So I think this is a good opportunity for a comment to help describe what is being checked here, and also potentially at the start of the method to describe the overall goal of this test. What is this test asserting, and how is it doing so?

Does this test fail without the logic in GlobalLock?

src/MSALWrapper.Test/GlobalLockTest.cs
```diff
@@ -0,0 +1,120 @@
...
+                        {
+                            Assert.Fail($"The thread should be blocked by {nameof(Thread)}");
+                        }
+
+                        sum++;
+                        Thread.Sleep(10);
```
kyle-rader, line 75: Is sleeping here needed for the test to make it's assertions? We generally want to avoid slowing down the tests if we can.

src/MSALWrapper.Test/GlobalLockTest.cs
```diff
@@ -0,0 +1,120 @@
...
+                {
+                    using (new GlobalLock(this.lockName))
+                    {
+                        if (Interlocked.CompareExchange(ref cas, 1, 0) == 1)
+                        {
+                            Assert.Fail($"The thread should be blocked by {nameof(Thread)}");
+                        }
+
+                        sum++;
+                        Thread.Sleep(50);
```
kyle-rader, line 108: Same question here on if we really do need to sleep, and if so, do we need to sleep this much?

src/MSALWrapper.Test/GlobalLockTest.cs
```diff
@@ -0,0 +1,120 @@
...
+                        Thread.Sleep(10);
+                        Interlocked.Exchange(ref cas, 0);
+                    }
+                });
+                task.Start();
+                tasks.Add(task);
+            }
+
+            Task.WaitAll(tasks.ToArray());
+            Assert.AreEqual(NumberOfThreads, sum);
```
kyle-rader, line 84: To be consistent with the existing tests let's try to use Fluent Assertions whenever possible. Fluent Assertions lets you write this assertion as
```csharp
sum.Should().Be(NumberOfThread);
```
Why use Fluent Assertions?
Fluent Assertions will nearly always give you better error messages when a test does fail, and for more complicated assertions gives you a a very powerful API to make more complicated assertions in a simple, readable way.

src/MSALWrapper.Test/GlobalLockTest.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        /// <summary>
+        /// Test mutex in threads.
+        /// </summary>
+        [Test]
+        public void TestMutexInThreads()
+        {
+            int cas = 0;
```
kyle-rader, line 93: Same comment on `cas`.

src/MSALWrapper/GlobalLock.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// <param name="lockName">Unique lock ID.</param>
+        /// <exception cref="ArgumentNullException">The <see cref="ArgumentNullException"/>.</exception>
+        public GlobalLock(string lockName)
+        {
+            if (lockName == null)
+            {
+                throw new ArgumentNullException($"{nameof(lockName)} must not be null");
+            }
+
+            this.mutex = new Mutex(false, lockName);
```
kyle-rader, line 34: Similar to the notion of "no magic numbers", I also like to encourage no "magic bools". For anyone reading `new Mutex(false, lockName);` it is not clear what the value of `false` is. Prefer extracting this into a named `const` value.

src/MSALWrapper/GlobalLock.cs
```diff
@@ -0,0 +1,80 @@
...
+            try
+            {
+                this.mutex.WaitOne();
```
kyle-rader, line 37: Again, here most people aren't familiar with the Mutex API, so a short comment here on why we are calling `WaitOne()` would be nice.

kyle-rader (changes requested):
Overall I _think_ this is looking good. The code is fairly clean, and StyleCop has forced a particular style! (Though I think we can disable the required XML comments for the tests - I don't find those useful there).
 I've left several comments and have a few general requests:
- [ ] Flesh out the pull request description with a little more detail as to Why this change is being made. The PR is public record of why and how this behavior was added.
- [ ] Detail in the PR description how this works across Windows, Mac, and Linux.
- [ ] Does it also behave the same in both netframework472, netstandard2.0, and netstandard2.1?
- [ ] I would like to get at least one more review on this from someone with more C# expertise than I just because concurrency can be a very fickle area of programming :).

src/MSALWrapper.Test/GlobalLockTest.cs
goagain: used semaphore instead

src/MSALWrapper.Test/GlobalLockTest.cs
goagain: Changed to sleep(1) which should be enough.

src/AzureAuth/CommandMain.cs
```diff
@@ -255,11 +255,21 @@ private int GetToken()
...
             {
-                ITokenFetcher tokenFetcher = this.TokenFetcher();
-                TokenResult tokenResult = (this.tokenFetcherOptions.Scopes == null
+                TokenResult tokenResult = null;
+                ITokenFetcher tokenFetcher = null;
+
+                // When running multiple AzureAuth processes with the same tenant and resource ID,
+                // They may prompt many times, which is annoying and unexpected.
+                // Use GlobalLock to ensure that only one process can access the corresponding resource at the same time.
+                string lockName = "Global\\" + this.Tenant + this.Resource;
```
reillysiemens, line 264: I thought the lock needed to be around the `Resource`, `Client`, and `Tenant` trio. Is that right, @kyle-rader?

src/AzureAuth/CommandMain.cs
```diff
@@ -255,11 +255,21 @@ private int GetToken()
...
+                using (new GlobalLock(lockName))
+                {
+                    tokenFetcher = this.TokenFetcher();
```
reillysiemens, line 267: Could this still be instantiated outside of the critical section?

src/MSALWrapper/GlobalLock.cs
reillysiemens: I agree with @kyle-rader here. Additionally, there's no timeout associated with this, so we might end up waiting forever. I'm not sure that's what we want to do.

src/MSALWrapper/GlobalLock.cs
```diff
@@ -0,0 +1,80 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Dispose method.
+        /// </summary>
+        /// <param name="disposing">disposing.</param>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!this.disposedValue)
```
reillysiemens, line 60: I _think_ I understand what `disposing` is for, but what is `disposedValue` for?

reillysiemens (changes requested):
I've got a couple of questions about the approach here. I also wonder if it wouldn't be possible to use `Mutex` directly rather than to wrap it with the `GlobalLock` class. Would love to get your thoughts on that option.

src/AzureAuth/CommandMain.cs
kyle-rader: Yep, great catch - we only want to synchronize calls that will benefit from one another's caching. Which is the whole trio.

src/MSALWrapper.Test/GlobalLockTest.cs
```diff
@@ -0,0 +1,120 @@
...
+            int sum = 0;
+            var threads = new List<Thread>();
+            for (int i = 0; i < 10; i++)
```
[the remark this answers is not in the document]
goagain: Good catch. I forgot to change this code block. Also, you are right, 2 is enough here

src/MSALWrapper.Test/GlobalLockTest.cs
```diff
@@ -0,0 +1,120 @@
...
+
+    /// <summary>
+    /// The <see cref="GlobalLockTest"/> test.
+    /// </summary>
+    internal class GlobalLockTest
+    {
+        /// <summary>
+        /// The number of threads or tasks.
+        /// </summary>
+        private const int NumberOfThreads = 10;
```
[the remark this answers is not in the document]
goagain: you are right, 2 is enough here

src/AzureAuth/CommandMain.cs
```diff
@@ -255,11 +255,21 @@ private int GetToken()
...
-                ITokenFetcher tokenFetcher = this.TokenFetcher();
-                TokenResult tokenResult = (this.tokenFetcherOptions.Scopes == null
+                TokenResult tokenResult = null;
+                ITokenFetcher tokenFetcher = null;
+
+                // When running multiple AzureAuth processes with the same tenant and resource ID,
+                // They may prompt many times, which is annoying and unexpected.
+                // Use GlobalLock to ensure that only one process can access the corresponding resource at the same time.
+                string lockName = "Global\\" + this.Tenant + this.Resource;
+                using (new GlobalLock(lockName))
```
kyle-rader, line 265: After discussing with Tucker, I think I agree with the overall sentiment that `Mutex` provides what we need directly, and we don't have to test it's behavior. We can change this to use Mutex directly, and call it's Wait method - also using the `WaitOne(TimeSpan)` method to also enable a Wait Timeout here as a guard against unlimited blocking.

```suggestion
                using (var authLock = new Mutex(lockName))

```

kyle-rader: Just before we wait for the lock - we also want to add a `this.logger.LogDebug` call to signal a wait is about to happen to help debug scenarios where the auth cli might look like it is hanging. We could include the lock name in that debug message as well.

kyle-rader (changes requested):

PR #5, opened 2022-03-18
Enable AZUREAUTH_CONFIG env var fallback if not given as --config

kyle-rader:
# Problem
We currently provide a `--config` option to load `Alias`s from a config file. In order to use `--alias` though you must always add a `--config`. This is okay for scripts with known paths, but is still very cumbersome when a developer wants to explicitly warm a particular auth path.

# Solution
Provide `AZUREAUTH_CONFIG` as an env var to check for a config file path if none is given on the CLI. This will enable developer environment setup scripts to set a known config file location without imposing any directory or naming structure on them.


reillysiemens (approved):
This looks great, @kyle-rader! 🚀🚢

shalinikhare27 (approved):

PR #6, opened 2022-03-18
Use Mutex to prevent multiple prompts

goagain:
# Background
Some users encounter that when accessing a resource at the same time, if the token has not been cached, many authentication dialog boxes will pop up, which is quite annoying.

Ref: #3 may contains more context.
# Solution
To avoid multiple prompts, a named mutex is used, which can prevent in both single process and multiple processes.
See [Mutex: Local vs Global](https://docs.microsoft.com/en-us/dotnet/api/system.threading.mutex?view=net-6.0#remarks)

# Testcase
Run the following command parallel but in one terminal server sessions.
dotnet run --project ./src/AzureAuth --debug --resource={resource_id} --client={client_id} --tenant={tenant_id} -m web

After discussing with Kyle and Tucker, we can simplify by literally using Mutex. Since unit test on parallel scenario is difficult, we can test manually.

### DIFF
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -264,13 +264,11 @@ Allowed values: [all, web, devicecode]";
                 ITokenFetcher tokenFetcher = this.TokenFetcher();
                 TokenResult tokenResult = null;
 
-                // When running multiple AzureAuth processes with the same tenant and resource ID,
+                // When running multiple AzureAuth processes with the same resource, client, and tenant IDs,
+
                 // They may prompt many times, which is annoying and unexpected.
                 // Use Mutex to ensure that only one process can access the corresponding resource at the same time.
-                string lockName = "Global\\"
-                    + this.Resource
-                    + this.Client
-                    + this.Tenant;
+                string lockName = $"Global\\{this.Resource}_{this.Client}_{this.Tenant}";
 
                 // First parameter InitiallyOwned indicated whether this lock is owned by current thread.
                 // It should be false otherwise a dead lock could occur.
@@ -282,11 +280,12 @@ Allowed values: [all, web, devicecode]";
                         mutex.WaitOne();
                     }
 
-                    // AbandonedMutexException could be thrown if another process exit without releasing mutex correctly.
+                    // An AbandonedMutexException could be thrown if another process exits without releasing the mutex correctly.
+
                     catch (AbandonedMutexException)
                     {
                         // lock eventually acquired
-                        this.logger.LogWarning("Another auth thread or process may exit unexpected.");
+                        this.logger.LogWarning("The authentication attempt mutex was abandoned. Another thread or process may have exited unexpectedly.");
                     }
 
                     try
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -270,7 +270,6 @@ Allowed values: [all, web, devicecode]";
                 TokenResult tokenResult = null;
 
                 // When running multiple AzureAuth processes with the same resource, client, and tenant IDs,
-
                 // They may prompt many times, which is annoying and unexpected.
                 // Use Mutex to ensure that only one process can access the corresponding resource at the same time.
                 string lockName = $"Local\\{this.Resource}_{this.Client}_{this.Tenant}";
@@ -283,7 +282,7 @@ Allowed values: [all, web, devicecode]";
                     try
                     {
                         // Wait for the other session to exit.
-                        lockAcquired = mutex.WaitOne(this.mutexTimeout);
+                        lockAcquired = mutex.WaitOne(this.promptMutexTimeout);
                     }
 
                     // An AbandonedMutexException could be thrown if another process exits without releasing the mutex correctly.
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -283,7 +283,7 @@ Allowed values: [all, web, devicecode]";
                     try
                     {
                         // Wait for the other session to exit.
-                        mutex.WaitOne();
+                        lockAcquired = mutex.WaitOne();
                     }
 
                     // An AbandonedMutexException could be thrown if another process exits without releasing the mutex correctly.
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Authentication.AzureAuth
     using System.IO.Abstractions;
     using System.Linq;
     using System.Runtime.CompilerServices;
-
+    using System.Threading;
     using McMaster.Extensions.CommandLineUtils;
 
     using Microsoft.Authentication.MSALWrapper;
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -283,7 +283,8 @@ Allowed values: [all, web, devicecode]";
                     // An AbandonedMutexException could be thrown if another process exits without releasing the mutex correctly.
                     catch (AbandonedMutexException)
                     {
-                        // lock eventually acquired
+                        // If another process crashes or exits accidently, we can still acquire the lock.
+                        // In this case, basicly we can just leave a log warning, because the worst side effect is propmting more than once.
                         this.logger.LogWarning("The authentication attempt mutex was abandoned. Another thread or process may have exited unexpectedly.");
                     }
 
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -298,8 +298,7 @@ Allowed values: [all, web, devicecode]";
 
                     if (!lockAcquired)
                     {
-                        this.logger.LogError("Authentication failed. The application did not gain access in the expected time, possibly because the resource handler was occupied by another process for a long time.");
-                        return 1;
+                        throw new TimeoutException("Authentication failed. The application did not gain access in the expected time, possibly because the resource handler was occupied by another process for a long time.");
                     }
 
                     try
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -53,9 +53,9 @@ Allowed values: [all, web, devicecode]";
         private ITokenFetcher tokenFetcher;
 
         /// <summary>
-        /// MutexTimeout is the timeout of Mutex to prevent infinite waiting.
+        /// The maximum time we will wait to acquire a mutex around prompting the user.
         /// </summary>
-        private TimeSpan mutexTimeout = TimeSpan.FromMinutes(15);
+        private TimeSpan promptMutexTimeout = TimeSpan.FromMinutes(15);
 
         /// <summary>
         /// Initializes a new instance of the <see cref="CommandMain"/> class.
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -281,7 +281,6 @@ Allowed values: [all, web, devicecode]";
                     }
 
                     // An AbandonedMutexException could be thrown if another process exits without releasing the mutex correctly.
-
                     catch (AbandonedMutexException)
                     {
                         // lock eventually acquired
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -7,7 +7,7 @@ namespace Microsoft.Authentication.AzureAuth
     using System.Collections.Generic;
     using System.IO.Abstractions;
     using System.Linq;
-
+    using System.Threading;
     using McMaster.Extensions.CommandLineUtils;
 
     using Microsoft.Authentication.MSALWrapper;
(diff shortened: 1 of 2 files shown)


src/AzureAuth/CommandMain.cs
```diff
@@ -262,10 +262,45 @@ private int GetToken()
             try
             {
                 ITokenFetcher tokenFetcher = this.TokenFetcher();
-                TokenResult tokenResult = (this.tokenFetcherOptions.Scopes == null
-                    ? tokenFetcher.GetAccessTokenAsync(this.CombinedAuthMode)
-                    : tokenFetcher.GetAccessTokenAsync(this.tokenFetcherOptions.Scopes, this.CombinedAuthMode))
-                    .Result;
+                TokenResult tokenResult = null;
+
+                // When running multiple AzureAuth processes with the same tenant and resource ID,
```
reillysiemens, line 267: Don't forget to mention the `this.Client` as well.
```suggestion
                // When running multiple AzureAuth processes with the same resource, client, and tenant IDs,

```

src/AzureAuth/CommandMain.cs
```diff
@@ -262,10 +262,45 @@ private int GetToken()
...
-                    .Result;
+                TokenResult tokenResult = null;
+
+                // When running multiple AzureAuth processes with the same tenant and resource ID,
+                // They may prompt many times, which is annoying and unexpected.
+                // Use Mutex to ensure that only one process can access the corresponding resource at the same time.
+                string lockName = "Global\\"
+                    + this.Resource
+                    + this.Client
+                    + this.Tenant;
```
reillysiemens, line 273: This is just my opinion, but it might be a little cleaner to use string interpolation here. Also, consider using an underscore to separate each of these IDs in case the lock name needs to be read for debugging later.
```suggestion
                string lockName = $"Global\\{this.Resource}_{this.Client}_{this.Tenant}";
```

reillysiemens: Yet another consideration would be to hash (maybe with a SHA algorithm?) all three values so that they aren't accidentally leaked anywhere else in the OS (I don't actually know what C# does with the lock name). I'm not sure if that's necessary as I don't know if this trio is really considered "secret", but it might be worth thinking about.

src/AzureAuth/CommandMain.cs
```diff
@@ -262,10 +262,45 @@ private int GetToken()
...
+                    {
+                        // Wait for other session exit.
+                        mutex.WaitOne();
+                    }
+
+                    // AbandonedMutexException could be thrown if another process exit without releasing mutex correctly.
+                    catch (AbandonedMutexException)
+                    {
+                        // lock eventually acquired
+                        this.logger.LogWarning("Another auth thread or process may exit unexpected.");
```
reillysiemens, line 289: I have a small grammatical suggestion here and I think just a little extra context would be useful as well.
```suggestion
                        this.logger.LogWarning("The authentication attempt mutex was abandoned. Another thread or process may have exited unexpectedly.");
```

src/AzureAuth/CommandMain.cs
```diff
@@ -262,10 +262,45 @@ private int GetToken()
...
+                // It should be false otherwise a dead lock could occur.
+                using (Mutex mutex = new Mutex(false, lockName))
+                {
+                    try
+                    {
+                        // Wait for other session exit.
+                        mutex.WaitOne();
+                    }
+
+                    // AbandonedMutexException could be thrown if another process exit without releasing mutex correctly.
```
reillysiemens, line 285: There's a small grammatical correction that can be made here.
```suggestion
                    // An AbandonedMutexException could be thrown if another process exits without releasing the mutex correctly.

```

reillysiemens (commented):
@goagain: This looks like a much more concise approach than #3! Really quick turnaround too after our last discussion.

I think it will be good to have other team members review this as well, so everyone understands this piece of the code, but I would be happy to approve this after you make some minor tweaks. 😀

goagain:
> @goagain: This looks like a much more concise approach than #3! Really quick turnaround too after our last discussion.
> 
> I think it will be good to have other team members review this as well, so everyone understands this piece of the code, but I would be happy to approve this after you make some minor tweaks. 😀

Thank you. So many grammatical error : (

kyle-rader:
> > @goagain: This looks like a much more concise approach than #3! Really quick turnaround too after our last discussion.
> > I think it will be good to have other team members review this as well, so everyone understands this piece of the code, but I would be happy to approve this after you make some minor tweaks. 😀
> 
> Thank you. So many grammatical error : (

It's okay! English is a difficult language to begin with and we all make mistakes here. We're all here to help one another learn and improve through practice :)

reillysiemens:
@goagain: Honestly, I didn't think there were that many grammatical errors. I promise you I'd have more in any language that wasn't English. 😆

kyle-rader (approved):

src/AzureAuth/CommandMain.cs
```diff
@@ -262,10 +262,43 @@ private int GetToken()
...
+                    try
+                    {
+                        // Wait for other session exit.
+                        mutex.WaitOne();
+                    }
+
+                    // An AbandonedMutexException could be thrown if another process exits without releasing the mutex correctly.
+                    catch (AbandonedMutexException)
+                    {
+                        // lock eventually acquired
```
reillysiemens, line 286: I think this comment could use just a little more clarification, but other than that things look good to me.

reillysiemens (approved):

src/AzureAuth/CommandMain.cs
goagain: I have the comment in this code block more detailed. Is that work for you?

src/AzureAuth/CommandMain.cs
```diff
@@ -262,10 +262,44 @@ private int GetToken()
...
-                    ? tokenFetcher.GetAccessTokenAsync(this.CombinedAuthMode)
-                    : tokenFetcher.GetAccessTokenAsync(this.tokenFetcherOptions.Scopes, this.CombinedAuthMode))
-                    .Result;
+                TokenResult tokenResult = null;
+
+                // When running multiple AzureAuth processes with the same resource, client, and tenant IDs,
+
+                // They may prompt many times, which is annoying and unexpected.
+                // Use Mutex to ensure that only one process can access the corresponding resource at the same time.
+                string lockName = $"Global\\{this.Resource}_{this.Client}_{this.Tenant}";
```
mjcheetham, line 271: According to the `Mutex` documentation, using the `Global\` (or `Local\`) prefix for the name only applies to running on a Terminal Services server. From the [Remarks section of the docs](https://docs.microsoft.com/en-us/dotnet/api/system.threading.mutex?view=net-6.0#remarks):

> Note: On a server that is running Terminal Services, a named system mutex can have two levels of visibility. If its name begins with the prefix "Global\\", the mutex is visible in all terminal server sessions. If its name begins with the prefix "Local\\", the mutex is visible only in the terminal server session where it was created. In that case, a separate mutex with the same name can exist in each of the other terminal server sessions on the server. If you do not specify a prefix when you create a named mutex, it takes the prefix "Local\\". Within a terminal server session, two mutexes whose names differ only by their prefixes are separate mutexes, and both are visible to all processes in the terminal server session. That is, the prefix names "Global\\" and "Local\\" describe the scope of the mutex name relative to terminal server sessions, not relative to processes.

Not sure if you intended to restrict to only one simultaneous global auth prompt across multiple concurrent users.

mjcheetham (commented):
Hello! 👋 I work on the Git Credential Manager team that also uses MSAL in a CLI tool. I'm excited to see the `authcli` hit public release! 🎉 😁 

I hate to jump in uninvited to comment here, but I had too considered adding some mutex in GCM to prevent multiple auth prompts in the past. I never actually got around to doing so, but I remember in my exploration reading about the multiple-user one-machine scenarios as commented below. Things may have changed in the implementation since then (over a year ago) so feel free to ignore me of course 😅

src/AzureAuth/CommandMain.cs
goagain: Hi @mjcheetham , thank you for your comment. Overall, yes, we do want restrict to only one auth prompt. For a same resource combination **{this.Resource}\_{this.Client}\_{this.Tenant}**, only the first successfully logged in window takes effect, the rest will use the cached authorization token.

kyle-rader: @mjcheetham  thanks so much for jumping in! Please do not hesitate, you are welcome here :). I agree we actually want `LOCAL\` (which if left off is the default) because of the scenario where two different logged in users are authenticating with the same triplet of GUIDs. 

It turns out the behavior of LOCAL, on Mac, also means two login shells actually don't get mutexed when they should,. So we will continue testing and might need to make this choice platform specific.

reillysiemens: @mjcheetham: Your uninvited commentary is actually a huge win for us. This kind of open, community-driven feedback is half the benefit of moving to GitHub for us. Thank you! ❤️

src/AzureAuth/CommandMain.cs
```diff
@@ -262,10 +262,44 @@ private int GetToken()
...
+
+                // First parameter InitiallyOwned indicated whether this lock is owned by current thread.
+                // It should be false otherwise a dead lock could occur.
+                using (Mutex mutex = new Mutex(false, lockName))
+                {
+                    try
+                    {
+                        // Wait for other session exit.
+                        mutex.WaitOne();
```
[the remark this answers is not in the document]
goagain: It will keep waiting until the mutex released.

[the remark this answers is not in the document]
shalinikhare27: Is that going to cause any experience issues?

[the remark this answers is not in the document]
kyle-rader: We still have a separate bug to enable a reliable timeout. So that _will_ get addressed :).

src/AzureAuth/CommandMain.cs
goagain: I wonder the definition of a **session**, if a client invoke our CLI multiple times, are these CLI actually in one **session**? If so, I guess local mutex is enough for this case.

goagain: @reillysiemens Shalini and I tested on MacOS. What we got at the time was that `Local` behaved differently on Windows and MacOS. On Windows, different PowerShells can share a local mutex, but on MacOS, multiple shells open multiple authentication prompts. If I use `Global`, the mutex works on the OS layer.

src/AzureAuth/CommandMain.cs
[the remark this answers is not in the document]
goagain: @reillysiemens I added a 15-minute timeout for acquiring the mutex. If timeout, a `TimeoutException` will be thrown, which can be catched by telemetry. Does that work for you?

src/AzureAuth/CommandMain.cs
```diff
@@ -52,6 +52,11 @@ public class CommandMain
         private Alias tokenFetcherOptions;
         private ITokenFetcher tokenFetcher;
 
+        /// <summary>
+        /// MutexTimeout is the timeout of Mutex to prevent infinite waiting.
```
reillysiemens, line 56:
```suggestion
        /// The maximum time we will wait to acquire a mutex around prompting the user.

```

src/AzureAuth/CommandMain.cs
```diff
@@ -52,6 +52,11 @@ public class CommandMain
         private Alias tokenFetcherOptions;
         private ITokenFetcher tokenFetcher;
 
+        /// <summary>
+        /// MutexTimeout is the timeout of Mutex to prevent infinite waiting.
+        /// </summary>
+        private TimeSpan mutexTimeout = TimeSpan.FromMinutes(15);
```
reillysiemens, line 58:
```suggestion
        private TimeSpan promptMutexTimeout = TimeSpan.FromMinutes(15);

```
Might be worth a little more specificity here. This particular mutex is guarding the number of user prompts.

src/AzureAuth/CommandMain.cs
```diff
@@ -262,10 +267,52 @@ private int GetToken()
...
             {
                 ITokenFetcher tokenFetcher = this.TokenFetcher();
-                TokenResult tokenResult = (this.tokenFetcherOptions.Scopes == null
-                    ? tokenFetcher.GetAccessTokenAsync(this.CombinedAuthMode)
-                    : tokenFetcher.GetAccessTokenAsync(this.tokenFetcherOptions.Scopes, this.CombinedAuthMode))
-                    .Result;
+                TokenResult tokenResult = null;
+
+                // When running multiple AzureAuth processes with the same resource, client, and tenant IDs,
+
```
reillysiemens, line 273: Looks like there's an accidental extra newline here. I think you can just delete this line.

reillysiemens (dismissed):
@goagain, thanks so much for the hard work on this feature! And again, thanks for your patience. This has been back and forth many times.

I've made some small suggestions here for final changes, but assuming those are fine with you I'm very happy to see this merged.

**Ship It!** 🚀🚢

shalinikhare27 (dismissed):

goagain:
@reillysiemens @shalinikhare27 Thanks for suggestion and reviewing. Seems like you need to re-approve again after applying suggestions.

reillysiemens (approved):

shalinikhare27 (approved):

src/AzureAuth/CommandMain.cs
reillysiemens: Just to close the loop here, @mjcheetham, the final decision we landed on here was to use `Local` on all supported platforms. As you pointed out it definitely seems like the right call on Windows and has essentially the same behavior on macOS except that different login shells (and therefore different terminal windows) can prompt at the same time. Likelihood of two uncached calls from two different terminals happening at the same time is slim though, so we figured that's a reasonable risk to take on for now.

PR #10, opened 2022-04-01
Remove short flags

reillysiemens:
This removes the short flags for `--tenant`, `--clear`, `--domain`, `--auth-mode`, and `--output`.

It also adds a `CHANGELOG.md` in the format described by [keepachangelog.com](https://keepachangelog.com/en/1.0.0/), so that we can start tracking changes between versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Removed
-- Short flags for `--tenant`, `--clear`, `--domain`, `--auth-mode`, and `--output`.
+- The `-t`, `-c`, `-d`, `-m`, and `-o` short flags.
 
 ## [v0.1.0] - 2022-03-30
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,16 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+### Removed
+- Short flags for `--tenant`, `--clear`, `--domain`, `--auth-mode`, and `--output`.
+
+## [v0.1.0] - 2022-03-30
+### Added
+- Initial project release.
+
+[Unreleased]: https://github.com/AzureAD/microsoft-authentication-cli/compare/v0.1.0...HEAD
+[v0.1.0]: https://github.com/AzureAD/microsoft-authentication-cli/releases/tag/v0.1.0
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -25,12 +25,12 @@ namespace Microsoft.Authentication.AzureAuth
     {
         private const string ResourceOption = "--resource";
         private const string ClientOption = "--client";
-        private const string TenantOption = "-t|--tenant";
+        private const string TenantOption = "--tenant";
         private const string ScopeOption = "--scope";
-        private const string ClearOption = "-c|--clear";
-        private const string DomainOption = "-d|--domain";
-        private const string AuthModeOption = "-m|--auth-mode";
-        private const string OutputOption = "-o|--output";
+        private const string ClearOption = "--clear";
+        private const string DomainOption = "--domain";
+        private const string AuthModeOption = "--auth-mode";
+        private const string OutputOption = "--output";
         private const string AliasOption = "--alias";
         private const string ConfigOption = "--config";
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,16 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+### Removed
+- Short flags for `--tenant`, `--clear`, `--domain`, `--auth-mode`, and `--output`.
```
kyle-rader, line 9: I think we should list the actual flags removed, someone glancing at this might easily misread this as removed flag - and see the long names and think those were removed.

CHANGELOG.md
```diff
@@ -0,0 +1,16 @@
+# Changelog
+All notable changes to this project will be documented in this file.
```
shalinikhare27, line 2: Just curious, why is this documentation necessary versus looking at PRs which explicitly in the title talks about the work that happened? 

Also, what would qualify as notable changes? Just looking for examples and need for this. :)

src/AzureAuth/CommandMain.cs
```diff
@@ -25,12 +25,12 @@ public class CommandMain
     {
         private const string ResourceOption = "--resource";
         private const string ClientOption = "--client";
-        private const string TenantOption = "-t|--tenant";
+        private const string TenantOption = "--tenant";
         private const string ScopeOption = "--scope";
-        private const string ClearOption = "-c|--clear";
-        private const string DomainOption = "-d|--domain";
-        private const string AuthModeOption = "-m|--auth-mode";
```
kyle-rader, line 32: For a future change, I think we might want to have this be just `--mode`. `auth` is redundant.

kyle-rader (changes requested):

CHANGELOG.md
kyle-rader: If you haven't already done so, I recommend reading all of https://keepachangelog.com/en/1.0.0/. It talks about some of the philosophy. 

Notable changes IMO are functional differences, that fall into one of the categories of `Changed`, `Added`, `Removed`. Sometimes explicitly adding a `Breaking Changes` is nice. 

Pull requests often have more detail that is needed or desired for a potential user of some tool. They just need to know at a high level what features have been added and don't care about the technical details (but if they want them - they can go to the pull request). So the Changelog is really a form of documentation that has a different target audience from pull requests.

shalinikhare27: Sounds fair! What's the best practice to remember to add these change logs?

kyle-rader: Heh, "Remember to Update them". In the past I've seen PR workflows (pipelines) added to check that the Changelog is edited as a part of each PR. But that's too heavy handed IMO. Not everything well warrant it. Like the PR Rick just merged, nothing ChangeLog worthy there, just house keeping. Also Review.

src/AzureAuth/CommandMain.cs
reillysiemens: Ah, yeah, that's a great suggestion for a follow-up PR.

CHANGELOG.md
reillysiemens: That makes much more sense than what I've done here. I'll go switch that.

CHANGELOG.md
reillysiemens: This is a great question, @shalinikhare27. I think @kyle-rader's done a pretty good job of explaining what this is for. The main audience is our users. Rather than expecting them to sift through PRs or commits between versions we can spell out all the important changes in a single place. Then, when they upgrade from `v0.1.0` to `v0.2.0` (or whatever) they can just read `CHANGELOG.md`. Other projects often just call this file `HISTORY`, but the format @kyle-rader suggested here provides a good framework for determining what changes are impactful enough to be documented.

As for remembering to update this changelog, like @kyle-rader said, you can automate some checks. I've done that in the past before and it because a pain really quickly. Would not recommend it! 🙅‍♂️ Instead, for now, I think a best effort approach is good enough and we can re-evaluate if that becomes problematic.

goagain (dismissed):

melindajohnson:
Simply curious to understand why we are getting rid of all the short flags. Isn't using  `-c` to clear the cache intuitive to users?

reillysiemens:
@melindajohnson: The idea is to use the opportunity presented to us by shipping a new pre-release version of this CLI to "wipe the slate clean" on all the short flags. Short flags can be ambiguous (i.e. should `-c` refer to `--clear` or `--config`?), while long flags always disambiguate.

We did half of that work already earlier when we stopped taking subcommands and only had `--client` and `--resource`. Having `-t|--tenant` made the CLI design inconsistent.

Rather than getting rid of all the short flags I'd say our plan here is to reset them and observe what patterns arise. That way we can eventually add them back in where we know they make sense for our users.

shalinikhare27 (approved):

melindajohnson (approved):

kyle-rader (approved):

goagain (approved):

PR #11, opened 2022-04-02
Add option `--prompt-hint` to enable customized header text for WAM prompts and web mode.

goagain:
# Overview
Taking advantage of this feature will give us a small UX improvement to let WAM based dialogs show header text to indicate what application is requesting auth. When an Auth prompt launches either by native window or web prompt, it’s often hard to know what actually triggered it. Enabling header text in WAM window could help users to know that.
Ref: [microsoft-authentication-library-for-dotnet #3125](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3125)
# Change log
Add option `--prompt-hint` in command line and option `prompt_hint` in TOML config file.
# Test Case
``` cmd
--debug --resource=6e979987-a7c8-4604-9b37-e51f06f08f1a --client=5af6def2-05ec-4cab-b9aa-323d75b5df40 --tenant=9188040d-6c67-4c5b-b112-36a304b66dad --mode=broker --prompt-hint="test header text"
```
## Note
The resource ID `1a0aa568-29b5-4a0c-b895-89da91675dcc` and client ID `c332fbf4-f566-4550-ba1f-83a2f68fe068` are randomly generated. The tenant ID `9188040d-6c67-4c5b-b112-36a304b66dad` is Microsoft account tenant ID. See [Microsoft identity platform access tokens](https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens).
## Overcome Sample
image

# Update on Apr 7th, 2022:
Per discussed with @kyle-rader, `caller` is a better name than `header-text`. With API `WithEmbeddedWebViewOptions`, the title in web mode can also be modified. Therefore, web mode can be also supported.

## Web Mode Overcome
image

# Update on Apr 13th, 2022:
* The final option name is `--prompt-hint`.
* Unit tests have been added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [v0.1.0] - 2022-03-30
 ### Added
 - Initial project release.
-- Option `--caller` to support custom text to prompt caller in web and WAM mode.
+- Option `--prompt-hint` to support custom text to prompt caller in web and WAM mode.
 
 [Unreleased]: https://github.com/AzureAD/microsoft-authentication-cli/compare/v0.1.0...HEAD
 [v0.1.0]: https://github.com/AzureAD/microsoft-authentication-cli/releases/tag/v0.1.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [v0.1.0] - 2022-03-30
 ### Added
 - Initial project release.
+- Option `--caller` to support custom text to prompt caller in web and WAM mode.
 
 [Unreleased]: https://github.com/AzureAD/microsoft-authentication-cli/compare/v0.1.0...HEAD
 [v0.1.0]: https://github.com/AzureAD/microsoft-authentication-cli/releases/tag/v0.1.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Option `--prompt-hint` to support custom text to prompt caller in web and WAM mode.
+
 ### Changed
 - Rename the `--auth-mode` flag to `--mode`.
 
@@ -14,7 +17,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [v0.1.0] - 2022-03-30
 ### Added
 - Initial project release.
-- Option `--prompt-hint` to support custom text to prompt caller in web and WAM mode.
 
 [Unreleased]: https://github.com/AzureAD/microsoft-authentication-cli/compare/v0.1.0...HEAD
 [v0.1.0]: https://github.com/AzureAD/microsoft-authentication-cli/releases/tag/v0.1.0
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Authentication.AzureAuth
         private const string ResourceOption = "--resource";
         private const string ClientOption = "--client";
         private const string TenantOption = "-t|--tenant";
-        private const string HeaderTextOption = "--header";
+        private const string HeaderTextOption = "--header-text";
         private const string ScopeOption = "--scope";
         private const string ClearOption = "-c|--clear";
         private const string DomainOption = "-d|--domain";
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Authentication.AzureAuth
         private const string ResourceOption = "--resource";
         private const string ClientOption = "--client";
         private const string TenantOption = "--tenant";
-        private const string HeaderTextOption = "--header-text";
+        private const string CallerOption = "--caller";
         private const string ScopeOption = "--scope";
         private const string ClearOption = "--clear";
         private const string DomainOption = "--domain";
@@ -103,8 +103,8 @@ Allowed values: [all, web, devicecode]";
         /// <summary>
         /// Gets or sets the customized header text for WAM prompts.
         /// </summary>
-        [Option(HeaderTextOption, "The customized header text for WAM prompts only", CommandOptionType.SingleValue)]
-        public string HeaderText { get; set; }
+        [Option(CallerOption, "The caller name text for WAM prompts and web ", CommandOptionType.SingleValue)]
+        public string Caller { get; set; }
 
         /// <summary>
         /// Gets or sets the scopes.
@@ -174,7 +174,7 @@ Allowed values: [all, web, devicecode]";
                 Client = this.Client,
                 Domain = this.PreferredDomain,
                 Tenant = this.Tenant,
-                HeaderText = this.HeaderText,
+                Caller = this.Caller,
                 Scopes = this.Scopes?.ToList(),
             };
 
@@ -321,7 +321,7 @@ Allowed values: [all, web, devicecode]";
                     new Guid(this.tokenFetcherOptions.Tenant),
                     osxKeyChainSuffix: Constants.AuthOSXKeyChainSuffix,
                     preferredDomain: this.tokenFetcherOptions.Domain,
-                    headerText: this.HeaderText);
+                    caller: this.Caller);
             }
 
             return this.tokenFetcher;
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Authentication.AzureAuth
         private const string ResourceOption = "--resource";
         private const string ClientOption = "--client";
         private const string TenantOption = "-t|--tenant";
-        private const string HeaderTextOption = "-h|--header";
+        private const string HeaderTextOption = "--header";
         private const string ScopeOption = "--scope";
         private const string ClearOption = "-c|--clear";
         private const string DomainOption = "-d|--domain";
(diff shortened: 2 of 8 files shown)


shalinikhare27:
I am curious why header text is an option? Isn't header text going to be automatically generated to let users know what triggered a prompt? Why would a user add header text as an option?

goagain:
> 



> I am curious why header text is an option? Isn't header text going to be automatically generated to let users know what triggered a prompt? Why would a user add header text as an option?

I think this option should be filled by the caller. Maybe @kyle-rader can explain more clear.

reillysiemens:
Technically this change looks sound. All the code is in the right spot and it's very clean, @goagain.

Following on from what @shalinikhare27 said though, I have some questions. I'm interested to know what you and @kyle-rader think about these issues.

1. I thought the goal was just to allow us to customize the WAM prompt to let users know they were being prompted by the `azureauth` CLI, not to allow per `resource-client-tenant` triplet customization of the prompt (though I see how that could be helpful). Do we have enough of a use case to add this to the config file?
2. Is `--header-text` the right name for this flag? It seems to me like this is narrowly scoped to be the header text for a WAM prompt. If we do want to allow this customization at a user level perhaps we could use `--wam-prompt` as a more specific name?
3. There's some awkwardness about having this as an option which only makes sense in conjunction with a specific argument to another option. That is, I _think_ this only makes sense when used with the `--mode broker` flag, but at best it would be a no-op with any other mode. Context dependent flags are often odd in my experience and might be indicative of a larger design issue.
4. This flag is not only context dependent, but also operating system specific. @mvanchaa worked on a feature to make `--mode broker` unavailable on macOS where it doesn't make sense. On a Mac this flag would be like a vestigal organ of sorts.

bgavrilMS:
> I am curious why header text is an option? Isn't header text going to be automatically generated to let users know what triggered a prompt? Why would a user add header text as an option?

The default header text is something like "Let's get you signed in". We exposed this in MSAL as it's currently the only customization option that WAM supports (in fact, that control is not even part of WAM, it's the Account Picker / Account Settings Pane control). 

There are 2 scenarios where you'd want to use this: 

- apps like git + Git Cred Manager, which may run in the background, may get into a situtation where interactive auth is required. This tends to surprise the user, who does not know where the prompt comes from (i.e. "Git needs to authenticate")
- apps like Visual Studio have many components that may trigger auth, so they want to have a way to differentiate between them ("The VS debugger needs to authenticate" vs "The VS Azure Explorer needs to authenticate").

bgavrilMS:
The code looks good @kyle-rader , but I just want to make you aware of the experience with the latest MSAL. If your app targets only Work and School accounts, i.e. authority ends in `/organization` or `/<tenant_id>` we decided to bypass the Account Setting Pane, because it has reliability issues and WAM's AAD plugin can fully handle the auth experience.

So this header text will only appear if the applicaiton targets both AAD and MSA users.

bgavrilMS (dismissed):

src/AzureAuth/CommandMain.cs
```diff
@@ -99,6 +100,12 @@ public CommandMain(CommandExecuteEventData eventData, ILogger<CommandMain> logge
         [Option(TenantOption, "The ID of the Tenant where the client and resource entities exist in", CommandOptionType.SingleValue)]
         public string Tenant { get; set; }
 
+        /// <summary>
+        /// Gets or sets the customized header text for WAM prompts.
```
mvanchaa, line 104:
```suggestion
        /// Gets or sets the customized header text for WAM prompts and web.
```
for consistency with line 106.

src/MSALWrapper/TokenFetcherPublicClient.cs
```diff
@@ -227,7 +235,7 @@ public async Task<TokenResult> GetAccessTokenAsync(IEnumerable<string> scopes, A
                 var pca = this.PCABroker();
                 var pcaWrapper = new PCAWrapper(pca);
                 IAccount account = await this.TryToGetCachedAccountAsync(pca, this.preferredDomain).ConfigureAwait(false) ?? PublicClientApplication.OperatingSystemAccount;
-                result = await this.GetTokenNormalFlowAsync(pcaWrapper, scopes, account);
+                result = await this.GetTokenNormalFlowAsync(pcaWrapper, scopes, null);
```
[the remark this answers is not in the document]
goagain: That's a typo for test. I uploaded it unexpectedly.

src/AzureAuth/CommandMain.cs
goagain: Thank you. I am reviewing. Since these commits contains lots changes, I converted this PR to a draft.

src/MSALWrapper/TokenFetcherPublicClient.cs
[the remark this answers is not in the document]
mvanchaa: oh sorry about that! Just realized that this PR is in draft mode!

src/AzureAuth/CommandMain.cs
mvanchaa: > Thank you. I am reviewing. Since these commits contains lots changes, I converted this PR to a draft.

Got it! sorry didn't notice the draft mode.

goagain: It's ready for review now ; )

shalinikhare27 (changes requested):
Waiting for Mac test.

shalinikhare27 (approved):
Approving since we cannot add this functionality in Mac and we have already tested it in Windows which is a huge win.

shalinikhare27:
@goagain I just realized there are no tests for the new argument introduced in the constructor. Can we please add some tests?

shalinikhare27 (changes requested):
Please add some tests.

reillysiemens:
@goagain: Code-wise this all looks relatively straightforward to me and it's a welcome change, especially with the added context from @bgavrilMS.

The only thing I want to push back on here is the name, `--caller`. I think it's not quite descriptive enough of what the option is for. My earlier suggestion was `--wam-prompt`, which might actually be too specific since it isn't just for WAM. Perhaps just `--prompt`? A "caller" could refer to any number of things, but a "prompt" usually has the association with text provided when asking something of the user.

kyle-rader:
2 things from me:
1. I think `--prompt` is confusing I think with the potential expectation that _it will cause an auth prompt_.
2. I'm okay with having this configurable in the ConfigFile, given callers can still override it on the CLI (And they should to help contextualize why they might be seeing a prompt).

reillysiemens:
@kyle-rader: You make a good point about `--prompt`. I'd be cool with @goagain's suggestion of `--prompt-hint`. Less room for confusion about that _causing_ an auth prompt, I think.

I also have no objection with this in the config file as well as on the command line as long as we find the right name for it. 😄

kyle-rader:
@goagain @reillysiemens I'm good with `--prompt-hint`

src/MSALWrapper.Test/TokenFetcherPublicClientTest.cs
```diff
@@ -371,6 +373,25 @@ public async Task GetTokenNormalFlowAsync_GetTokenInteractive_Throws_OperationCa
             tokenFetcher.ErrorsList[2].Message.Should().Be("Interactive Auth (with extra claims) timed out after 15 minutes.");
         }
 
+        /// <summary>
+        /// .
```
reillysiemens, line 377: We could probably use a better comment here than `.`.

reillysiemens (changes requested):
I caught one test method that could use a better comment, otherwise I'd be happy to put a ✅ on this.

src/MSALWrapper.Test/TokenFetcherPublicClientTest.cs
goagain: Good catch. I forgot to write the final comment😅

reillysiemens (approved):

reillysiemens (approved):

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [v0.1.0] - 2022-03-30
 ### Added
 - Initial project release.
+- Option `--prompt-hint` to support custom text to prompt caller in web and WAM mode.
```
kyle-rader, line 17: This line should go under the unreleased section. v0.1.0 is already released ;)

src/MSALWrapper.Test/TokenFetcherPublicClientTest.cs
```diff
@@ -73,7 +75,7 @@ public void Setup()
              .AddTransient<TokenFetcherPublicClient>((provider) =>
              {
                  var logger = provider.GetService<ILogger<TokenFetcherPublicClient>>();
-                 return new TokenFetcherPublicClient(logger, ResourceId, ClientId, TenantId);
+                 return new TokenFetcherPublicClient(logger, ResourceId, ClientId, TenantId, promptHint: this.promptHint);
```
kyle-rader, line 78: Here I think we should build a prompt hint to pass in, so that is always has `AzureAuth CLI` at the beginning. Perhaps a helper method to construct our prompt hint, that we can call here. 

Something like
```C#
private const string PromptHintPrefix = "Azure AUth CLI";
private string GetPromptHint() {
  if (string.IsNullOrEmpty(this.PromptHint) {
    return PromptHintPrefix;
  }
  else
  {
    return $"{PromptHintPrefix: {this.PromptHint}";
  }
}
```

kyle-rader (changes requested):

mvanchaa (dismissed):

CHANGELOG.md
reillysiemens: Wow. I totally missed that. Good catch, @kyle-rader.

src/MSALWrapper/TokenFetcherPublicClient.cs
```diff
@@ -125,6 +131,8 @@ public TokenFetcherPublicClient(ILogger logger, Guid resourceId, Guid clientId,
             this.resourceId = resourceId;
             this.clientId = clientId;
 
+            this.headerText = headerText;
```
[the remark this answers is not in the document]
reillysiemens: I agree with the default, but I don't know about the `AzureAuth CLI: <input text here>` option. For some consumers we might be an implementation detail and they might prefer we just present their specified prompt text unmodified. I'm not sure we have any customers to query on the matter though. I could be totally off base here.

src/MSALWrapper.Test/TokenFetcherPublicClientTest.cs
goagain: Created a new issue #24

src/MSALWrapper/TokenFetcherPublicClient.cs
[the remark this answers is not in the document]
goagain: Created a new issue https://github.com/AzureAD/microsoft-authentication-cli/issues/24

kyle-rader (approved):

PR #14, opened 2022-04-06
Add installation scripts

reillysiemens:
## Windows

This adds a PowerShell bootstrap script which can be used to quickly install the application without the hassle of extracting the DLLs from a release `.zip`, adding the binary to the `$PATH`, etc.

You can test this for yourself by opening a PowerShell terminal and running the installation script from this branch. The `-Verbose` flag will provide extra debugging output so you can see what's going on.

```powershell
$env:AZUREAUTH_VERSION = 'v0.1.0'
iex "& { $(irm https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/install-scripts/install/install.ps1) } -Verbose"
```

This script doesn't do a few things right now.

1. It doesn't allow downloading the latest release. We can get this information from the GitHub API in JSON form. PowerShell has JSON parsing capabilities baked in, but macOS does not and I'm focusing on the absolute minimum for having the same behavior across all supported platforms.
2. It doesn't clean up old versions. We can add this functionality later, but my lack of PowerShell knowledge is a hindrance and we want to ship this feature quickly.
4. It can't be used to uninstall the application. We will likely want a mechanism for that as well, but it might be best in a separate script.

## macOS

It also adds a shell script for Unix systems (currently only supporting macOS).

You can test this for yourself by opening a terminal and running the installation script from this branch. The `AZUREAUTH_VERBOSE_INSTALL` environment variable will provide extra debugging output so you can see what's going on.

```bash
export AZUREAUTH_VERSION='v0.1.0'
export AZUREAUTH_VERBOSE_INSTALL=1
curl -sL https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/install-scripts/install/install.sh | sh
```

The limitations are currently the same as the Windows version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,17 +22,22 @@ On Windows we provide a PowerShell bootstrap script, which will download and ext
 means of downloading the latest release, so you **must** specify your desired version via the `$AZUREAUTH_VERSION`
 environment variable.
 
-To install the application, set the environment variable (`v0.1.0` is an example. See
-[releases](https://github.com/AzureAD/microsoft-authentication-cli/releases) for the latest).
+To install the application, run
 
 ```powershell
+# v0.1.0 is an example. See https://github.com/AzureAD/microsoft-authentication-cli/releases for the latest.
 $env:AZUREAUTH_VERSION = 'v0.1.0'
+iex "& { $(irm https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/main/install/install.ps1) } -Verbose"
 ```
 
-Then execute the installation script (excuse the verbosity, we avoided using `Invoke-Expression`).
+Or, if you want a method more resilient to failure than `Invoke-Expression`, run
 
 ```powershell
-$url = 'https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/main/install/install.ps1'; $script = "${env:TEMP}\install.ps1"; irm $url -OutFile $script; if ($?) { &$script }; if ($?) { rm $script }
+# v0.1.0 is an example. See https://github.com/AzureAD/microsoft-authentication-cli/releases for the latest.
+$env:AZUREAUTH_VERSION = 'v0.1.0'
+$script = "${env:TEMP}\install.ps1"
+$url = 'https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/main/install/install.ps1'
+Invoke-WebRequest $url -OutFile $script; if ($?) { &$script }; if ($?) { rm $script }
 ```
 
 **Note**: The script does not signal currently running processes to update their environments, so you'll need to
@@ -44,15 +49,11 @@ On macOS we provide a shell bootstrap script, which will download and extract th
 and automatically add the `azureauth` binary to your `$PATH`. We don't currently provide a means of downloading the
 latest release, so you **must** specify your desired version via the `$AZUREAUTH_VERSION` environment variable.
 
-To install the application, set the environment variable (`v0.1.0` is an example. See
-[releases](https://github.com/AzureAD/microsoft-authentication-cli/releases) for the latest).
+To install the application, run
+
 ```bash
+# v0.1.0 is an example. See https://github.com/AzureAD/microsoft-authentication-cli/releases for the latest.
 export AZUREAUTH_VERSION='v0.1.0'
-```
-
-Then execute the installation script.
-
-```bash
 curl -sL https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/main/install/install.sh | sh
 ```
 
--- a/README.md
+++ b/README.md
@@ -33,6 +33,24 @@ iex "& { $(irm https://raw.githubusercontent.com/AzureAD/microsoft-authenticatio
 **Note**: The script does not signal currently running processes to update their environments, so you'll need to
 relaunch applications before the `$PATH` changes take effect.
 
+## macOS
+
+On macOS we provide a shell bootstrap script, which will download and extract the application to `$HOME/.azureauth`
+and automatically add the `azureauth` binary to your `$PATH`. We don't currently provide a means of downloading the
+latest release, so you **must** specify your desired version via the `$AZUREAUTH_VERSION` environment variable.
+
+To install the application, run
+
+```bash
+# v0.1.0 is just an example here. See https://github.com/AzureAD/microsoft-authentication-cli/releases for the latest.
+export AZUREAUTH_VERSION='v0.1.0'
+curl https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/main/install/install.sh | sh
+```
+
+**Note**: The script currently only updates the `$PATH` in `~/.bashrc` and `~/.zshrc`. It does not update the environment
+of currently running processes, so you'll need to relaunch applications (or source your shell profile) before the `$PATH`
+changes take effect.
+
 # Data Collection
 
 The software may collect information about you and your use of the software and send it to Microsoft. Microsoft may use
--- a/README.md
+++ b/README.md
@@ -27,7 +27,9 @@ To install the application, run
 ```powershell
 # v0.1.0 is just an example here. See https://github.com/AzureAD/microsoft-authentication-cli/releases for the latest.
 $env:AZUREAUTH_VERSION = 'v0.1.0'
-iex "& { $(irm https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/main/install/install.ps1) }"
+$script = "${env:TEMP}\install.ps1"
+$url = 'https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/main/install/install.ps1'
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -13,6 +13,26 @@ The CLI is designed for authenticating and returning an access token for public
...
+`%LOCALAPPDATA%\AzureAuth` and automatically add the `azureauth` binary to your `$PATH`. We don't currently provide a
+means of downloading the latest release, so you **must** specify your desired version via the `$AZUREAUTH_VERSION`
+environment variable.
+
+To install the application, run
+
+```powershell
+# v0.1.0 is just an example here. See https://github.com/AzureAD/microsoft-authentication-cli/releases for the latest.
+$env:AZUREAUTH_VERSION = 'v0.1.0'
+iex "& { $(irm https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/main/install/install.ps1) }"
```
kyle-rader, line 30: `irm` doesn't seem to exist in PowerShell 5 - which is the default PowerShell on Windows. I think the installs script should aim to be 5 and 7 compatible.

kyle-rader:
```
❯ iex "& { $(irm https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/main/install/install.ps1) }"
irm : 404: Not Found
At line:1 char:12
+ iex "& { $(irm https://raw.githubusercontent.com/AzureAD/microsoft-au ...
+            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:HttpWebRequest) [Invoke-RestMethod], WebException
    + FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShell.Commands.InvokeRestMethodCommand
```

kyle-rader: Actually, this doesn't work for me in PS 7.2.2 either :/
```
PowerShell 7.2.2
Copyright (c) Microsoft Corporation.

https://aka.ms/powershell
Type 'help' to get help.


~
❯ iex "& { $(irm https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/main/install/install.ps1) }"
Invoke-RestMethod: 404: Not Found
```

kyle-rader: `iwr` might be a good alternative.

kyle-rader (changes requested):

install/install.ps1
```diff
@@ -0,0 +1,65 @@
...
+# We don't currently have good cross-platform options for determining the latest release version, so we require
+# knowledge of the specific target version, which the user should set as an environment variable.
+$version = $Env:AZUREAUTH_VERSION
+if ($null -eq $version) {
+    # Write-Error is terminal with ErrorActionPreference='Stop', so we continue to hit the exit
+    # and set an exit code.
+    Write-Error 'No $AZUREAUTH_VERSION specified, unable to download a release' -ErrorAction:Continue
+    exit 1
+}
+
```
kyle-rader, line 17: I am trying to confirm this catches my not having the env var set but it doesn't seem to work in PS5??

```
❯ iex "& { $(irm https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/install-scripts/install/install.ps1) } -Verbose"
iex : At line:18 char:41
+ $repo = ($null -eq $Env:AZUREAUTH_REPO) ? 'AzureAD/microsoft-authenti ...
+                                         ~
Unexpected token '?' in expression or statement.
At line:1 char:1
+ iex "& { $(irm https://raw.githubusercontent.com/AzureAD/microsoft-au ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ParserError: (:) [Invoke-Expression], ParseException
    + FullyQualifiedErrorId : UnexpectedToken,Microsoft.PowerShell.Commands.InvokeExpressionCommand
```

kyle-rader: `?` was added in PS 7, but not in 5.
https://stackoverflow.com/questions/31341998/ternary-operator-in-powershell

install/install.ps1
```diff
@@ -0,0 +1,65 @@
...
+$script:ErrorActionPreference='Stop'
+
+# We don't currently have good cross-platform options for determining the latest release version, so we require
+# knowledge of the specific target version, which the user should set as an environment variable.
+$version = $Env:AZUREAUTH_VERSION
+if ($null -eq $version) {
+    # Write-Error is terminal with ErrorActionPreference='Stop', so we continue to hit the exit
+    # and set an exit code.
+    Write-Error 'No $AZUREAUTH_VERSION specified, unable to download a release' -ErrorAction:Continue
+    exit 1
```
kyle-rader, line 15: I think we want `return 1` here so that executing this doesn't end the callers entire PS session. In old style terminal it closes, in Windows Terminal (new) it leaves a dead session tab.

reillysiemens: Hmm... I don't think `return 1` actually sets the exit code, which is important. I'll have to look into this. Good catch.

kyle-rader:
@reillysiemens - unfortunately I'm also getting an error on Intel Mac:
```
❯ curl -sL https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/install-scripts/install/install.sh | sh
Creating /Users/kyrader/.azureauth
Downloading https://github.com/AzureAD/microsoft-authentication-cli/releases/download/0.1.0/azureauth-0.1.0-osx-x64.tar.gz to /Users/kyrader/.azureauth/azureauth-0.1.0-osx-x64.tar.gz
Extracting /Users/kyrader/.azureauth/azureauth-0.1.0-osx-x64.tar.gz to /Users/kyrader/.azureauth/0.1.0
tar: Error opening archive: Unrecognized archive format
```

README.md
```diff
@@ -13,6 +13,44 @@ The CLI is designed for authenticating and returning an access token for public
...
+```
```
kyle-rader, line 31: UX: I'd prefer we split these into 2 separate blocks so that each is copy-able on it's own. Some terminals will warn you about trying to paste multi-lines, and often I don't like to paste multi-lines into my terminal away because I can't change what's happening. It appears you can't add suggestions that span other code blocks but I suggest making the comment normal text, and then making each of these it's own code block.

README.md
```diff
@@ -13,6 +13,44 @@ The CLI is designed for authenticating and returning an access token for public
...
+
+## macOS
+
+On macOS we provide a shell bootstrap script, which will download and extract the application to `$HOME/.azureauth`
+and automatically add the `azureauth` binary to your `$PATH`. We don't currently provide a means of downloading the
+latest release, so you **must** specify your desired version via the `$AZUREAUTH_VERSION` environment variable.
+
+To install the application, run
+
+```bash
```
kyle-rader, line 44: Same usability comment here. It will far easier to use these if each is in it's own block.

kyle-rader (changes requested):

README.md
reillysiemens: > so that each is copy-able on its own

That's already the case. Nothing forces someone to copy all the lines at once.

> Some terminals will warn you about trying to paste multi-lines

That seems like reasonable behavior

> I can't change what's happening

Unless you're using a bad terminal emulator you usually _can_ change what's happening. Though, I honestly can't speak for `cmd.exe` in this case, but the text is specific to PowerShell anyway.

> I suggest making the comment normal text

I purposefully made the text there a comment because I thought it would increase the likelihood that a dev about to paste code into their terminal would read it, but I'm fine to move it above as well.

I'm happy to make these changes because I'm not really tied to the presentation here, but I'm not convinced that they improve the UX of this document in a significant way.

kyle-rader: Tucker and talked about this offline and I agreed with his comments, leaving the block as one is the better 90% use case to make this just work for people.

README.md
```diff
@@ -13,6 +13,54 @@ The CLI is designed for authenticating and returning an access token for public
 | Ubuntu (linux)   | ❌ (not yet) | ✅ | ✅ | ✅ | ❌ (not yet) |
 <br/>
 
+# Installation
+
+## Windows
+
+On Windows we provide a PowerShell bootstrap script, which will download and extract the application to
+`%LOCALAPPDATA%\AzureAuth` and automatically add the `azureauth` binary to your `$PATH`. We don't currently provide a
```
kyle-rader, line 21:
```suggestion
`%LOCALAPPDATA%\Programs\AzureAuth` and automatically add the `azureauth` binary to your `$PATH`. We don't currently provide a
```

kyle-rader (changes requested):

kyle-rader (approved):

melindajohnson (approved):

goagain (approved):

mvanchaa (approved):

shalinikhare27 (commented):
Late in the game, but looks good. :)

PR #16, opened 2022-04-07
Print a warning message when opening a browser window for auth

shivanshu3:
Note that this will not pollute stdout since the access token is writted to stdout. Warning messages are written to stderr.

### DIFF
--- a/src/MSALWrapper/TokenFetcherPublicClient.cs
+++ b/src/MSALWrapper/TokenFetcherPublicClient.cs
@@ -30,6 +30,8 @@ namespace Microsoft.Authentication.MSALWrapper
         /// </summary>
         public const string OEAUTH_MSAL_DISABLE_CACHE = "OEAUTH_MSAL_DISABLE_CACHE";
 
+        private const string INTERACTIVE_AUTH_MESSAGE = "Opening a browser window for interactive auth...";
+
         /// <summary>
         /// The _errors.
         /// </summary>
@@ -314,8 +316,6 @@ namespace Microsoft.Authentication.MSALWrapper
 
             try
             {
-                const string interactiveAuthMessage = "Opening a browser window for interactive auth...";
-
                 try
                 {
                     try
@@ -333,7 +333,7 @@ namespace Microsoft.Authentication.MSALWrapper
                     {
                         this.ErrorsList.Add(ex);
                         this.logger.LogDebug($"Silent auth failed, re-auth is required.\n{ex.Message}");
-                        this.logger.LogWarning(interactiveAuthMessage);
+                        this.logger.LogWarning(INTERACTIVE_AUTH_MESSAGE);
                         var tokenResult = await this.CompleteWithin(
                             this.interactiveAuthTimeout,
                             "Interactive Auth",
@@ -347,7 +347,7 @@ namespace Microsoft.Authentication.MSALWrapper
                 {
                     this.ErrorsList.Add(ex);
                     this.logger.LogDebug($"Silent auth failed, re-auth is required.\n{ex.Message}");
-                    this.logger.LogWarning(interactiveAuthMessage);
+                    this.logger.LogWarning(INTERACTIVE_AUTH_MESSAGE);
                     var tokenResult = await this.CompleteWithin(
                         this.interactiveAuthTimeout,
                         "Interactive Auth (with extra claims)",
--- a/src/MSALWrapper/TokenFetcherPublicClient.cs
+++ b/src/MSALWrapper/TokenFetcherPublicClient.cs
@@ -314,6 +314,8 @@ namespace Microsoft.Authentication.MSALWrapper
 
             try
             {
+                const string interactiveAuthMessage = "Opening a browser window for interactive auth...";
+
                 try
                 {
                     try
@@ -331,6 +333,7 @@ namespace Microsoft.Authentication.MSALWrapper
                     {
                         this.ErrorsList.Add(ex);
                         this.logger.LogDebug($"Silent auth failed, re-auth is required.\n{ex.Message}");
+                        this.logger.LogWarning(interactiveAuthMessage);
                         var tokenResult = await this.CompleteWithin(
                             this.interactiveAuthTimeout,
                             "Interactive Auth",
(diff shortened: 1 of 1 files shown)


src/MSALWrapper/TokenFetcherPublicClient.cs
```diff
@@ -331,6 +333,7 @@ public async Task<TokenResult> GetTokenNormalFlowAsync(IPCAWrapper pcaWrapper, I
                     {
                         this.ErrorsList.Add(ex);
                         this.logger.LogDebug($"Silent auth failed, re-auth is required.\n{ex.Message}");
+                        this.logger.LogWarning(interactiveAuthMessage);
```
shalinikhare27, line 336:
```suggestion
                        this.logger.LogWarning($"Opening a browser window for interactive auth.\n{ex.Message}");

```
for consistency.

src/MSALWrapper/TokenFetcherPublicClient.cs
```diff
@@ -344,6 +347,7 @@ public async Task<TokenResult> GetTokenNormalFlowAsync(IPCAWrapper pcaWrapper, I
                 {
                     this.ErrorsList.Add(ex);
                     this.logger.LogDebug($"Silent auth failed, re-auth is required.\n{ex.Message}");
+                    this.logger.LogWarning(interactiveAuthMessage);
```
shalinikhare27, line 350:
```suggestion
                    this.logger.LogWarning($"Opening a browser window for interactive auth.\n{ex.Message}");

```

src/MSALWrapper/TokenFetcherPublicClient.cs
```diff
@@ -314,6 +314,8 @@ public async Task<TokenResult> GetTokenNormalFlowAsync(IPCAWrapper pcaWrapper, I
 
             try
             {
+                const string interactiveAuthMessage = "Opening a browser window for interactive auth...";
+
```
shalinikhare27, line 318: [nit]: extra line

src/MSALWrapper/TokenFetcherPublicClient.cs
```diff
@@ -314,6 +314,8 @@ public async Task<TokenResult> GetTokenNormalFlowAsync(IPCAWrapper pcaWrapper, I
 
             try
             {
+                const string interactiveAuthMessage = "Opening a browser window for interactive auth...";
```
shalinikhare27, line 317: We can use the warning message directly while logging the warning.

goagain (changes requested):
Why use LogWarning? I think LogInfo is a better choice.

src/MSALWrapper/TokenFetcherPublicClient.cs
shivanshu3: Ah, the reason I did it this way was to avoid duplicating the same message string twice in the code.

shalinikhare27: Makes sense. Probably move it as a `const string` on the top.

shivanshu3:
@goagain - I want this message to be displayed by default without having to enable info/debug/trace logging. I also want this log to go to stderr instead of stdout because only the access token should be printed on stdout and nothing else. That's why I used LogWarning.

shalinikhare27:
@shivanshu3 we have a bug on our plate (sent on Teams) to do the exact same thing after we refactor our code. I am okay you doing this change now since I wouldn't want you to wait couple of weeks for this.

@reillysiemens - Since we are shopping from github anytime soon, and our code in ADO and github are now quite different, do we just add this change now? 

Shivan, I am thinking your concern addressed versus shipping CLI from github can be a challenge.

shivanshu3:
@shalinikhare27 that's ok, we're not in any rush to get this change deployed soon.

reillysiemens:
@shivanshu3: As @shalinikhare27 mentioned, we're not yet in a position to do continuous deployment (and we may not ever go that route). We've got a feature open to improve our CI pipeline, but probably won't finish it until early May. After that point we'll try to ship as rapidly as we can.

We can, however, land this feature now and make sure it gets included in our next release. I'll go ahead and do a review now.

src/MSALWrapper/TokenFetcherPublicClient.cs
reillysiemens: What is the actual text that MSAL comes back with in this scenario? Is it helpful to the user or are we better off with just this simple message?

My thinking here is that the full message seems really useful for debugging, but might be less useful as warning text.

src/MSALWrapper/TokenFetcherPublicClient.cs
reillysiemens: Same thought as above here.

src/MSALWrapper/TokenFetcherPublicClient.cs
reillysiemens: Did you mean a `const string` at the class level, @shalinikhare27? If so, that's reasonable to me.

shalinikhare27: That's right.

src/MSALWrapper/TokenFetcherPublicClient.cs
shalinikhare27: Giving it a second thought, we wouldn't need to add this detail to the customer. Simple message should be good.

src/MSALWrapper/TokenFetcherPublicClient.cs
```diff
@@ -30,6 +30,8 @@ public class TokenFetcherPublicClient : ITokenFetcher
         /// </summary>
         public const string OEAUTH_MSAL_DISABLE_CACHE = "OEAUTH_MSAL_DISABLE_CACHE";
 
+        private const string INTERACTIVE_AUTH_MESSAGE = "Opening a browser window for interactive auth...";
```
shalinikhare27, line 33:
```suggestion
        private const string INTERACTIVE_AUTH_MESSAGE = "Opening a browser window for interactive auth.";

```
[nit]

shalinikhare27 (dismissed):

goagain (approved):

reillysiemens (approved):

melindajohnson (approved):

kyle-rader (commented):
As Shalini said, we do have a bug related to this that I think should cover this scenario. I'd prefer we hold off on merging this for now, given it will just have to be moved again. The messaging will also need to be slightly different per auth mode and platform.

shalinikhare27:
@shivanshu3 closing this PR as we are doing a major refactoring and @kyle-rader has a bug on his plate to wrap this up.

PR #17, opened 2022-04-08
Clean up xml doc comments for example projects

mvanchaa:
Fixed `StyleCop` related issues on example projects. 
Removed `StyleCop` NuGet dependency on the example projects because we don't need `StyleCop` warnings or errors on them as there wont be any active developments in these projects.


goagain:
I think we'd better reuse the existing `stylecop.json` and `GlobalSupressions.cs`, instead of creating a new one.
Did you see the warnings in Program.cs:1:1. I cannot figure out why.

examples/Demo.Console.NET6/Demo.Console.NET6.csproj
```diff
@@ -20,6 +20,10 @@
     <PackageReference Include="Microsoft.TeamFoundationServer.Client" Version="16.153.0" />
     <PackageReference Include="NLog" Version="4.7.6" />
     <PackageReference Include="NLog.Extensions.Logging" Version="1.6.5" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.1.118">
+      <PrivateAssets>all</PrivateAssets>
```
shalinikhare27, line 24: Curious why we needed to add these? Just stylecop to trigger stylecop should have been enough.

examples/Demo.Console.NET6/stylecop.json
```diff
@@ -0,0 +1,13 @@
+﻿{
```
shalinikhare27, line 1: We can add stylecop as a link to this project so that we do not need to include it as a separate file here. We already have a global style cop file as the stylecop folder here which should be used here.

shalinikhare27:
> I think we'd better reuse the existing `stylecop.json` and `GlobalSupressions.cs`, instead of creating a new one. Did you see the warnings in Program.cs:1:1. I cannot figure out why.

Pretty much the same thing I am suggesting here. Let's reuse the global styelcop json file and figure out why is github complaining about the Program.cs file. It's related to xml comments in the header.

shalinikhare27 (changes requested):
Waiting for to address my comments.

shalinikhare27:
@kyle-rader Can't find your comment in this PR but replying back to this - "I think we explicitly removed StyleCop from the Demo projects as strictly enforcing the rules here isn't as important?"

I agree with you and did advise @mvanchaa the same. She was keen on going this route because there were barely files to add stylecop rules.

@mvanchaa - You can still nuke this PR by removing stylecop policies on demo projects as it's not needed. :)

mvanchaa:
@shalinikhare27 @goagain I can see `stylecop.json` files under all the projects in Visual Studio. I don't know why. So I got confused about the global stylecop usage and thought of creating one for `.NET6` demo project. Attaching the screenshot. 
I 
image

I will remove stylecop policies on this demo project and just fix stylecop rules. I did not get any warnings or errors related to stylecop on `.NET6` demo project without adding the stylecop NuGet package to the project.

goagain:
> 

The shared `stylecop.json` has a shortcut icon on the bottom-left corner.

mvanchaa:
> > 
> 
> The shared `stylecop.json` has a shortcut icon on the bottom-left corner.

oh that makes perfect sense now. Thanks Rick!

examples/Demo.Console.NET6/Demo.Console.NET6.csproj
mvanchaa: I dont know why, stylecop did not trigger any warnings or errors on this project.

reillysiemens: That's curious... seems like we should be able to follow the same pattern as our other projects. I think it would be good to understand why this is needed before we merge.

examples/Demo.Console.NET6/Program.cs
```diff
@@ -1,22 +1,24 @@
 // Copyright (c) Microsoft Corporation.
```
reillysiemens, line 1: We might need to adjust the configuration of stylecop here. This is the correct comment and we don't need XML surrounding it, but GitHub Actions is still complaining that
```
The file header XML is invalid.
```

examples/Demo.Console.NET6/stylecop.json
reillysiemens: Yeah, let's definitely avoid duplicating this config where we can. 😃

examples/Demo.Console.NETFramework472/Program.cs
```diff
@@ -39,14 +39,10 @@ internal class Program
...
-        /// The on execute.
+        /// This method executes the auth process. This method is invoked when lasso is launched.
         /// </summary>
-        /// <param name="logger">
-        /// The logger.
-        /// </param>
-        /// <returns>
-        /// The <see cref="int"/>.
-        /// </returns>
+        /// <param name="logger"> The logger. </param>
```
reillysiemens, line 44: No need for the extra spaces, I think.

```suggestion
        /// <param name="logger">The logger.</param>
```

reillysiemens (changes requested):
Just some small stuff here before merging. Thanks for working on this, @mvanchaa. 😄

mvanchaa:
StyleCop was installed on all projects except `Demo.Console.NET6` project. Hence, there were no stylecop errors. I installed it directly on the `Demo.Console.NET6` project, and hence there is an additional `stylecop.json` file. I removed that file now and I installed `StyleCop` on the solution level which originally included all but this project.

examples/Demo.Console.NET6/Program.cs
mvanchaa: > StyleCop was installed on all projects except `Demo.Console.NET6` project. Hence, there were no stylecop errors. I installed it directly on the `Demo.Console.NET6` project, and hence there is an additional `stylecop.json` file. I removed that file now and I installed `StyleCop` on this project at the solution level which originally included `StyleCop` on all but this project.

To be clear, `StyleCop` at solution level looked like this. Changed it now.  
image

shalinikhare27:
Can you please add a descriptive PR on removing stylecop from example files etc. Since I am signing this off, I want Tucker and Kyle to know what exactly happened here.

shalinikhare27 (approved):

goagain:
@mvanchaa Can you update to the latest version? Otherwise after our approvals, you have to update the branch and request our re-approvals.

goagain (approved):

melindajohnson (approved):

PR #18, opened 2022-04-08
Args in Demo do not match the number of needed arguments.

goagain:
The number of arguments in `Demo.Console.Net6` should be 3 instead of 2. The index has also been changed to 0-based.
Ref: [Main() and command-line arguments
](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/program-structure/main-command-line)


reillysiemens (approved):
Nice catch, @goagain! Looks good to me!

shalinikhare27 (approved):

PR #21, opened 2022-04-11
Created an interface for PCACache to help separate responsibilities for caching logic.

shalinikhare27:
# What and Why
This PR contains an `IPCACache` that implements `SetupTokenCache`. This class will be shared across Broker, Web and DeviceCode auth flow and so I separated the responsibilities.

# Things to note
Moved `msal disable cache` flag into Constants file.
Created a `IPCACache` that implements `SetupTokenCache` and expects an errorList as an input. Currently this method is part of the `TokenFetcherPublicClient` class.

# Test
Nothing to test so far. This is just a refactor.

# Class diagram
The class diagram below gives a representation on how these classes are used. For brevity, I have only represented `AuthFlowBroker` but you could have any auth flow and the flow will look the same more or less.

```mermaid
classDiagram
      IAuthFlows <|-- AuthFlows
      IAuthFlows <|-- AuthFlowBroker
      AuthFlowBroker*-- PCACache
      AuthFlowBroker*-- AccountsProvider
      AuthFlowBroker*-- PCAWrapper
      AuthFlowBroker*-- TaskExecutor
     AuthFlowBroker*-- MsalHttpClientFactoryAdaptor
     AuthFlowBroker*-- TokenResultExtensions
      IAuthFlows : +Task<TokenResult> GetTokenAsync()
      class AuthFlows{
          +ILogger logger
          +Guid resourceId
          +Guid clientId
          +Guid tenantId
          +bool verifyPersistence = false
      }
      class AuthFlowBroker{
          +ILogger logger
          +Guid clientId
          +Guid tenantId
          +bool verifyPersistence = false
      }
```

### DIFF
--- a/src/MSALWrapper/Constants.cs
+++ b/src/MSALWrapper/Constants.cs
@@ -26,5 +26,10 @@ namespace Microsoft.Authentication.MSALWrapper
         /// The aad redirect uri.
         /// </summary>
         public static readonly Uri AadRedirectUri = new Uri("http://localhost");
+
+        /// <summary>
+        /// The msal disabled cache.
+        /// </summary>
+        internal const string OEAUTH_MSAL_DISABLE_CACHE = "OEAUTH_MSAL_DISABLE_CACHE";
     }
 }
--- a/src/MSALWrapper/Constants.cs
+++ b/src/MSALWrapper/Constants.cs
@@ -29,7 +29,6 @@ namespace Microsoft.Authentication.MSALWrapper
 
         /// <summary>
         /// The name of an environment variable used to disable file cache configuration.
-
         /// </summary>
         internal const string OEAUTH_MSAL_DISABLE_CACHE = "OEAUTH_MSAL_DISABLE_CACHE";
     }
--- a/src/MSALWrapper/PCACache/IPCACache.cs
+++ b/src/MSALWrapper/PCACache/IPCACache.cs
@@ -12,12 +12,7 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlows
     internal interface IPCACache
     {
         /// <summary>
-        /// The setup token cache.
-        /// </summary>
-        void SetupTokenCache();
-
-        /// <summary>
-        /// The setup token cache.
+        /// Sets up the token cache.
         /// </summary>
         /// <param name="errorsList">The errors list.</param>
         void SetupTokenCache(List<Exception> errorsList);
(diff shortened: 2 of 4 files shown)


kyle-rader (changes requested):

src/MSALWrapper/PCACache/IPCACache.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// The PCACache interface.
+    /// </summary>
+    internal interface IPCACache
+    {
+        /// <summary>
```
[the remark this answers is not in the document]
shalinikhare27: Makes sense. Got rid of the override one.

kyle-rader (approved):

src/MSALWrapper/Constants.cs
```diff
@@ -26,5 +26,10 @@ public static class Constants
         /// The aad redirect uri.
         /// </summary>
         public static readonly Uri AadRedirectUri = new Uri("http://localhost");
+
+        /// <summary>
+        /// The name of an environment variable used to disable file cache configuration.
+        /// </summary>
+        internal const string OEAUTH_MSAL_DISABLE_CACHE = "OEAUTH_MSAL_DISABLE_CACHE";
```
reillysiemens, line 33: Not a blocker (and in fact it would be a breaking change if we made it), but this should probably be `AZUREAUTH_MSAL_DISABLE_CACHE` to match our current environment variable naming convention.

src/MSALWrapper/PCACache/IPCACache.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.Authentication.MSALWrapper.AuthFlows
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// The PCACache interface.
+    /// </summary>
+    internal interface IPCACache
```
reillysiemens, line 12: Not a blocker, but more of an open question that may only resolve itself as we include other auth types. Is this cache actually only usable for public client flows or would it also be suitable for confidential client ones as well? 🤔

reillysiemens (approved):

src/MSALWrapper/Constants.cs
shalinikhare27: Good catch!

src/MSALWrapper/PCACache/IPCACache.cs
shalinikhare27: Kyle and I were looking into confidential client top level and did see cache. I will have better answers as I deep dive into each Auth flow. 

@kyle-rader do you know top of your head?

PR #19, opened 2022-04-11
Redesigned BrokerPCATokenFetcher and refactored more code

shalinikhare27:
Used composition and polymorphism to refactor and redesign the TokenFetcherPublicClient class.
Created an interface IAuthFlow that implements a GetTokenAsync method.
Each of the current AuthFlows - Broker, Web, DeviceCode are moved into their own class and they implement the IAuthFlow interface.
AuthFlows class creates a list of Auth flows to execute one by one.
Created an interface for AccountsProvider and PCACache to help separate responsibilities for providing an account and cache set respectively.
Refactored utility code into its own classes - MsalHttpClientFactoryAdapter for providing httpclient, TokenResultExtensions for augmenting tokenResult with auth type, and TaskExecutor to abstract timeout handling.

This PR specifically focuses on the Broker auth flow. Web and DeviceCode auth flow will be part of the next PR.

Please click Preview to check out the class diagram below:

```mermaid
classDiagram
      IAuthFlows <|-- AuthFlows
      IAuthFlows <|-- BrokerPCATokenFetcher 
      BrokerPCATokenFetcher *-- PCACache
      BrokerPCATokenFetcher *-- AccountsProvider
      BrokerPCATokenFetcher *-- PCAWrapper
      BrokerPCATokenFetcher *-- TaskExecutor           
      IAuthFlows : +Task<TokenResult> GetTokenAsync()
      PCAWrapper *-- IPublicClientApplication
      class AuthFlows{
          +ILogger logger
          +Guid resourceId
          +Guid clientId
          +Guid tenantId
          +IEnumerable<string> scopes = null
          +string osxKeyChainSuffix = null
          +string preferredDomain = null, 
          +bool verifyPersistence = false
      }
      class BrokerPCATokenFetcher {
          +ILogger logger
          +Guid clientId
          +Guid tenantId
          +IEnumerable<string> scopes = null
          +string osxKeyChainSuffix = null
          +string preferredDomain = null, 
          +bool verifyPersistence = false
      }
```

PR #22, opened 2022-04-11
Refactored common utilities into its own class

shalinikhare27:
# What
Refactored (not 100% new code) utility code into its own classes - 
MsalHttpClientFactoryAdapter for providing MSAL httpclient.

TokenResultExtensions for augmenting tokenResult with auth type.

TaskExecutor to abstract timeout handling.

# Why
Each of these were individual methods in the `TokenFetcherPublicClient` class but used commonly between auth flows. Since we are separating auth flows into its own class, separating them them will enable us to use them commonly.

# Things to note
TaskExecutor - `CompleteWithin` method is used bythe `GetTokenNormalFlowAsync` method and during device code flow. Which means we use this method no matter which auth flow we try to execute. I moved it to its own class and gave it `logger` and `List<Exception> errorsList` as well so that we can match the usage just as before.

# Test
Nothing to test so far. This is just a refactor.

# Class diagram
The class diagram below gives a representation on how these classes are used. For brevity, I have only represented `AuthFlowBroker` but you could have any auth flow and the flow will look the same more or less.

```mermaid
classDiagram
      IAuthFlows <|-- AuthFlows
      IAuthFlows <|-- AuthFlowBroker
      AuthFlowBroker*-- PCACache
      AuthFlowBroker*-- AccountsProvider
      AuthFlowBroker*-- PCAWrapper
      AuthFlowBroker*-- TaskExecutor
     AuthFlowBroker*-- MsalHttpClientFactoryAdaptor
     AuthFlowBroker*-- TokenResultExtensions
      IAuthFlows : +Task<TokenResult> GetTokenAsync()
      class AuthFlows{
          +ILogger logger
          +Guid resourceId
          +Guid clientId
          +Guid tenantId
          +bool verifyPersistence = false
      }
      class AuthFlowBroker{
          +ILogger logger
          +Guid clientId
          +Guid tenantId
          +bool verifyPersistence = false
      }
```

### DIFF
--- a/src/MSALWrapper/MsalHttpClientFactoryAdaptor.cs
+++ b/src/MSALWrapper/MsalHttpClientFactoryAdaptor.cs
@@ -16,7 +16,6 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlows
         /// The get http client.
         /// </summary>
         /// <returns>An instance of <see cref="HttpClient"/>.</returns>
-
         public HttpClient GetHttpClient()
         {
             // MSAL calls this method each time it wants to use an HTTP client.
--- a/src/MSALWrapper/MsalHttpClientFactoryAdaptor.cs
+++ b/src/MSALWrapper/MsalHttpClientFactoryAdaptor.cs
@@ -1,0 +1,33 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.MSALWrapper.AuthFlows
+{
+    using System.Net.Http;
+    using System.Net.Http.Headers;
+    using Microsoft.Identity.Client;
+
+    /// <summary>
+    /// The msal http client factory adaptor.
+    /// </summary>
+    internal class MsalHttpClientFactoryAdaptor : IMsalHttpClientFactory
+    {
+        /// <summary>
+        /// The get http client.
+        /// </summary>
+        /// <returns>The <see cref="HttpClient"/>.</returns>
+        public HttpClient GetHttpClient()
+        {
+            // MSAL calls this method each time it wants to use an HTTP client.
+            // We ensure we only create a single instance to avoid socket exhaustion.
+            HttpClientHandler handler = new HttpClientHandler();
+            var client = new HttpClient(handler);
+            client.DefaultRequestHeaders.CacheControl = new CacheControlHeaderValue
+            {
+                NoCache = true,
+            };
+
+            return client;
+        }
+    }
+}
--- a/src/MSALWrapper/MsalHttpClientFactoryAdaptor.cs
+++ b/src/MSALWrapper/MsalHttpClientFactoryAdaptor.cs
@@ -15,7 +15,8 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlows
         /// <summary>
         /// The get http client.
         /// </summary>
-        /// <returns>The <see cref="HttpClient"/>.</returns>
+        /// <returns>An instance of <see cref="HttpClient"/>.</returns>
+
         public HttpClient GetHttpClient()
         {
             // MSAL calls this method each time it wants to use an HTTP client.
--- a/src/MSALWrapper/MsalHttpClientFactoryAdaptor.cs
+++ b/src/MSALWrapper/MsalHttpClientFactoryAdaptor.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlows
     internal class MsalHttpClientFactoryAdaptor : IMsalHttpClientFactory
     {
         /// <summary>
-        /// The get http client.
+        /// Gets the msal http client.
         /// </summary>
         /// <returns>An instance of <see cref="HttpClient"/>.</returns>
         public HttpClient GetHttpClient()
--- a/src/MSALWrapper/TaskExecutor.cs
+++ b/src/MSALWrapper/TaskExecutor.cs
@@ -1,0 +1,46 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.MSALWrapper.AuthFlows
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Extensions.Logging;
+
+    /// <summary>
+    /// The task executor.
+    /// </summary>
+    internal class TaskExecutor
+    {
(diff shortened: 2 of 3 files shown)


src/MSALWrapper/TaskExecutor.cs
```diff
@@ -0,0 +1,46 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Extensions.Logging;
+
+    /// <summary>
+    /// The task executor.
+    /// </summary>
+    internal class TaskExecutor
```
kyle-rader, line 15: Can this be a static class? (i.e. you should not be able to create an instance of it)

src/MSALWrapper/TaskExecutor.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        /// <summary>
+        /// The complete within.
+        /// </summary>
+        /// <typeparam name="T">The type.</typeparam>
+        /// <param name="timeout">The timeout.</param>
+        /// <param name="taskName">The task name.</param>
+        /// <param name="getTask">The get task.</param>
```
kyle-rader, line 23:
```suggestion
        /// <param name="getTask">A function that return the task you want to complete within the given timeout.</param>

```

src/MSALWrapper/TaskExecutor.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <param name="errorsList">The errors list.</param>
+        /// <param name="logger">The logger.</param>
+        /// <returns>The <see cref="Task"/>.</returns>
+        internal static async Task<T> CompleteWithin<T>(TimeSpan timeout, string taskName, Func<CancellationToken, Task<T>> getTask, List<Exception> errorsList = null, ILogger logger = null)
```
kyle-rader, line 27: Why make these optional? Anytime you are writing a new interface, it's typically better to avoid using default values, especially nulls, as this makes it far easier for null ref exceptions to creep in. (It also complicates the signature). I think it's more straight forward here to say we always need those arguments.

kyle-rader: This is also just my preference, but anytime I need a logger in a method, because it's such a basic part of functionality, I usually pass the ILogger as the first argument, but again this is just my preference, and is hopefully somewhat common throughout the code. (My only real reason for changing here would be for consistency).

src/MSALWrapper/TokenResultExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Microsoft.Authentication.MSALWrapper.AuthFlows
+{
+    /// <summary>
+    /// The token result extensions.
+    /// </summary>
+    internal static class TokenResultExtensions
+    {
+        /// <summary>
+        /// The set authentication type.
```
kyle-rader, line 12:
```suggestion
        /// Set the AuthType property to the given authType if the token result is not null. 

```

src/MSALWrapper/MsalHttpClientFactoryAdaptor.cs
```diff
@@ -0,0 +1,33 @@
...
+    using System.Net.Http.Headers;
+    using Microsoft.Identity.Client;
+
+    /// <summary>
+    /// The msal http client factory adaptor.
+    /// </summary>
+    internal class MsalHttpClientFactoryAdaptor : IMsalHttpClientFactory
+    {
+        /// <summary>
+        /// The get http client.
```
kyle-rader, line 16: Update Summary here.

src/MSALWrapper/MsalHttpClientFactoryAdaptor.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// </summary>
+        /// <returns>The <see cref="HttpClient"/>.</returns>
```
kyle-rader, line 18:
```suggestion
        /// <returns>An instance of <see cref="HttpClient"/>.</returns>

```

src/MSALWrapper/TokenResultExtensions.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.MSALWrapper.AuthFlows
+{
+    /// <summary>
+    /// The token result extensions.
```
kyle-rader, line 7:
```suggestion
    /// Extension methods to the <see cref="TokenResult"/> class.

```

kyle-rader (changes requested):
Double check the XML comments all up given this MSALWrapper library is likely to be used by more than just us still.

src/MSALWrapper/TaskExecutor.cs
shalinikhare27: Yes, we should be able to.

src/MSALWrapper/TaskExecutor.cs
shalinikhare27: I agree with you here. I made the `logger` and `errorList` not nullable. I also rearranged `logger` to show first. 

I wanted to use `CompleteWithin` method for my main, first stop after CLI is invoked class to have a global timeout - AuthFlows (for reference please see this [file](https://github.com/AzureAD/microsoft-authentication-cli/pull/19/files#diff-8176dc989cffe851b038282da9231b77737a3a4d1293c2fddcedf26b66a12792) - lines 70-77

shalinikhare27: Meaning you agree with my latest changes where `errorList` is not optional?

kyle-rader (changes requested):
Great updates!

kyle-rader (approved):

reillysiemens (approved):

src/MSALWrapper/MsalHttpClientFactoryAdaptor.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Microsoft.Authentication.MSALWrapper.AuthFlows
+{
+    using System.Net.Http;
+    using System.Net.Http.Headers;
+    using Microsoft.Identity.Client;
+
+    /// <summary>
+    /// The msal http client factory adaptor.
+    /// </summary>
+    internal class MsalHttpClientFactoryAdaptor : IMsalHttpClientFactory
```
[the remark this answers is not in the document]
shalinikhare27: Good catch, Tucker! I ported the code over and I see it used interchangeably as Msal and MSAL. We can get it cleaned up if needed in my future PR.

PR #26, opened 2022-04-14
User/kyrader/auth broker suggestions

kyle-rader:
@shalinikhare27 and I paired on some updates to this addition of the `AuthFlowBroker` class so that the tests can inject the mocked PCAWrapper directly into the `AuthFlowBroker` during test setup. 

The changes to PCACache let us fully encapsulate cache setup within the PCAWrapper, in a backwards compatible way. Now the tests never create an actual PCA making them runnable under net6.0 again.


shalinikhare27 (approved):
Kyle and I paired on updating AuthFlowBroker. Approving as I was part of the process.

reillysiemens (approved):

PR #30, opened 2022-04-18
Supress XML Doc requirements in MSALWrapper.Test project.

kyle-rader:
Before moving to GitHub we enabled a default set of StyleCop rules. To adhere to these rules as breaking changes a lot of small mechanical changes were made to satisfy the rules. The XML Docs comment rules make a lot of sense and help keep us honest when writing XML docs for the Library portions of the code. 

The Test projects however are not getting much value and this rules slows down writing tests IMO.


goagain (approved):
Agree. The current comments in Test project were meaningless. Writing them makes me headache. Should we do same thing for other test projects?

src/MSALWrapper.Test/Supressions.cs
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
```
shalinikhare27, line 2: Instead of creating a new Supression.cs file, we can use the `GlobalSupression.cs` file under the `Shared` folder. You can suppress rules for a specific project.

kyle-rader: Ah I avoided adding to `GlobalSuppression.cs` because it is named GlobalSuppression and implies that it applies to all projects. All our projects currently reference this so If I just add the lines to it I'm sure it would apply to all projects, which I don't want.

If there is a syntax to use here to scope it to one project that's good to know, but I still don't think we should put project specific overrides into the global file. Having a Suppressions.cs per project for any project level customization is okay IMO.

goagain: I agree with @kyle-rader , I think we can have a different shared `GlobalSuppression.cs` for test projects.

shalinikhare27:
> Agree. The current comments in Test project were meaningless. Writing them makes me headache. Should we do same thing for other test projects?

I vote "yes" here.

src/MSALWrapper.Test/Supressions.cs
shalinikhare27: If it's intentional, I am okay with this. But yes, we could use scope, rather "target" even in the GlobalSupression.cs file to supress rules for a specific project.

shalinikhare27 (approved):

kyle-rader:
> Agree. The current comments in Test project were meaningless. Writing them makes me headache. Should we do same thing for other test projects?

I will do the other test project in it's own PR.

src/MSALWrapper.Test/Supressions.cs
shalinikhare27: @kyle-rader @goagain Worth renaming to `GlobalSupressionTest.cs` file then?

goagain (approved):

src/MSALWrapper.Test/Supressions.cs
goagain: > @kyle-rader @goagain Worth renaming to `GlobalSupressionTest.cs` file then?

It's ok to me.

kyle-rader: Great points! I've moved the new file to the stylecop folder as `GlobalSupressions.Test.cs` - that way it can be re-used in the other test project the same way.

goagain (dismissed):

stylecop/GlobalSupressions.Test.cs
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation.
```
shalinikhare27, line 1: We already have a `Shared` folder with stylecop files. Should we create a folder under MSALWrapper.Test and move test stylecop stuff there instead of keeping it at src level?

shalinikhare27 (approved):

melindajohnson (approved):

stylecop/GlobalSupressions.Test.cs
kyle-rader: Discussed offline - moved the stylecop folder into src.

src/AzureAuth.Test/AzureAuth.Test.csproj
```diff
@@ -23,8 +23,8 @@
 
   <ItemGroup>
     <!-- Stylecop required items -->
-    <AdditionalFiles Include="..\..\stylecop\stylecop.json" Link="stylecop.json" />
-    <Compile Include="..\..\stylecop\GlobalSupressions.cs" Link="GlobalSupressions.cs" />
+    <AdditionalFiles Include="..\stylecop\stylecop.json" Link="stylecop.json" />
+    <Compile Include="..\stylecop\GlobalSuppressions.cs" Link="GlobalSuppressions.cs" />
```
goagain, line 27:
```suggestion
    <Compile Include="..\stylecop\GlobalSuppressions.cs" Link="GlobalSuppressions.cs" />
    <Compile Include="..\stylecop\GlobalSuppressions.Test.cs" Link="GlobalSuppressions.Test.cs" />
```

goagain: Should we change it in this PR or create a new one?

goagain (changes requested):

src/AzureAuth.Test/AzureAuth.Test.csproj
kyle-rader: This change was small enough Shalini and I thought it was okay :)

PR #34, opened 2022-04-21
AuthFlowExecutor that orchestrates different auth flows to get token and collect errors

shalinikhare27:
# What
Added `AuthFlowExecutor` class and its test.

# Why
This is a top level class that runs through a list of auth flows and mainly does 2 things
1. Gets a token
2. Collects error

If the auth flow result or token result is null, we still collect the errors. If the auth flow result is success, we assign the token and break.

# Test
Units tests were added for cases as the following:
1. Single auth flow returns token result.
2. Single auth flow returns null token result.
3. Single auth flow returns null auth flow result.
4. Single auth flow returns null token result with errors.
5. Single auth flow returns null auth flow result with errors.
6. Similar tests were added for two and three auth flows respectively to check the behavior.

### DIFF
--- a/src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
+++ b/src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
@@ -28,6 +28,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         private IServiceProvider serviceProvider;
         private MemoryTarget logTarget;
         private TokenResult tokenResult;
+        private IEnumerable<IAuthFlow> authFlows;
 
         /// <summary>
         /// The test setup.
@@ -40,6 +41,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
             this.logTarget = new MemoryTarget("memory_target");
             loggingConfig.AddTarget(this.logTarget);
             loggingConfig.AddRuleForAllLevels(this.logTarget);
+            this.authFlows = new List<IAuthFlow>();
 
             // Setup Dependency Injection container to provide logger and out class under test (the "subject")
             this.serviceProvider = new ServiceCollection()
@@ -54,6 +56,44 @@ namespace Microsoft.Authentication.MSALWrapper.Test
 
             // Mock successful token result
             this.tokenResult = new TokenResult(new JsonWebToken(TokenResultTest.FakeToken));
+        }
+
+        [Test]
+        public void ConstructorWith_BothNullArgs()
+        {
+            Action authFlowExecutor = () => new AuthFlowExecutor(null, null);
+
+            // Assert
+            authFlowExecutor.Should().Throw<ArgumentNullException>();
+        }
+
+        [Test]
+        public void ConstructorWith_Null_Logger()
+        {
+            Action authFlowExecutor = () => new AuthFlowExecutor(null, this.authFlows);
+
+            // Assert
+            authFlowExecutor.Should().Throw<ArgumentNullException>();
+        }
+
+        [Test]
+        public void ConstructorWith_Null_AuthFlows()
+        {
+            var logger = this.serviceProvider.GetService<ILogger<AuthFlowExecutor>>();
+            Action authFlowExecutor = () => new AuthFlowExecutor(logger, null);
+
+            // Assert
+            authFlowExecutor.Should().Throw<ArgumentNullException>();
+        }
+
+        [Test]
+        public void ConstructorWith_Valid_Arguments()
+        {
+            var logger = this.serviceProvider.GetService<ILogger<AuthFlowExecutor>>();
+            Action authFlowExecutor = () => new AuthFlowExecutor(logger, this.authFlows);
+
+            // Assert
+            authFlowExecutor.Should().NotThrow<ArgumentNullException>();
         }
 
         [Test]
--- a/src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
+++ b/src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
@@ -560,6 +560,52 @@ namespace Microsoft.Authentication.MSALWrapper.Test
             result.Errors[3].Should().BeEquivalentTo(errors3[1]);
         }
 
+        [Test]
+        public async Task HasThreeAuthFlows_Returns_Early_With_TokenResultAndErrors()
+        {
+            var errors1 = new[]
+            {
+                new Exception("Exception 1"),
+            };
+
+            var errors2 = new[]
+            {
+                new Exception("Exception 2"),
+                new Exception("Exception 3"),
+            };
+
+            var errors3 = new[]
+            {
+                new Exception("This is a catastrophic failure. AuthFlow result is null!"),
+            };
+
+            var authFlowResult1 = new AuthFlowResult(null, errors1);
+            var authFlowResult2 = new AuthFlowResult(this.tokenResult, errors2);
+
+            var authFlow1 = new Mock<IAuthFlow>(MockBehavior.Strict);
+            authFlow1.Setup(p => p.GetTokenAsync()).ReturnsAsync(authFlowResult1);
+
+            var authFlow2 = new Mock<IAuthFlow>(MockBehavior.Strict);
+            authFlow2.Setup(p => p.GetTokenAsync()).ReturnsAsync(authFlowResult2);
+
+            var authFlow3 = new Mock<IAuthFlow>(MockBehavior.Strict);
+            authFlow3.Setup(p => p.GetTokenAsync()).ReturnsAsync((AuthFlowResult)null);
+
+            // Act
+            var authFlow = this.Subject(new[] { authFlow1.Object, authFlow2.Object, authFlow3.Object });
+            var result = await authFlow.GetTokenAsync();
+
+            // Assert
+            authFlow1.VerifyAll();
+            authFlow2.VerifyAll();
+            result.TokenResult.Should().Be(this.tokenResult);
+            result.Success.Should().BeTrue();
+            result.Errors.Should().HaveCount(3);
+            result.Errors[0].Should().BeEquivalentTo(errors1[0]);
+            result.Errors[1].Should().BeEquivalentTo(errors2[0]);
+            result.Errors[2].Should().BeEquivalentTo(errors2[1]);
+        }
+
         private AuthFlowExecutor Subject(IEnumerable<IAuthFlow> authFlows)
         {
             var logger = this.serviceProvider.GetService<ILogger<AuthFlowExecutor>>();
--- a/src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
+++ b/src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
@@ -1,0 +1,190 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.MSALWrapper.Test
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using FluentAssertions;
+    using Microsoft.Authentication.MSALWrapper;
+    using Microsoft.Authentication.MSALWrapper.AuthFlow;
+    using Microsoft.Authentication.MSALWrapper.Test;
+    using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Identity.Client;
+    using Microsoft.IdentityModel.JsonWebTokens;
+    using Moq;
+    using NLog.Extensions.Logging;
+    using NLog.Targets;
+    using NUnit.Framework;
+
+    /// <summary>
+    /// The main auth flow test.
+    /// </summary>
+    public class AuthFlowExecutorTest
+    {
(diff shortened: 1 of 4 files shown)


src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -0,0 +1,63 @@
...
+            this.authflows = authFlows;
+        }
+
+        /// <summary>
+        /// Gets the auth flow result.
+        /// </summary>
+        /// <returns>The <see cref="Task"/>.</returns>
+        public async Task<AuthFlowResult> GetTokenAsync()
+        {
+            AuthFlowResult authFlowResult = new AuthFlowResult(null, new List<Exception>());
```
kyle-rader, line 37: Consider naming `authFlowResult` -> `result` and then inside the loop renaming `result` -> `attempt`

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -0,0 +1,63 @@
...
+                if (result == null)
+                {
+                    authFlowResult.Errors.Add(new Exception("This is a catastrophic failure. AuthFlow result is null!"));
+                }
+                else
+                {
+                    foreach (var error in result.Errors)
+                    {
+                        authFlowResult.Errors.Add(error);
+                    }
```
kyle-rader, line 50: Consider moving this logic into a method on the `AuthFlowResult` class.

kyle-rader (changes requested):
Good job enumerating test cases!

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
shalinikhare27: Done in the latest commit.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
shalinikhare27: Renamed.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AuthFlowExecutor"/> class.
+        /// </summary>
+        /// <param name="logger">The logger.</param>
+        /// <param name="authFlows">The list of auth flows.</param>
+        public AuthFlowExecutor(ILogger logger, IEnumerable<IAuthFlow> authFlows)
+        {
+            this.logger = logger;
+            this.authflows = authFlows;
```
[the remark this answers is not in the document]
shalinikhare27: I added 4 tests to check null arguments for the constructor. Do let me know if I can handle it differently.

goagain (approved):

mvanchaa (dismissed):

kyle-rader (changes requested):

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -34,30 +33,28 @@ public AuthFlowExecutor(ILogger logger, IEnumerable<IAuthFlow> authFlows)
...
             foreach (var authFlow in this.authflows)
             {
-                var result = await authFlow.GetTokenAsync();
-                if (result == null)
+                var attempt = await authFlow.GetTokenAsync();
+                if (attempt == null)
                 {
-                    authFlowResult.Errors.Add(new Exception("This is a catastrophic failure. AuthFlow result is null!"));
+                    result.Errors.Add(new Exception("This is a catastrophic failure. AuthFlow result is null!"));
+                    this.logger.LogError($"This is a catastrophic failure. AuthFlow result is null!");
```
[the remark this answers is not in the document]
shalinikhare27: Okay, makes sense. I will change it to LogDebug instead.

src/MSALWrapper/AuthFlow/AuthFlowResult.cs
```diff
@@ -39,5 +39,18 @@ public bool Success
         {
             get { return this.TokenResult != null; }
         }
+
+        /// <summary>
+        /// Adds the errors from the attempted auth flows to the main error list on Auth flow executor.
+        /// </summary>
+        /// <param name="authFlowResult">A <see cref="MSALWrapper.AuthFlowResult"/> that creates a main list of errors after attempting all the auth flows in the list.</param>
+        /// <param name="result">A <see cref="MSALWrapper.AuthFlowResult"/> that collects errors from the attempted auth flows at a time.</param>
+        internal static void AddErrorsToAuthFlowExecutorList(AuthFlowResult authFlowResult, AuthFlowResult result)
```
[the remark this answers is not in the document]
shalinikhare27: I don't think I fully understand here. The current signature takes 2 arguments to iterate over foreach loop and add local errors to main error list.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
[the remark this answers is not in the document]
shalinikhare27: Changed.

mvanchaa (dismissed):

melindajohnson (approved):

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <returns>The <see cref="Task"/>.</returns>
+        public async Task<AuthFlowResult> GetTokenAsync()
+        {
+            AuthFlowResult result = new AuthFlowResult(null, new List<Exception>());
+            foreach (var authFlow in this.authflows)
+            {
+                var attempt = await authFlow.GetTokenAsync();
+                if (attempt == null)
+                {
+                    result.Errors.Add(new Exception("This is a catastrophic failure. AuthFlow result is null!"));
```
reillysiemens, line 42: Is it common in C# to just use a plain `Exception` like that? I only ask because in my experience with other languages it's more common in a situation like this to create one's own exception class.

reillysiemens: If this error bubbles up to the user (which I think it does), will it actually be helpful to them? 🤔

reillysiemens: Also, is it really catastrophic if it's happening in a loop where we might recover?

reillysiemens: Maybe something more like
```csharp
result.Errors.Add(new Exception($"Failed to get an auth token from ${authFlow.Name}"))
```
or, in lieu of `authFlow.Name` the equivalent of `typeof` or something in C#?

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
```diff
@@ -0,0 +1,623 @@
...
+
+        [Test]
+        public void ConstructorWith_Null_AuthFlows()
+        {
+            var logger = this.serviceProvider.GetService<ILogger<AuthFlowExecutor>>();
+            Action authFlowExecutor = () => new AuthFlowExecutor(logger, null);
+
+            // Assert
+            authFlowExecutor.Should().Throw<ArgumentNullException>();
+        }
```
reillysiemens, line 87: We don't appear to accept any combination of nulls here because we do
```csharp
this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
this.authflows = authFlows ?? throw new ArgumentNullException(nameof(authFlows));
```
These tests would maybe be necessary in Python, but can't we crank up the strictness of the compiler and disallow nulls here?

Seems to me like in C# we should be able to rely on the compiler to do this kind of testing and save us from writing the null checking code above. Maybe I have the wrong idea though.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
kyle-rader: +1, It is less common to use `Exception` in this way, and I agree making our own or at least naming the AuthFlow would be nice. There are a few ways to do that, one of them would be adding a Name method to the IAuthFlow interface which I would support. Every AuthFlow must implement it's get token method, and a Name method. Open to other ideas though

src/MSALWrapper/AuthFlow/AuthFlowResult.cs
```diff
@@ -11,11 +11,19 @@ namespace Microsoft.Authentication.MSALWrapper
...
+            : this(null, null)
+        {
+        }
+
         /// <summary>
         /// Initializes a new instance of the <see cref="AuthFlowResult"/> class.
         /// </summary>
         /// <param name="tokenResult">A <see cref="MSALWrapper.TokenResult"/>.</param>
-        /// <param name="errors">A list of errors encountered while getting (or failing to get) the given token result.</param>
+        /// <param name="errors">A list of errors encountered while getting (or failing to get) the given token result. Will initialize a new empty List if null is given.</param>
```
reillysiemens, line 26: Why initialize a new empty `List` if `null` is given instead of just requiring a new empty `List` be given? Could the compiler also save us from `null` here?

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
kyle-rader: I think you have the right idea @reillysiemens but we have to turn on no nulls for the entire project. I think that could be it's own small feature.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
shalinikhare27: I have created a work item and added to shield bugs to create a custom class for this exception. I discussed this with Kyle and wanted to scope this feature work since there's tons of enhancements that we can do. 

It is catastrophic because we get a null AuthFlowResult which is highly unlikely since it's composed of TokenResult and error list. Having a null TokenResult is common but having a null error list is not, so something got to be really off if it did happen. I had unit tests to cover cases when AuthFlowResult itself returns null.

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
shalinikhare27: I do have tests for null checks in the constructor. Kyle wanted me to catch and throw immediately if we get a null logger or auth flows because that means an error on the caller's side.

src/MSALWrapper/AuthFlow/AuthFlowResult.cs
shalinikhare27: `null` is a valid input for the compiler. This code chunk was suggested by @kyle-rader so that anytime we have null, we initialize the errors as an empty list.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
reillysiemens: Scoping that to another PR is totally reasonable, @shalinikhare27. Sounds good to me.

As far as the catastrophe goes, I was thinking that failing **all** auth flows is the real catastrophe and failing an individual one, while troubling, is more of an inconvenience.

@kyle-rader: I think we can get away with a `nameof` for now instead of a `.Name()` method on that interface. That can be implemented now with no real impact on anything else @shalinikhare27 has merged.

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
reillysiemens: If we don't have an issue for disabling nulls project-wide we should do that. I don't know the best way to capture things like this though where the ramification would be "here's a whole slew of tests you can just delete".

reillysiemens: @shalinikhare27: I think you've done the right thing here given the current configuration of our project. I'm going to resolve this so you're not blocked on it. The point I was making is that in the future we should be able to just avoid writing these kinds of tests altogether and rely on the compiler. 😄

src/MSALWrapper/AuthFlow/AuthFlowResult.cs
reillysiemens: Yeah, I think this is actually the same situation I was describing above where we should be able to just rely on the compiler to save us from a `null`. I forgot that you have to turn that on project-wide and couldn't just do it on a per-namespace basis. Another thing that should get noted as a thing that would be resolved by disabling nulls.

kyle-rader: I don't think forcing the caller to also make a new list should be the default path. But I also didn't like having to pass nulls (passing a null is a code smell to me that the signature you want doesn't exist yet).

reillysiemens: Wouldn't a variadic function signature be a more normal choice then? Only call with one argument if that's the one you have?

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
shalinikhare27: Kyle pushed a change for this. :)

reillysiemens (approved):

kyle-rader (approved):
Nice work :)

PR #41, opened 2022-04-22
Device Code Auth Flow

shalinikhare27:
# What
Redesigned the `TokenFetcherPublicClient` into different auth flows. This PR includes an `DeviceCode` that implements `IAuthFlow`.

# Why
These changes are necessary to close out some open bugs and customer scenarios:
- Better unit testing
- Interactive auth timeout fails to cancel the MSAL embedded web view 
- Provide env var to disable all interaction
- Get rid of coupling auth flows into a massive if else loop

# Things to note
I resolved a bug in this PR which is we MUST try silent auth before trying device code. Kyle wanted me to add to the same PR. Here's more context from the bug:
When we use device code to get a Token result, we currently don’t check the cache before trying device code auth flow to get an access token. The fix would be to call silent auth flow (checking the cache) before trying device code auth flow.

# Test
Nothing to test so far. This is just a refactor. We can do some integration tests once I update the CLI in a follow up PR.
Unit tests were added.

# Class diagram

```mermaid
classDiagram
      IAuthFlow <|-- AuthFlow
      IAuthFlow <|-- DeviceCode
      DeviceCode*-- PCACache
      PCAWrapper*-- AccountsProvider
      DeviceCode*-- PCAWrapper
      DeviceCode*-- TaskExecutor
     DeviceCode*-- MsalHttpClientFactoryAdaptor
     DeviceCode*-- TokenResultExtensions
      IAuthFlow : +Task<TokenResult> GetTokenAsync()
      class AuthFlow{
          +ILogger logger
          +IEnumerable<IAuthFlow> authFlows
      }
      class DeviceCode{
          +ILogger logger
          +Guid clientId
          +Guid tenantId
          +bool verifyPersistence = false,
          +string promptHint = null
      }
```

### DIFF
--- a/src/MSALWrapper.Test/AuthFlow/DeviceCodeTest.cs
+++ b/src/MSALWrapper.Test/AuthFlow/DeviceCodeTest.cs
@@ -1,0 +1,150 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.MSALWrapper.Test
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using FluentAssertions;
+    using Microsoft.Authentication.MSALWrapper;
+    using Microsoft.Authentication.MSALWrapper.AuthFlow;
+    using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Identity.Client;
+    using Microsoft.IdentityModel.JsonWebTokens;
+    using Moq;
+    using NLog.Extensions.Logging;
+    using NLog.Targets;
+    using NUnit.Framework;
+
+    internal class DeviceCodeTest
+    {
+        private const string TestUser = "user@microsoft.com";
+
+        // These Guids were randomly generated and do not refer to a real resource or client
+        // as we don't need either for our testing.
+        private static readonly Guid ResourceId = new Guid("6e979987-a7c8-4604-9b37-e51f06f08f1a");
+        private static readonly Guid ClientId = new Guid("5af6def2-05ec-4cab-b9aa-323d75b5df40");
+        private static readonly Guid TenantId = new Guid("8254f6f7-a09f-4752-8bd6-391adc3b912e");
+
+        private string promptHint = "test prompt hint";
+
+        private IServiceProvider serviceProvider;
+        private MemoryTarget logTarget;
+
+        // MSAL Specific Mocks
+        private Mock<IPCAWrapper> pcaWrapperMock;
+        private IEnumerable<string> scopes = new string[] { $"{ResourceId}/.default" };
+        private TokenResult tokenResult;
+
+        /// <summary>
+        /// The test setup.
+        /// </summary>
+        [SetUp]
+        public void Setup()
+        {
+            // Setup in memory logging target with NLog - allows making assertions against what has been logged.
+            var loggingConfig = new NLog.Config.LoggingConfiguration();
+            this.logTarget = new MemoryTarget("memory_target");
+            loggingConfig.AddTarget(this.logTarget);
+            loggingConfig.AddRuleForAllLevels(this.logTarget);
+
+            this.pcaWrapperMock = new Mock<IPCAWrapper>(MockBehavior.Strict);
+
+            // Setup Dependency Injection container to provide logger and out class under test (the "subject")
+            this.serviceProvider = new ServiceCollection()
+             .AddLogging(loggingBuilder =>
+             {
+                 // configure Logging with NLog
+                 loggingBuilder.ClearProviders();
+                 loggingBuilder.SetMinimumLevel(Microsoft.Extensions.Logging.LogLevel.Trace);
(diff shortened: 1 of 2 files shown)


src/MSALWrapper.Test/AuthFlow/DeviceCodeTest.cs
```diff
@@ -0,0 +1,150 @@
...
+            this.tokenResult = new TokenResult(new JsonWebToken(TokenResultTest.FakeToken));
+        }
+
+        /// <summary>
+        /// Get a new instance of the class under test.
+        /// </summary>
+        /// <returns>The <see cref="AuthFlow.DeviceCode"/> registered in the <see cref="Setup"/> method.</returns>
+        public AuthFlow.DeviceCode Subject() => this.serviceProvider.GetService<AuthFlow.DeviceCode>();
+
+        [Test]
```
kyle-rader, line 82: This is the PR where I want to fix the bug in device code flow where we don't first try to do silent auth. The flow here should be almost identical to web, but the call to GetTokenInteractive is replaced with the call for Device Code flow.

kyle-rader (changes requested):

src/MSALWrapper.Test/AuthFlow/DeviceCodeTest.cs
shalinikhare27: Added silent auth in my latest commit.

src/MSALWrapper/AuthFlow/DeviceCode.cs
```diff
@@ -0,0 +1,163 @@
...
+
+        /// <summary>
+        /// Gets the token for a resource.
+        /// </summary>
+        /// <returns>A <see cref="Task"/> of <see cref="TokenResult"/>.</returns>
+        public async Task<AuthFlowResult> GetTokenAsync()
+        {
+            IAccount account = await this.pcaWrapper.TryToGetCachedAccountAsync(this.preferredDomain)
+                ?? null;
+            this.logger.LogDebug($"GetTokenNormalFlowAsync: Using account '{account.Username}'");
```
reillysiemens, line 71: Is this log statement still correct? Should we be referencing `GetTokenNormalFlowAsync`? 🤔

src/MSALWrapper/AuthFlow/DeviceCode.cs
```diff
@@ -0,0 +1,163 @@
...
+                        "Get Token Silent",
+                        (cancellationToken) => this.pcaWrapper.GetTokenSilentAsync(
+                            this.scopes,
+                            account,
+                            cancellationToken),
+                        this.errors)
+                        .ConfigureAwait(false);
+                    tokenResult.SetAuthenticationType(AuthType.Silent);
+
+                    return new AuthFlowResult(tokenResult, this.errors);
```
reillysiemens, line 89: Could you not also just let this be handled by the final `return` by declaring `tokenResult` outside the scope of this series of `try` blocks. It would have to be `null` initially, but that would just match the final `return` we have here.

```csharp
var tokenResult = null;
```

Then, in this place you'd just set
```csharp
tokenResult = await TaskExecutor.CompleteWithin(...);
```

If you did it this way you've have a single `return` statement for this whole method, which I find helps make things easier to keep track of.

```csharp
return new AuthFlowResult(tokenResult, this.errors);
```

melindajohnson (approved):

reillysiemens (changes requested):
Two small questions here, otherwise I think this looks great and would like to ship it soon.

src/MSALWrapper/AuthFlow/DeviceCode.cs
shalinikhare27: Nope, let me update it because we don't completely try the "normal" flow.

src/MSALWrapper/AuthFlow/DeviceCode.cs
reillysiemens: The code is effectively the same. Let's be consistent for now. It's a minor modification we can return to later in another refactor if we want to.

reillysiemens (approved):

kyle-rader (approved):

src/MSALWrapper.Test/AuthFlow/DeviceCodeTest.cs
```diff
@@ -0,0 +1,255 @@
...
+             {
+                 // configure Logging with NLog
+                 loggingBuilder.ClearProviders();
+                 loggingBuilder.SetMinimumLevel(Microsoft.Extensions.Logging.LogLevel.Trace);
+                 loggingBuilder.AddNLog(loggingConfig);
+             })
+             .AddTransient<AuthFlow.DeviceCode>((provider) =>
+             {
+                 var logger = provider.GetService<ILogger<AuthFlow.DeviceCode>>();
+                 return new AuthFlow.DeviceCode(logger, ClientId, TenantId, this.scopes, pcaWrapper: this.pcaWrapperMock.Object, promptHint: this.promptHint);
```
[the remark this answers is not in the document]
shalinikhare27: Will clean it up as a follow up

PR #44, opened 2022-04-22
Connect CLI with new IAuthFlow interfaces for executing auth.

kyle-rader:
# Replace `TokenFetcherPublicClient` with `IAuthFlow`s
The primary goal of this PR is to remove TokenFetcherPublicClient - proving we have decoupled all of it's business logic into more targeted classes with better ownership.

## Overview
* `CommandMain.cs` is a great place to start reviewing. The changes here show the main goal of moving from TokenFetcherPublicCLient to a combination of new classes/interfaces including `IAuthFlow`, `AuthFlowExecutor` (which implemented `IAuthFlow` and `AuthFlowFactory` which takes the main arguments of the CLI and returns a list of `IAuthFlows instances we should try, in order. `AuthFlowExecutor` takes that list and executes on it.
* Removed Example dotnet projects. These project were mainly there as a demonstration of how to use the MSALWrapper library. We don't intend on shipping this externally however so we don't want those examples. The CLI itself is an example of how to use MSAL directly and we hope to upstream or help inspire some of the higher level APIs our wrapper library has directly into MSAL.
* Cache Clearing - The concept of cache clearing is really only applicable to a PCA using device or web flows. WAM (the broker on Windows) maintains it's own internal cache that we can never clear. We had to add one more extension to `IPCAWrapper` to provide everything we need to clear directly in Command main. We could abstract this logic away into PCAWrapper in the future but I'm fine with there in CommandMain for now.

## Testing
Each auth flow is well unit tested. We also manually tested running the CLI in each auth mode from cold and warm cache starts. (This is how we found 1 bug in the logging of the `IAccount` being used in web and device code flows. We TDD'd the fix for that by adding a test for each case.
- [x] Test manually on Windows.
- [x] Test manually on OSX (Arm).
- [x] Test Manually on OSX (Intel).

## Future Thoughts
* Cache Clearing is really a separate operation all together from authenticating and it may make more sense to move that logic into it's own command. Doing this may also be a good time to move authing into it's own command.


reillysiemens (dismissed):

reillysiemens (approved):

shalinikhare27:
Hi Kyle, firstly amazing job here. My only question is that we expect callers to tell us what "auth-mode" they'd like to try and if they do not provide an "auth-mode" we will try, broker, web and devicecode in the order?

shalinikhare27 (approved):

melindajohnson (approved):

kyle-rader:
@shalinikhare27 there is an existing default choice of auth mode defined by the CLI itself. And the "Default" AuthMode is cross-compiled so that on WIn default is broker, and on mac it's web (system web browser).

shalinikhare27:
> @shalinikhare27 there is an existing default choice of auth mode defined by the CLI itself. And the "Default" AuthMode is cross-compiled so that on WIn default is broker, and on mac it's web (system web browser).

Beautiful! I loved the implementation here. :)

reillysiemens:
@kyle-rader, @shalinikhare27: I know its a lot, but could I ask y'all to summarize the changes in `CHANGELOG.md`. Either in this PR or a follow-up PR would be fine.

kyle-rader:
re: @reillysiemens Yes - I will do a follow up changelog update.

Issue #46, opened 2022-04-26
"Latest" install folder isn't always the "latest"


marcelais:
(All version numbers are examples.)  If you install v0.2.0 of the AuthTool and then subsequently install v0.1.0, the 'latest' folder will point at the v0.1.0 version and not the v0.2.0 version.

Using the latest folder is great for not having to update config files when a new version is released, but if that folder can "regress" to an easlier version, it's usefullness is more limited.

reillysiemens:
Thanks for reporting this, @marcelais. Unfortunately, at the moment we don't have any great ideas for a solution that don't involve teaching the installation scripts how to understand SemVer, which we aren't keen on at the moment.

The most likely path forward will be for us to pursue proper package management for the various supported platforms (e.g. `winget` packages, `.deb`/`.rpm` packages, etc.), but we don't have an ETA for that just yet.

For now, I'd like to acknowledge this issue by leaving it open and invite any suggestions for a good cross-platform solution.

kyle-rader:
I think this bug is more important if we want to enable other applications to be able to install the CLI using the scripts. I was just thinking through using the scripts in a library wrapper, and we wouldn't want App 1 to install and use 0.3.0, another app installs 0.2.0 and reset latest back, at least internally this could confuse things.

kyle-rader:
@marcelais , (and team), @reillysiemens  and I have talked through how to change the installation to avoid thrashing on the version `latest` is pointing to.

We think we want to achieve
* The ability to install yourself, and have `azureauth` on the `PATH`. 
* The ability to install a version, and not update the `PATH`
* Avoid any SemVer comparison in the install scripts.

# Versioned installs only (no more `latest`

Main logic:
* clean install X (default): installs X, and replaces current version on PATH with X.
* clean install X (no-path): install X.

## Clean Install 3 (default)
* install verion 3.
* add version 3 to PATH.

## Clean Install 3 (no-path)
* install version 3

## Dirty Install 4 (default) (with 3 on PATH)
* install version 4
* find version 3,
* replace version 3 with 4 on PATH

## Dirty Install 4 (no-path) (with 3 on PATH)
* install version 4

## Dirty Install 2 (default) (with 3 on the PATH)
* install 2
* find version 3,
* replace version 3 with 2 on PATH

## Dirty Install 2 (no-path) (with 3 on PATH)
* install 2

## MonoRepo Scenario:
### Clean Install version X (no-path)
* install version X
[...]

PR #49, opened 2022-04-27
Stop renaming directories, use directory junctions on Windows

### LINKED ISSUE #47: Latest folder is a copy on Windows
The installer creates a "latest" folder which corresponds to the most recent version of the auth tool.  On windows, this folder is a complete copy rather than a link.

While making a symbolic link requires administrator permission, creating a directory junction (`mklink /j latest v0.2.0`) does not and can be used here to avoid having a second copy of the tool.

reillysiemens:
We've run into an issue where we get spurious `Access to the path ... is denied` errors like
```
 ❯ iex "& { $(irm https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/main/install/install.ps1) } -Verbose"VERBOSE: Creating C:\Users\rsiemens\AppData\Local\Programs\AzureAuth
VERBOSE: Downloading https://github.com/AzureAD/microsoft-authentication-cli/releases/download/v0.2.0/azureauth-v0.2.0-win10-x64.zip to C:\Users\rsiemens\AppData\Local\Programs\AzureAuth\azureauth-v0.2.0-win10-x64.zip
VERBOSE: Extracting C:\Users\rsiemens\AppData\Local\Programs\AzureAuth\azureauth-v0.2.0-win10-x64.zip to C:\Users\rsiemens\AppData\Local\Programs\AzureAuth\v0.2.0
Rename-Item:
Line |
  47 |  Rename-Item $extractedDirectory $targetDirectory
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Access to the path 'C:\Users\rsiemens\AppData\Local\Programs\AzureAuth\azureauth-v0.2.0-win10-x64' is denied.
```

Since this is intermittent, our best understanding of the issue right now is that there might be some sort of file/directory lock that is held unnecessarily long between these lines.

https://github.com/AzureAD/microsoft-authentication-cli/blob/230463d93ed6482b76b5a57140d639fab3cb5eca/install/install.ps1#L49-L51

As we don't directly control any locking done by `System.IO.Compression.ZipFile.ExtractToDirectory` the easiest option to solve this seems to be just not renaming directories in the first place.

Instead of being extracted to a directory like `v0.2.0` the application will now be extracted to something like `azureauth-v0.2.0-win10-x64` (or the equivalent on other platforms).

This also resolves #47 by using a directory junction instead of copying the contents of the `.zip` file twice.

### DIFF
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -23,7 +23,6 @@ $azureauthDirectory = if ([string]::IsNullOrEmpty($Env:AZUREAUTH_INSTALL_DIRECTO
     $Env:AZUREAUTH_INSTALL_DIRECTORY
 }
 $extractedDirectory = ([System.IO.Path]::Combine($azureauthDirectory, $releaseName))
-$targetDirectory = ([System.IO.Path]::Combine($azureauthDirectory, $version))
 $latestDirectory = ([System.IO.Path]::Combine($azureauthDirectory, "latest"))
 $zipFile = ([System.IO.Path]::Combine($azureauthDirectory, $releaseFile))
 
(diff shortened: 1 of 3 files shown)


goagain (commented):
What if customers have already used old version and created that folder?

reillysiemens:
Do you mean what if they've already created a `v0.1.0` or `v0.2.0` directory, @goagain? In that case it will just stay there, inert, not doing anything. That's not ideal, but it's not different from what happens if they upgrade from `v0.1.0` to `v0.2.0` as we don't clean up old directories right now.

kyle-rader:
> What if customers have already used old version and created that folder?

Great question @goagain , given this script is pulled from main in the install script line it will "just update". We are not changing the path to the `latest` version, only how it is created given we are still under version 1.0.0 and we know our main initial users are using latest, this is an acceptable breaking change.

goagain:
> Do you mean what if they've already created a `v0.1.0` or `v0.2.0` directory, @goagain? In that case it will just stay there, inert, not doing anything. That's not ideal, but it's not different from what happens if they upgrade from `v0.1.0` to `v0.2.0` as we don't clean up old directories right now.

Oh I see. Before you create the `latest` folder, you test then delete it. That makes sense.

reillysiemens:
Yeah, whether it's a directory or a junction we always just point it to whatever was just requested for installation. As pointed out in #46 there's a subtle bug with that assumption, but that's a separate problem.

Anyway, great question, @goagain. For the scope of this PR and under the current SemVer guidance I think this is probably a safe change.

install/install.ps1
```diff
@@ -39,28 +38,18 @@ if (Test-Path -Path $extractedDirectory) {
     Remove-Item -Force -Recurse $extractedDirectory
 }
 
-if (Test-Path -Path $targetDirectory) {
-    Write-Verbose "Removing pre-existing target directory at ${targetDirectory}"
-    Remove-Item -Force -Recurse $targetDirectory
-}
```
kyle-rader, line 45: Why do we not still need to remove the `$extractDirectory` if it already exists?

reillysiemens: We still do remove the extracted directory. This was removing the `$targetDirectory`, which was `v0.1.0`/`v0.2.0`, etc.

kyle-rader (approved):

goagain (approved):

Issue #24, opened 2022-04-14
Is a prefix for prompt hint needed?


goagain:
Here I think we should build a prompt hint to pass in, so that is always has `AzureAuth CLI` at the beginning. Perhaps a helper method to construct our prompt hint, that we can call here. 

Something like
```C#
private const string PromptHintPrefix = "Azure AUth CLI";
private string GetPromptHint() {
  if (string.IsNullOrEmpty(this.PromptHint) {
    return PromptHintPrefix;
  }
  else
  {
    return $"{PromptHintPrefix}: {this.PromptHint}";
  }
}
```

_Originally posted by @kyle-rader in https://github.com/AzureAD/microsoft-authentication-cli/pull/11#discussion_r849974649_

reillysiemens:
From talking to some of our users it sounds like there's interest from them in better understanding where these authentication prompts are stemming from. One angle is to do what you've suggested here and add a well-known "`PromptHintPrefix`".

@kyle-rader also suggested that we could try to get the display name of the requested Azure resource and lean on that to establish further legitimacy.

PR #51, opened 2022-04-27
Add "AzureAuth" as prefix

goagain:
Added `AzureAuth` as the prefix.
By default, the title bar will show `AzureAuth`.
If option `--prompt-hint` is set as "sample", the title bar will show `AzureAuth: sample`

Related issue: #24

### DIFF
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -380,9 +380,9 @@ Allowed values: [all, web, devicecode]";
                 var scopes = this.Scopes ?? new string[] { $"{this.authSettings.Resource}/.default" };
 
                 string promptHint = PromptHintPrefix;
-                if (string.IsNullOrEmpty(promptHint))
+                if (!string.IsNullOrEmpty(this.authSettings.PromptHint))
                 {
-                    promptHint = $"{PromptHintPrefix}: {promptHint}";
+                    promptHint = $"{PromptHintPrefix}: {this.authSettings.PromptHint}";
                 }
 
                 var authFlows = AuthFlowFactory.Create(
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -38,6 +38,8 @@ namespace Microsoft.Authentication.AzureAuth
         private const string AliasOption = "--alias";
         private const string ConfigOption = "--config";
 
+        private const string PromptHintPrefix = "AzureAuth";
+
 #if PlatformWindows
         private const string AuthModeHelperText = @"Authentication mode. Default: broker.
 You can use any combination of modes with multiple instances of the -m flag.
@@ -110,6 +112,7 @@ Allowed values: [all, web, devicecode]";
 
         /// <summary>
         /// Gets or sets the customized prompt hint text for WAM prompts and web mode.
+        /// By default there is "AzureAuth" as prefix.
         /// </summary>
         [Option(PromptHintOption, "The prompt hint text for WAM prompts and web mode.", CommandOptionType.SingleValue)]
         public string PromptHint { get; set; }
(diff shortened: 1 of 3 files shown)


src/AzureAuth/CommandMain.cs
```diff
@@ -376,14 +379,20 @@ private IAuthFlow AuthFlow()
                 // TODO: Really we need to get rid of Resource
                 var scopes = this.Scopes ?? new string[] { $"{this.authSettings.Resource}/.default" };
 
+                string promptHint = PromptHintPrefix;
+                if (string.IsNullOrEmpty(promptHint))
```
reillysiemens, line 383: If we set `promptHint` to be `PromptHintPrefix`, which is *always* a `const string` we've set beforehand, isn't it impossible for
```csharp
if (string.IsNullOrEmpty(promptHint) { ... }
```
to ever evaluate to `false`?

If so, does this not lead to a prompt hint of `AzureAuth: AzureAuth` every time? 🤔

src/AzureAuth/CommandMain.cs
```diff
@@ -376,14 +379,20 @@ private IAuthFlow AuthFlow()
...
+
                 var authFlows = AuthFlowFactory.Create(
                     this.logger,
                     this.CombinedAuthMode,
                     new Guid(this.authSettings.Client),
                     new Guid(this.authSettings.Tenant),
                     scopes,
                     this.PreferredDomain,
-                    this.authSettings.PromptHint,
+                    promptHint,
```
reillysiemens, line 395: Do we lose the user-configured prompt hint this way?

reillysiemens (changes requested):

src/AzureAuth/CommandMain.cs
reillysiemens: Oh, sorry, I misread it. I think what I meant is this will never be `true`, so we'll never include the users's prompt because this block can't execute. I could also just be confused here though...

goagain: Sorry. I made a mistake here. it should be 
```C#
                string promptHint = PromptHintPrefix;
                if (!string.IsNullOrEmpty(this.authSettings.PromptHint))
                {
                    promptHint = $"{PromptHintPrefix}: {this.authSettings.PromptHint}";
                }
```

goagain: I tried to make my code clear before committing, but made a huge mistake. Sad.

src/AzureAuth/CommandMain.cs
goagain: Same mistake with the upper conversation.

src/AzureAuth/CommandMain.cs
reillysiemens: Nothing to be sorry about, it seems like an easy enough mistake to make and it looks like you've corrected it. 😄 

My follow-up question is, this seems like the sort of thing that might have been caught with a test. Is there a way for us to test this so we don't regress? If so, let's try to add that in as well before we merge. ✅

goagain: I slightly changed the code and added two unit tests.😁

kyle-rader (changes requested):
Overall I support this change :)

src/AzureAuth/CommandMain.cs
```diff
@@ -167,6 +169,23 @@ public Alias TokenFetcherOptions
...
+        /// <returns>the combined value.</returns>
+        public static string GetActualPromptHint(string promptHintOption)
+        {
+            string promptHint = PromptHintPrefix;
+            if (!string.IsNullOrEmpty(promptHintOption))
+            {
+                promptHint = $"{PromptHintPrefix}: {promptHintOption}";
+            }
+
+            return promptHint;
```
kyle-rader, line 186: This is a case where I would support also using multiple return because of how contained the logic is. Also I think it's generally easier for readers if you avoid using `!` if you can. Consider:
```suggestion
            if (string.IsNullOrEmpty(promptHintOption))
            {
                return PromptHintPrefix;
            }
            return $"{PromptHintPrefix}: {promptHintOption}";
```

kyle-rader (changes requested):
Extracting the logic cleans this up a lot! I have a few more requests :)

reillysiemens (approved):

kyle-rader (approved):

PR #58, opened 2022-05-03
Collect auth settings and arg validation status in telemetry.

kyle-rader:
## Problem
Right now we only collect the client/resource/tenant if passed directly on the CLI. This makes calls using a config file all look the same. We want to be able to distinguish the applications being authenticated too. 

## Solution
1 step in the right direction is to ensure we collect the IDs used either from the CLI or config file (or both if different) and if the arg validation passed all up.

## Future work
Also we want to collect the correlation ID from each auth attempt made. WHich will be it's own feature.

### DIFF
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -254,12 +254,17 @@ Allowed values: [all, web, devicecode]";
         {
             if (!this.EvaluateOptions())
             {
+                this.eventData.Add("validargs", false);
                 return 1;
             }
 
+            this.eventData.Add("validargs", true);
             this.eventData.Add("settings_client", this.authSettings.Client);
             this.eventData.Add("settings_tenant", this.authSettings.Resource);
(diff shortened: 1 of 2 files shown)


reillysiemens (approved):

reillysiemens (approved):

src/AzureAuth/CommandMain.cs
```diff
@@ -254,9 +254,19 @@ public int OnExecute()
...
             {
+                this.eventData.Add("validargs", false);
                 return 1;
             }
 
+            this.eventData.Add("validargs", true);
+            this.eventData.Add("settings_client", this.authSettings.Client);
+            this.eventData.Add("settings_tenant", this.authSettings.Resource);
+            this.eventData.Add("settings_resource", this.authSettings.Tenant);
+            this.eventData.Add("settings_prompthint", this.authSettings.PromptHint);
```
kyle-rader, line 265: Gonna extract this into a method in the future with some small clean ups.

goagain (approved):

PR #64, opened 2022-05-09
Fix PowerShell Install script to work on WS2012 and 2016

kyle-rader:
# Support Installation on WinServer2012 and 2016
On Windows Server 2012/16 you have an older version of PowerShell by default. There are 2 incompatibilities right now with our install script and those versions.
* The configured encryption options for TLS doesn't have the `+=` operator.
   * The fix: remove it. In order to download the install script in the first place, you have to have already enabled TLS1.2. 
   
* `New-Item` doesn't take an argument called `'Target`.
   * Instead, shell out to cmd.exe which has the same interface on all OS's for creating directory junctions.

### DIFF
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -54,7 +54,7 @@ if (Test-Path -Path $latestDirectory) {
 
 # We use a directory junction here because not all Windows users will have permissions to create a symlink.
 Write-Verbose "Linking ${latestDirectory} to ${extractedDirectory}"
-cmd.exe /Q /C "mklink /J $latestDirectory $extractedDirectory"
+cmd.exe /Q /C "mklink /J $latestDirectory $extractedDirectory" > $null
 if (!$?) {
     Write-Error "Linking failed!"
 }
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -49,7 +49,8 @@ if (Test-Path -Path $latestDirectory) {
 
 # We use a directory junction here because not all Windows users will have permissions to create a symlink.
 # We create this junction with cmd.exe's mklink because it has a stable interface across all active versions of Windows and Windows Server,
-# while PowerShell's New-Item has breaking changes and doesn't have the -Target param in 4.0 (the default PowerSehll on Win Server 2012).
+# while PowerShell's New-Item has breaking changes and doesn't have the -Target param in 4.0 (the default PowerShell on Win Server 2012).
+
 Write-Verbose "Linking ${latestDirectory} to ${extractedDirectory}"
 cmd.exe /Q /C "mklink /J $latestDirectory $extractedDirectory" > $null
 if (!$?) {
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -55,6 +55,9 @@ if (Test-Path -Path $latestDirectory) {
 # We use a directory junction here because not all Windows users will have permissions to create a symlink.
 Write-Verbose "Linking ${latestDirectory} to ${extractedDirectory}"
 cmd.exe /Q /C "mklink /J $latestDirectory $extractedDirectory"
+if (!$?) {
+    Write-Error "Linking failed!"
+}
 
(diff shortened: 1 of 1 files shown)


goagain (changes requested):
I got error `Expressions are only allowed as the first element of a pipeline.`

My `[Net.ServicePointManager]::SecurityProtocol` is `SystemDefault`.

install/install.ps1
```diff
@@ -31,7 +31,7 @@ $null = New-Item -ItemType Directory -Force -Path $azureauthDirectory
 
 # Without this, System.Net.WebClient.DownloadFile will fail on a client with TLS 1.0/1.1 disabled
 if ([Net.ServicePointManager]::SecurityProtocol.ToString().Split(',').Trim() -notcontains 'Tls12') {
-    [Net.ServicePointManager]::SecurityProtocol += [Net.SecurityProtocolType]::Tls12
+    [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12;
```
reillysiemens, line 34: Was this semicolon important the whole time? 😱

install/install.ps1
```diff
@@ -54,7 +54,10 @@ if (Test-Path -Path $latestDirectory) {
 
 # We use a directory junction here because not all Windows users will have permissions to create a symlink.
 Write-Verbose "Linking ${latestDirectory} to ${extractedDirectory}"
-$null = New-Item -Path $latestDirectory -Target $extractedDirectory -ItemType Junction
+cmd.exe /Q /C "mklink /J $latestDirectory $extractedDirectory" > $null
+if (!$?) {
+    Write-Error "Linking failed!"
+}
```
reillysiemens, line 60: Understanding _why_ we use `cmd.exe` here instead of just using PowerShell would probably be good.

reillysiemens (approved):
I think it would be nice to have a comment explaining why we use `cmd.exe`, but I don't think that's a blocker.

install/install.ps1
goagain, line 60: Would you mind explaining why the old version 
```ps1
$null = New-Item -Path $latestDirectory -Target $extractedDirectory -ItemType Junction
```
does not work?

For the newer version, what is the difference between
```ps1
cmd.exe /Q /C "mklink /J $latestDirectory $extractedDirectory" > $null
```
and
```ps1
mklink /J $latestDirectory $extractedDirectory > null
```

goagain: Same think.

install/install.ps1
kyle-rader: No, not important. The += operator, that's not implemented in PowerShell 4.0.

kyle-rader: I considered removing this, because it should be needed. In order to download this script itself, you have to have used TLS1.2 - that's why I added it to the README code blocks.

install/install.ps1
kyle-rader: I've updated the PR Description to address this

goagain (dismissed):

install/install.ps1
kyle-rader: Removed.

reillysiemens (dismissed):

install/install.ps1
```diff
@@ -53,8 +48,13 @@ if (Test-Path -Path $latestDirectory) {
 }
 
 # We use a directory junction here because not all Windows users will have permissions to create a symlink.
+# We create this junction with cmd.exe's mklink because it has a stable interface across all active versions of Windows and Windows Server,
+# while PowerShell's New-Item has breaking changes and doesn't have the -Target param in 4.0 (the default PowerSehll on Win Server 2012).
```
mvanchaa, line 52: nit
```suggestion
# while PowerShell's New-Item has breaking changes and doesn't have the -Target param in 4.0 (the default PowerShell on Win Server 2012).

```

PR #65, opened 2022-05-11
Return NullTokenResultException when the authflow returns null.

goagain:
# Description
Currently in the `AuthFlowExecutor` we just return a generic exception when the `AuthFlowResult` is `null` -> this is a catastrophic and very edge case for which we could create a custom Exception class. 
# Change
Added a new exception type `NullTokenResultException`. If an auth flow returns `null`, this exception will be returned.

### DIFF
--- a/src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
+++ b/src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
@@ -5,6 +5,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
     using FluentAssertions;
@@ -163,7 +164,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         {
             var errors1 = new[]
             {
-                new Exception(NullAuthFlowResultExceptionMessage),
+                new BadAuthFlowImplementException(NullAuthFlowResultExceptionMessage),
             };
             var authFlow1 = new Mock<IAuthFlow>(MockBehavior.Strict);
             authFlow1.Setup(p => p.GetTokenAsync()).ReturnsAsync((AuthFlowResult)null);
@@ -237,7 +238,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         {
             var errors1 = new[]
             {
-                new Exception(NullAuthFlowResultExceptionMessage),
+                new BadAuthFlowImplementException(NullAuthFlowResultExceptionMessage),
             };
             var authFlowResult = new AuthFlowResult();
 
@@ -265,7 +266,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         {
             var errors1 = new[]
             {
-                new Exception(NullAuthFlowResultExceptionMessage),
+                new BadAuthFlowImplementException(NullAuthFlowResultExceptionMessage),
             };
 
             var errors2 = new[]
@@ -300,7 +301,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         {
             var errors1 = new[]
             {
-                new Exception(NullAuthFlowResultExceptionMessage),
+                new BadAuthFlowImplementException(NullAuthFlowResultExceptionMessage),
             };
 
             var errors2 = new[]
@@ -410,7 +411,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         {
             var expectedError = new[]
             {
-                new Exception(NullAuthFlowResultExceptionMessage),
+                new BadAuthFlowImplementException(NullAuthFlowResultExceptionMessage),
             };
 
             var authFlowResult1 = new AuthFlowResult();
@@ -454,7 +455,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
 
             var errors3 = new[]
             {
-                new Exception(NullAuthFlowResultExceptionMessage),
+                new BadAuthFlowImplementException(NullAuthFlowResultExceptionMessage),
             };
 
             var authFlowResult1 = new AuthFlowResult(null, errors1);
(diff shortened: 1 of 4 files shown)


goagain:
Should we return an error list or throw the exception directly?

src/MSALWrapper/AuthFlow/BadAuthFlowImplementException.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.MSALWrapper.AuthFlow
+{
+    using System;
+
+    /// <summary>
+    /// The bad authflow implementation exception.
```
reillysiemens, line 9: I think we could use a more descriptive summary here. What does this exception signify?

src/MSALWrapper/AuthFlow/BadAuthFlowImplementException.cs
```diff
@@ -0,0 +1,24 @@
...
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.MSALWrapper.AuthFlow
+{
+    using System;
+
+    /// <summary>
+    /// The bad authflow implementation exception.
+    /// </summary>
+    public class BadAuthFlowImplementException : Exception
```
reillysiemens, line 11: I think we need a different name.
1. `Bad` doesn't necessarily capture what's gone wrong here. _Why_ is it bad?
2. `Implement` (which should probably be `Implementation`) also doesn't seem right. I think it's not necessarily that the implementation is bad, but that the auth flow returned `null`.

@kyle-rader can remind us, but I think this just means that the flow failed, not that the implementation was bad.

reillysiemens (changes requested):
I think the naming is the only thing I'm worried about. Otherwise this is great.

It's not clear to me whether we should raise or keep returning like we do now though. What do you think is the right call, @goagain?

src/MSALWrapper/AuthFlow/BadAuthFlowImplementException.cs
goagain: Renamed it to `NullTokenResultException`, which should be more descriptive.

src/MSALWrapper/AuthFlow/BadAuthFlowImplementException.cs
goagain: Added more details, which should help.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -51,7 +51,7 @@ public async Task<AuthFlowResult> GetTokenAsync()
                 if (attempt == null)
                 {
                     var oopsMessage = $"Auth flow '{authFlow.GetType().Name}' returned a null AuthFlowResult.";
-                    result.Errors.Add(new Exception(oopsMessage));
+                    result.Errors.Add(new NullTokenResultException(oopsMessage));
```
varshini-elayappen, line 54: >NullTokenResultException [](http://example.com/codeflow?start=42&length=24)

While we are here - do we know how often we are seeing a null authflowresult failure right now? If we are seeing this error in telemetry - are we able to understand why we are seeing them? do we have enough data to understand the scenario or do we need to log more information here to make it more useful?

reillysiemens: I haven't explored telemetry enough to know whether we see these. @mvanchaa, do you have any insight here? 🤔

mvanchaa: I will take a look at it

mvanchaa: Looks like we never got this issue. I see zero errors in this scenario.

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
```diff
@@ -616,6 +616,29 @@ public async Task HasThreeAuthFlows_Returns_Early_With_TokenResultAndErrors()
             result.Errors.Should().BeEquivalentTo(new[] { errors1[0], errors2[0], errors2[1] });
         }
 
+        [Test]
+        public async Task HasNullResultAuthFlow_Returns_With_BadImplementationException()
```
mvanchaa, line 620: This method needs renaming according to the new Exception.

goagain: Thanks. I removed this unit test because this unit test has been covered by [SingleAuthFlow_Returns_Null_AuthFlowResult](https://github.com/AzureAD/microsoft-authentication-cli/blob/9128e9de96aa2840843fb80daadab9c0978ad724/src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs#L162).

mvanchaa: oops! got it!

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
reillysiemens: That's surprising to me. Does that seem right to you, @kyle-rader? 🤔

goagain: Yeah, ideally this exception should never happen.

reillysiemens: I'm also happy to approve this PR and move this discussion to a different forum.

reillysiemens (approved):
I'm approving this because I think having a custom exception is an improvement on just using `Exception`, but I would like to see us follow up on @varshini-elayappen's question about whether we see this in telemetry and can we make the information more useful.

mvanchaa (approved):

melindajohnson (approved):

PR #70, opened 2022-05-25
Adding Option and Environment variable for AzureAuth Cache file

goagain:
# Summary
Add Environment variable `AZUREAUTH_CACHE_FILE` and option `--cache` to support custom cache file name.

Option `--cache` is only valid on Windows, and its help text will also hide on non-Windows platform.


src/MSALWrapper/PCACache.cs
```diff
@@ -44,7 +44,16 @@ internal PCACache(ILogger logger, Guid tenantId, string osxKeyChainSuffix = null
...
+
+            var azureAuthCacheFile = Environment.GetEnvironmentVariable(Constants.AZUREAUTH_CACHE_FILE);
+            if (string.IsNullOrEmpty(azureAuthCacheFile))
+            {
+                this.cacheFileName = $"msal_{tenantId}.cache";
+            }
+            else
+            {
+                this.cacheFileName = azureAuthCacheFile;
+            }
```
reillysiemens, line 56: I don't always love them, but I think this is one case where a ternary would actually improve things.

```suggestion
            this.cacheFileName = string.IsNullOrEmpty(azureAuthCacheFile) ? $"msal_{tenantId}.cache" : azureAuthCacheFile;
```

reillysiemens: Also, should we be using `string.IsNullOrWhiteSpace()` instead of `string.IsNullOrEmpty()` here? I don't think it would be good for us to allow a filename of `" "`.

reillysiemens (changes requested):
Some small suggestions here, but after my `string.IsNullOrWhiteSpace()` vs `string.IsNullOrEmpty()` question I also think we might want to add some tests to validate this behavior.

src/MSALWrapper.Test/PCAWrapperTest.cs
```diff
@@ -154,6 +156,46 @@ public async Task TryToGetCachedAccountAsync_MultipleAccounts_WithPreferredDomai
             result.Should().BeNull();
         }
 
+        [Test]
+        public void CustomizedCacheFileWithInvalidCharacter()
+        {
+            // Setup
+            string filenameWithInvalidCharacter = "filename_with_invalid_character/";
```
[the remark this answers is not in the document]
goagain: Actually, I tried to use like 
```C#
string filenameWithInvalidCharacter = "filename_with_invalid_character" + Path.GetInvalidFileNameChars().First();
```

However, the unit test still failed. I don't have a MacBook neither a Linux, so for now I have no idea what happened.

kyle-rader (changes requested):
Summarizing our discussion from pair review: The summary right now is a code smell of too much responsibility in the constructor of the PCAWrapper class.

I do love we're unit testing the logic of validating a cache file name - but let's move it somewhere more distinct.

Summary:
* Move logic of building the cache file name into AzureAuth
   * This will let us define the constant in the `EnvVars.cs` file and 
   * make use if `IEnv` from Lasso for the env interaction - making testing easier to do without any side affects.
   * Unit test crafting the cache File name as a single function, it could be in a new class, itr could be a static method somewhere in AzureAuth, happy to discuss more on where an ideal place is.
*  Have PCAWrapper take a required parameter for cache file name.

goagain:
Frankly, with the increasement of parameters, it is becoming complex and dangerous to add a parameter. I believe some refactor is needed. For example, integrated all PCAWrapper's optional parameters into a new class could be a better idea.

goagain:
Also, the current mode is not compile-time safe. For example, in the function `foo(int a, string b)`, when the parameters do not match, the compiler will prompt an error. However, for the function `bar(string a, string b)` with the same parameter types, the compiler cannot give any error message when filling in the wrong parameter order. This makes any changes to the code very careful. Furthermore, because such calls may span multiple modules, there is no guarantee that unit tests can cover such scenarios 100%. One solution is to enforce named parameters for all calls with optional parameters. This avoids problems with the order of function calls, but does not avoid the need to change a large number of classes per modification. Another option is to use fluent interfaces or use aggregate classes. I'm not very experienced in this area, can you give me some advice.

src/AzureAuth/CommandMain.cs
```diff
@@ -159,6 +159,13 @@ public CommandMain(CommandExecuteEventData eventData, ILogger<CommandMain> logge
         [Option(ConfigOption, "The path to a configuration file.", CommandOptionType.SingleValue)]
         public string ConfigFilePath { get; set; }
 
+        /// <summary>
+        /// Gets or sets the cache file name.
+        /// </summary>
+        [LegalFileName]
+        [Option(CacheFileNameOption, "The file name of cache", CommandOptionType.SingleOrNoValue)]
```
kyle-rader, line 166:
```suggestion
        [Option(CacheFileNameOption, "Override the default cache file location.", CommandOptionType.SingleOrNoValue)]

```

src/AzureAuth/CommandMain.cs
```diff
@@ -37,7 +37,7 @@ public class CommandMain
         private const string OutputOption = "--output";
         private const string AliasOption = "--alias";
         private const string ConfigOption = "--config";
-
+        private const string CacheFileNameOption = "--cache-filename";
```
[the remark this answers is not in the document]
kyle-rader: There is also a question about platform for this option. You only have a cache file on Windows.

kyle-rader (changes requested):

src/MSALWrapper/PCACache.cs
```diff
@@ -37,14 +39,21 @@ internal class PCACache
...
         /// <param name="tenantId">The tenant id.</param>
+        /// <param name="cacheFileName">The cache file name.</param>
         /// <param name="osxKeyChainSuffix">The osx key chain suffix.</param>
-        internal PCACache(ILogger logger, Guid tenantId, string osxKeyChainSuffix = null)
+        internal PCACache(ILogger logger, Guid tenantId, string cacheFileName, string osxKeyChainSuffix = null)
         {
             this.logger = logger;
             this.osxKeyChainSuffix = string.IsNullOrWhiteSpace(osxKeyChainSuffix) ? $"{tenantId}" : $"{osxKeyChainSuffix}.{tenantId}";
             string appData = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
             this.cacheDir = Path.Combine(appData, ".IdentityService");
```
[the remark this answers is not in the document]
goagain: `cacheDir` is still used at line 73.

src/MSALWrapper/PCACache.cs
```diff
@@ -37,14 +39,21 @@ internal class PCACache
...
-            this.cacheFileName = $"msal_{tenantId}.cache";
+
+            if (string.IsNullOrWhiteSpace(cacheFileName))
+            {
+                throw new ArgumentNullException($"{nameof(cacheFileName)} should not be null or whitespace.");
+            }
+
+            this.cacheFileName = cacheFileName;
```
kyle-rader, line 56: It's unclear to me how MSAL uses the cache Dir if the env name is a different complete full PATH. We should amke sure we understand how this works. 

I think we might want to actually take the full path for a cache file - and here use Path methods to split the parent dir and file into their separate parts, since that's what we need to give to the cache storage options. 

Does my confusion / worry make sense?

src/AzureAuth/CommandMain.cs
```diff
@@ -291,13 +305,19 @@ private bool ValidateOptions()
                 validOptions = false;
             }
 
+            if (!LegalFileNameChecker.IsValidFilename(this.CacheFileName))
+            {
+                this.logger.LogError($"The {CacheFileNameOption} field or environment varable {nameof(EnvVars.AZUREAUTH_CACHE_FILE)} is invalid");
+                validOptions = false;
+            }
```
kyle-rader, line 312: I think you've added this check (again) here because we might have pulled the filename from the environment - which will skip the `LegalFileNameAttribute` check. I'd rather not have 2 copies of this check, and instead do all the checking in 1 place - here. This way we don't need to add the attribute class at all. We can still customize the error message in the checking logic to say if the it was the option, or the env var that was used and had an invalid filename.

src/MSALWrapper/PCACache.cs
kyle-rader: I'm guessing since we tested setting the fileName arg to a full path in our hotfix, that that actually works, but if this is the case, we should at least comment this awkwardness.

src/AzureAuth/CommandMain.cs
```diff
@@ -159,6 +159,13 @@ public CommandMain(CommandExecuteEventData eventData, ILogger<CommandMain> logge
         [Option(ConfigOption, "The path to a configuration file.", CommandOptionType.SingleValue)]
         public string ConfigFilePath { get; set; }
 
+        /// <summary>
+        /// Gets or sets the cache file name.
+        /// </summary>
+        [LegalFileName]
+        [Option(CacheFileNameOption, "Override the default cache file location.", CommandOptionType.SingleOrNoValue)]
```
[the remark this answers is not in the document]
goagain: I don't think here must have a value. This parameter is from command line, and if it is empty it will be replaced by `msal_{tenant_id}.cache`. Also, it could be replaced by environment variable. If this option is required, all existing command line may not work.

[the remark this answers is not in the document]
kyle-rader: We want to use `SingleValue` because providing the option `--cache` by itself has no meaning - and we want to treat that as bad input. i.e. if you are using `--cache` you must give us a value for it.

src/MSALWrapper/PCACache.cs
goagain: After I read the interfaces from MSAL, I think it may become more complicated if we allow user to use absolute path.  The signature is 
```csharp
public StorageCreationPropertiesBuilder(string cacheFileName, string cacheDirectory)
```
Is it worth to deal with the path in our side?

src/AzureAuth/CommandMain.cs
goagain: That's true. We shouldn't repeat ourselves.

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -384,6 +389,69 @@ public void TestPromptHintPrefixWithoutOption()
...
+        [Test]
+        public void TestCacheFileOptionWithNoParameter()
+        {
+            CommandMain subject = this.serviceProvider.GetService<CommandMain>();
+            subject.Resource = "f0e8d801-3a50-48fd-b2da-6476d6e832a2";
+            subject.Client = "e19f71ed-3b14-448d-9346-9eff9753646b";
+            subject.Tenant = "9f6227ee-3d14-473e-8bed-1281171ef8c9";
+
+            subject.EvaluateOptions().Should().BeTrue();
+            subject.CacheFileName.Should().Be($"msal_{subject.Tenant}.cache");
```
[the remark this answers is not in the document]
goagain: Current unit test should look better.

src/AzureAuth/CommandMain.cs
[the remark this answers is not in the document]
goagain: Sorry I thought `SingleValue` means if there is no such option the program won't run.

src/MSALWrapper/PCACache.cs
kyle-rader: We have to allow the user to give an absolute path (and in the internal version already do)

src/MSALWrapper.Test/AuthFlow/BrokerTest.cs
```diff
@@ -40,6 +40,7 @@ public class BrokerTest
         private Mock<IPCAWrapper> pcaWrapperMock;
         private Mock<IAccount> testAccount;
         private IEnumerable<string> scopes = new string[] { $"{ResourceId}/.default" };
+        private string cacheFilePath = $"msal_{TenantId}.cache";
```
kyle-rader, line 43:
```suggestion
        private string cacheFilePath = $"test_cache_file";

```

src/AzureAuth/CommandMain.cs
```diff
@@ -167,6 +173,31 @@ public Alias TokenFetcherOptions
...
+                if (!string.IsNullOrEmpty(envCacheFile))
+                {
+                    return envCacheFile;
+                }
+
+                // Use default cache file name.
+                return $"msal_{this.authSettings.Tenant}.cache";
+            }
+        }
+
```
kyle-rader, line 200: I dislike that we need a separate function called `Wrapped***` here. Another option would be to implement this `get` function as the get method for the property itself (`CacheFilePath`) and use a private member to back the property. You have to implement `set` as well so that if the CLI parses the option, it can set it. But otherwise the private member will remain null, and you can memoize the env var during the `get` method. Does that make sense?

src/AzureAuth/CommandMain.cs
```diff
@@ -159,6 +159,12 @@ public CommandMain(CommandExecuteEventData eventData, ILogger<CommandMain> logge
         [Option(ConfigOption, "The path to a configuration file.", CommandOptionType.SingleValue)]
         public string ConfigFilePath { get; set; }
 
+        /// <summary>
+        /// Gets or sets the cache file name.
+        /// </summary>
+        [Option(CacheOption, "Override the default cache file location.", CommandOptionType.SingleValue)]
```
kyle-rader, line 165: Given the cache file is Windows Only, but we need the default value to compile still for Mac. I think we can conditionally set the option to hide this flag from the help text on Mac. CommandLineUtils has a way to do this, and we have a build constant you can use to set a bool if on windows, or not.

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -384,6 +389,142 @@ public void TestPromptHintPrefixWithoutOption()
...
+            CommandMain subject = this.serviceProvider.GetService<CommandMain>();
+            subject.Resource = "f0e8d801-3a50-48fd-b2da-6476d6e832a2";
+            subject.Client = "e19f71ed-3b14-448d-9346-9eff9753646b";
+            subject.Tenant = "9f6227ee-3d14-473e-8bed-1281171ef8c9";
+
+            string path = "/usr/bin/relative.cache";
+            subject.CacheFilePath = path;
+            subject.EvaluateOptions().Should().BeTrue();
+            subject.WrappedCacheFilePath.Should().Be(path);
+        }
```
kyle-rader, line 526: The cache file is never created on Mac/Linux so I don't think we need this test.

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -384,6 +389,142 @@ public void TestPromptHintPrefixWithoutOption()
...
+            CommandMain subject = this.serviceProvider.GetService<CommandMain>();
+            subject.Resource = "f0e8d801-3a50-48fd-b2da-6476d6e832a2";
+            subject.Client = "e19f71ed-3b14-448d-9346-9eff9753646b";
+            subject.Tenant = "9f6227ee-3d14-473e-8bed-1281171ef8c9";
+
+            string path = "../test/relative.cache";
+            subject.CacheFilePath = path;
+            subject.EvaluateOptions().Should().BeTrue();
+            subject.WrappedCacheFilePath.Should().Be(path);
+        }
```
kyle-rader, line 492: I think this test should assert failure and we force the caller to give us an absolute path.

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -384,6 +389,142 @@ public void TestPromptHintPrefixWithoutOption()
...
+        [Test]
+        public void TestCacheFileOptionWithNoParameter()
+        {
+            CommandMain subject = this.serviceProvider.GetService<CommandMain>();
+            subject.Resource = "f0e8d801-3a50-48fd-b2da-6476d6e832a2";
+            subject.Client = "e19f71ed-3b14-448d-9346-9eff9753646b";
+            subject.Tenant = "9f6227ee-3d14-473e-8bed-1281171ef8c9";
+
+            subject.EvaluateOptions().Should().BeTrue();
+            subject.WrappedCacheFilePath.Should().Be($"msal_{subject.Tenant}.cache");
```
kyle-rader, line 474: This is where changing from WrappedCacheFilePath to just the CacheFilePath property should be asserted that it's the full absolute path. Given it will differ by user, the assertion will have to be a little bit more flexible but I think it can assert that it is a Rooted path, and ends with this filename.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Environment variable `AZUREAUTH_CACHE_FILE` and option `--cache` to support custom cache file name.
```
kyle-rader, line 9:
```suggestion
- Environment variable `AZUREAUTH_CACHE_FILE` and option `--cache` to support a custom cache location on Windows.
```

goagain:
@reillysiemens @kyle-rader Thank you for giving so many advices. I have made changes to improve the code quality.
Would you mind reviewing my pull request again?

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -384,6 +388,119 @@ public void TestPromptHintPrefixWithoutOption()
...
+        {
+            string filenameFromEnv = "C:\\test\\absolute_from_env.cache";
+            this.envMock.Setup(env => env.Get("AZUREAUTH_CACHE_FILE")).Returns(filenameFromEnv);
+
+            CommandMain subject = this.serviceProvider.GetService<CommandMain>();
+            subject.Resource = "f0e8d801-3a50-48fd-b2da-6476d6e832a2";
+            subject.Client = "e19f71ed-3b14-448d-9346-9eff9753646b";
+            subject.Tenant = "9f6227ee-3d14-473e-8bed-1281171ef8c9";
+
+            subject.EvaluateOptions().Should().BeTrue();
```
kyle-rader, line 423: Are we missing an assertion here that the CaceFile property on the `subject` is equal to the expect cache file?

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -384,6 +388,119 @@ public void TestPromptHintPrefixWithoutOption()
...
+        }
+
+        /// <summary>
+        /// The test to evaluate a relative cache file name in enviroment variables.
+        /// </summary>
+        [Test]
+        [Platform("Win")] // Only valid on Windows
+        public void TestCacheFileOptionWithNormalFilePathFromEnv()
+        {
+            string filenameFromEnv = "C:\\test\\absolute_from_env.cache";
```
kyle-rader, line 415: This test description here claims to be a relative path, but the path used is still absolute.

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -384,6 +388,119 @@ public void TestPromptHintPrefixWithoutOption()
...
+            string appData = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+            string absolutePath = Path.Combine(appData, ".IdentityService", $"msal_{subject.Tenant}.cache");
+            string expected = absolutePath;
+
+            subject.CacheFilePath.Should().Be(expected);
+        }
+
+        /// <summary>
+        /// The test to evaluate a relative cache path.
+        /// </summary>
```
kyle-rader, line 471: The <Summary> blocks in this test file should no be required, so I would only add them if they are adding more description than is visible from the test name itself.

kyle-rader (changes requested):
I think this is nearly done, just 1 small question in the tests.

src/AzureAuth.Test/CommandMainTest.cs
goagain: 😅 I missed that.

src/AzureAuth.Test/CommandMainTest.cs
goagain: Thank you, I changed.

src/AzureAuth.Test/CommandMainTest.cs
reillysiemens: I think @kyle-rader means to remove the summary.

src/AzureAuth/CommandMain.cs
```diff
@@ -159,6 +162,40 @@ public CommandMain(CommandExecuteEventData eventData, ILogger<CommandMain> logge
...
+                string envCacheFile = this.env.Get(EnvVars.AZUREAUTH_CACHE_FILE);
+                if (!string.IsNullOrEmpty(envCacheFile))
+                {
+                    return envCacheFile;
+                }
+
+                // Use default cache file path.
+                string appData = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+                string absolutePath = this.fileSystem.Path.Combine(appData, ".IdentityService", $"msal_{this.authSettings.Tenant}.cache");
+                return absolutePath;
```
kyle-rader, line 190: Not blocking, but generally I'd prefer to see logic like this extracted into a method, and then used only to memoize (cache) the result using a private member to back this Property. It's a small performance gain to only access the Env once. We are going to access this getter at least twice - once to validate, and once to use. But the gain here is small enough I don't want to block on this now.

src/AzureAuth/EnvVars.cs
```diff
@@ -18,6 +18,11 @@ internal static class EnvVars
         /// </summary>
         public static readonly string Config = $"{EnvVarPrefix}_CONFIG";
 
+        /// <summary>
+        /// The name of an environment variable used to override the cache file path.
+        /// </summary>
+        public static readonly string AZUREAUTH_CACHE_FILE = $"{EnvVarPrefix}_CACHE_FILE";
```
kyle-rader, line 24: Concistency - The `AZUREAUTH_CONFIG` value is for a file but does not use `FILE` in the name. So I'd prefer we name the env var: `AZUREAUTH_CACHE` to be consistent.

kyle-rader (changes requested):

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -384,6 +388,121 @@ public void TestPromptHintPrefixWithoutOption()
...
+        [Test]
+        [Platform("Win")] // Only valid on Windows
+        public void TestCacheFileOptionWithRelativePath()
+        {
+            CommandMain subject = this.serviceProvider.GetService<CommandMain>();
+            subject.Resource = "f0e8d801-3a50-48fd-b2da-6476d6e832a2";
+            subject.Client = "e19f71ed-3b14-448d-9346-9eff9753646b";
+            subject.Tenant = "9f6227ee-3d14-473e-8bed-1281171ef8c9";
+
+            string path = "../test/relative.cache";
```
reillysiemens, line 483: This says it's a Windows only test, but it uses non-Windows path separators. Is that correct? :thinking:

src/AzureAuth/EnvVars.cs
goagain: Changed the Env Var to "AZUREAUTH_CACHE", also changed the variable name to Cache for consistency.

src/AzureAuth.Test/CommandMainTest.cs
goagain: That's right. I changed it to Windows-style.

src/AzureAuth.Test/CommandMainTest.cs
goagain: The summary is required here. I added more description instead.

kyle-rader (approved):
Nice work on this!

reillysiemens (approved):

PR #75, opened 2022-05-31
Check out code to use scripts

reillysiemens:
Without checking out the code it isn't possible to run `./bin/version.py` because the file doesn't exist yet.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,6 +19,8 @@ jobs:
   validate:
     runs-on: ubuntu-latest
     steps:
+    - name: Checkout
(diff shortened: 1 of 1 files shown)


goagain:
Is this only for Linux?

goagain:
> 

Oh, I see. It only validate the version number, not for building.

reillysiemens:
@kyle-rader: I'm not yet convinced we need the extra speed boost from doing a shallow clone. I'll keep that in my back pocket as an improvement for later. Or, if you like, you could open an issue for it.

@goagain: No, this isn't only for Linux. It's just validating that the release version is valid before proceeding to next steps.

PR #83, opened 2022-06-02
Upgrade .NET to .NET 6.0

goagain:
Upgrade .NET to .NET 6.0

### DIFF
--- a/src/MSALWrapper/MSALWrapper.csproj
+++ b/src/MSALWrapper/MSALWrapper.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <!-- Package Naming, Building, & Versioning -->
     <PackageId>Microsoft.Authentication.MSALWrapper</PackageId>
-    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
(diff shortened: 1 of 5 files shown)


src/MSALWrapper/MSALWrapper.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <!-- Package Naming, Building, & Versioning -->
     <PackageId>Microsoft.Authentication.MSALWrapper</PackageId>
-    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>net6.0</TargetFrameworks>
```
kyle-rader, line 6: Since this is a library I'd prefer we not change this unless we have a good reason to. net6.0 is capable of consuming both netstandard2.0 and 2.1. We could change to just netstandard2.1.
Basically libraries should always target netstandard versions - which define the dotnet interface they require, and then any number of actual target frameworks for applications can consume it.

kyle-rader (changes requested):
I think this is most of the code changes needed for this update - but, as you called out, we need to test on mac, and Windows (all windows versions) thouroughly before merging this. All auth modes, all states.

PR #84, opened 2022-06-03
Add certificates to the user store

reillysiemens:
For unknown reasons the certificates we add with `certutil` seem to be added to the local machine certificate store rather than the current user store and it doesn't appear as though `EsrpClient.exe` has any idea what to do in that case, so it just fails...

This PR should resolve that by using the `-user` flag (although that usually results in a UAC prompt, so I'm not 💯% sure this will work).


kyle-rader (approved):
I love only being able to test in prod (ty GitHub Actions)

PR #88, opened 2022-06-03
Add status badges

reillysiemens:
This PR just adds some simple status badges using [shields.io](https://shields.io/). This is similar to what [GCM](https://github.com/GitCredentialManager/git-credential-manager) does, but with a little more flair for quick visual identifiers on things you might want to know when choosing a new tool (are tests passing, which version are they on, what's the license, etc.).

Note that the release badge uses SemVer and includes pre-releases, so we never need to update it.

## Before
readme-before-badge

## After
readme-after-badge


melindajohnson (approved):

goagain (approved):
That's so pretty!

PR #91, opened 2022-06-06
Change version scheme

reillysiemens:
It turns out when we attempted to release `v0.3.1` I forgot that we needed the `-p:Version=` flag to `dotnet publish`. Using a `v` prefix makes that more difficult, so we've just dropped that and decided to use `0.3.1` instead of `v0.3.1`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,7 @@ on:
         description: 'Version'
         required: true
         type: string
-      # We use SemVer, anything before v1.0.0 is a pre-release, but this could also include versions like v1.1.0-beta.
+      # We use SemVer, anything before 1.0.0 is a pre-release, but this could also include versions like 1.1.0-beta.
       prerelease:
         description: 'Prerelease'
         required: true
@@ -66,7 +66,7 @@ jobs:
     - name: Test
       run: dotnet test --no-restore --configuration release
     - name: Build artifacts
-      run: dotnet publish src/AzureAuth/AzureAuth.csproj --configuration release --self-contained true --runtime ${{ matrix.runtime }} --output dist/${{ matrix.runtime }}
+      run: dotnet publish src/AzureAuth/AzureAuth.csproj -p:Version=${{ github.event.inputs.version }} --configuration release --self-contained true --runtime ${{ matrix.runtime }} --output dist/${{ matrix.runtime }}
     - name: Upload artifacts
       uses: actions/upload-artifact@v3
       with:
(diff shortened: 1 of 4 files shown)


goagain (approved):
That means our release will be like `azureauth-0.3.1-osx-x64.tar.gz` without the letter 'v'.

goagain:
Can we keep the prefix `v` and use a transitive argument to replace it?
For example:
`{{github.inputs.version}}` = `0.3.1`
`$WrappedVersion`=`v{{github.inputs.version}}`

then we could have:
`-p:Version=${{ github.event.inputs.version }}` and
`azureauth-${{WrappedVersion}}-win10-x64.zip`

kyle-rader (approved):

melindajohnson (approved):

kyle-rader:
Oops, sorry @goagain I just saw your last comment. Overall I've been arguing that we should not have the `v` in the first place because:
1. it is not valid SemVer
2. it complicates scripts that are trying to read a valid semver from us
3. We claim to use Sem Ver
4. We don't have a way, at least not now to bake the matching version with v into out dotnet binary without other downstream changes.
5. It's common yes, but not standard and it's easy to forget. I think it is better to have something that is as common, but does not require you remember it.
6. When we talk about releases and versions, you don't say "V zero dot three dot zero" generally we say "zero dot three do zero".

The main argument for a `v` prefix is to be able to easily filter the git tags. But we don't use tags for anything else at present and I don't think that one ability is worth the cognitive overhead for anyone using AzureAuth. We can easily filter tags ourselves in other ways _if and when we need to_ - but solving fro that is solving for a problem that does not exist at the expense of the current experience.

goagain:
> Oops, sorry @goagain I just saw your last comment. Overall I've been arguing that we should not have the `v` in the first place because:
> 
> 1. it is not valid SemVer
> 2. it complicates scripts that are trying to read a valid semver from us
> 3. We claim to use Sem Ver
> 4. We don't have a way, at least not now to bake the matching version with v into out dotnet binary without other downstream changes.
> 5. It's common yes, but not standard and it's easy to forget. I think it is better to have something that is as common, but does not require you remember it.
> 6. When we talk about releases and versions, you don't say "V zero dot three dot zero" generally we say "zero dot three do zero".
> 
> The main argument for a `v` prefix is to be able to easily filter the git tags. But we don't use tags for anything else at present and I don't think that one ability is worth the cognitive overhead for anyone using AzureAuth. We can easily filter tags ourselves in other ways _if and when we need to_ - but solving fro that is solving for a problem that does not exist at the expense of the current experience.

Thank you @kyle-rader for explaining. I remember it says `less is more`.

PR #92, opened 2022-06-07
Fix windows archive errors

reillysiemens:
Now we can _actually_ release `0.3.1`. 🎉


kyle-rader:
Notable change here is that we were zipping the binaries directly into the zip target, but our installation scripts expect the zip archive to have the named, versioned directory in it.

kyle-rader (approved):

shalinikhare27 (approved):

goagain (approved):

PR #95, opened 2022-06-09
Revert "Stop azureauth before install"

reillysiemens:
Reverts AzureAD/microsoft-authentication-cli#93

PR #93, opened 2022-06-09
Stop azureauth before install

reillysiemens:
Installation can fail if azureauth is running at the same time. This will attempt to stop azureauth first.

PR #96, opened 2022-06-09
Detour installation script

goagain:
# Summary
Changing installation is dangerous because it is out of version control. We made mistakes in #93 and #95.

Detouring the installation script to release version can guarantee our customers always use a release version and avoid this problem.

# Questions

The cost is we cannot deliver the installation script as fast as we can. But if our versions are becoming more and more stable, I think that cost is worthy.


reillysiemens (approved):

kyle-rader:
I have tested on Windows10. @goagain can you also do a test on Win Server, just to really check that string interpolation works in PowerShell 4/5. My Trust in PowerShell across all it's versions is gone.

goagain:
> 

@kyle-rader, I Tested on Windows 2012 and it works on PowerShell 4.0

kyle-rader (approved):

PR #99, opened 2022-06-09
Stop azureauth before install

reillysiemens:
Installation can fail if azureauth is running at the same time. This will attempt to stop azureauth first. We tried to add this in #93, but had to revert it in #95. Using a `try`/`catch` block now prevents this from triggering `$ErrorActionPreference='Stop'` in PowerShell 5.

### DIFF
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -33,6 +33,14 @@ Write-Verbose "Downloading ${releaseUrl} to ${zipFile}"
 $client = New-Object System.Net.WebClient
 $client.DownloadFile($releaseUrl, $zipFile)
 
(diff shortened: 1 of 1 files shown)


kyle-rader (approved):

goagain (approved):
It works on Win2012, but I am still wondering if there is a more graceful way to implement. Using an empty try-catch block seems a little bit ugly. 😅

varshini-elayappen (approved):
:shipit:

reillysiemens:
@goagain: I absolutely agree that using an empty `catch` block here is a little ugly. If you find a better solution I would be happy to accept it in a PR. 😎

PR #101, opened 2022-06-09
Adding Integrated Windows Authentication to auth mode and token result.

shalinikhare27:
This is the first PR for integrating IWA and contains the following:
1. Add IWA to auth modes if platform is Windows.
2. Added tests to validate the changes.
3. Added Interactive Windows Auth to TokenResult to capture Auth type.

### DIFF
--- a/src/MSALWrapper.Test/AuthModeTest.cs
+++ b/src/MSALWrapper.Test/AuthModeTest.cs
@@ -15,19 +15,31 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         [Test]
         public void AllIsAll()
         {
-            (AuthMode.Broker | AuthMode.Web | AuthMode.DeviceCode).Should().Be(AuthMode.All);
+            (AuthMode.IWA | AuthMode.Broker | AuthMode.Web | AuthMode.DeviceCode).Should().Be(AuthMode.All);
         }
 
         [Test]
         public void WindowsDefaultModes()
         {
             var subject = AuthMode.Default;
+            subject.IsIWA().Should().BeTrue();
             subject.IsBroker().Should().BeTrue();
             subject.IsWeb().Should().BeTrue();
             subject.IsDeviceCode().Should().BeFalse();
         }
 
         [TestCase(AuthMode.All, true)]
+        [TestCase(AuthMode.IWA, true)]
+        [TestCase(AuthMode.Broker, false)]
+        [TestCase(AuthMode.Web, false)]
+        [TestCase(AuthMode.DeviceCode, false)]
+        public void IWAIsExpected(AuthMode subject, bool expected)
+        {
+            subject.IsIWA().Should().Be(expected);
+        }
+
+        [TestCase(AuthMode.All, true)]
+        [TestCase(AuthMode.IWA, false)]
         [TestCase(AuthMode.Broker, true)]
         [TestCase(AuthMode.Web, false)]
         [TestCase(AuthMode.DeviceCode, false)]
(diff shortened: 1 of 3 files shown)


src/MSALWrapper.Test/AuthModeTest.cs
```diff
@@ -64,14 +78,19 @@ public void OtherCombos()
...
             subject.IsWeb().Should().BeTrue();
+
+            subject = AuthMode.IWA | AuthMode.Broker;
+            subject.IsIWA().Should().BeTrue();
+            subject.IsBroker().Should().BeTrue();
         }
 
         [Test]
-        public void WebOrDeviceCodeIsNotbroker()
+        public void WebOrDeviceCodeIsNotBrokerAndIWA()
```
reillysiemens, line 88: Should this be `WebOrDeviceCodeIsNotBrokerOrIWA`? If so, could that be a separate test instead of mixing these two?

reillysiemens (dismissed):

src/MSALWrapper/TokenResult.cs
```diff
@@ -25,6 +25,11 @@ public enum AuthType
         /// Device code flow auth type.
         /// </summary>
         DeviceCodeFlow,
+
+        /// <summary>
+        /// Interactive Windows auth flow auth type.
```
melindajohnson, line 30:
```suggestion
        /// Integrated Windows auth flow auth type.

```

src/MSALWrapper.Test/AuthModeTest.cs
shalinikhare27: It's the same thing. We are just validating that if we have `web` or `devicecode` as auth modes then broker and IWA will be false. Let me know your preference if we should use `or` or `and`.

reillysiemens: I think `or` makes more sense to me in this context if that's alright. 😃

src/MSALWrapper/TokenResult.cs
```diff
@@ -25,6 +25,11 @@ public enum AuthType
         /// Device code flow auth type.
         /// </summary>
         DeviceCodeFlow,
+
+        /// <summary>
+        /// Integrated Windows auth flow auth type.
+        /// </summary>
+        IntegratedWindowsAuthenticationFlow,
```
kyle-rader, line 32: @mvanchaa - Heads Up I expect your upcoming telemetry work to collide with this auth_type change (maybe not as an actual merge conflict but as something to reconcile). 

Right now we had Silent, Interactive, and DeviceCodeFlow - however, DeviceCodeFlow could technically be Silent.  SO I expect we want to drop Silent/Interactive here and move to having each authFlow have an interactive bool field. We will know which flow it was from the event itself, and then this auth_type field can be deprecated (Since the type will be dynamically computed from the AuthFlow class in use).

src/MSALWrapper/AuthMode.cs
```diff
@@ -27,15 +27,20 @@ public enum AuthMode : short
...
+        /// <summary>
+        /// Integrated Windows auth mode.
+        /// </summary>
+        IWA = 1 << 3,
+
         /// <summary>
         /// All auth modes.
         /// </summary>
-        All = Broker | Web | DeviceCode,
+        All = IWA | Broker | Web | DeviceCode,
```
kyle-rader, line 38: I'd be more comfortable is the non-existent mode was not added to this fields used by the CLI already, though it is a not a breaking change given it just won't be created.

kyle-rader: (This is not a request for changes, I'm okay to approve as is).

kyle-rader (approved):

reillysiemens (approved):

PR #102, opened 2022-06-10
Wait after killing a task before removing the directory

reillysiemens:
We've observed that there may be a concurrency issue where after running `taskkill` the OS still has locks open for the DLLs that it was using. This is definitely a band-aid fix, but for now if we sleep for about a second that appears to be enough time to let those locks close and allow the installer to proceed.


goagain (approved):

kyle-rader (approved):
I have seen the madness with my own eyes.

kyle-rader:
We have fun here.

PR #105, opened 2022-06-13
Integrated Windows Authentication implementation

shalinikhare27:
This is the main PR that integrates and implements IWA as one of the auth flows.

1. Created a new class for IWA and added tests. 
- We check for accounts in cache and try to get token silently.
- If no accounts in cache, then we try IWA as the first auth flow as it works with any Windows version.
- The error handling is mostly inspired by Edge's codebase since if they shell out to AzureAuth in future, the experience should be seamless. I will put link in chat for the code link.
3. Added IWA to AuthFactory class and added tests.
- We try IWA as the first Auth flow since it works with any Windows version.

### DIFF
--- a/src/MSALWrapper.Test/AuthFlow/AuthFlowFactoryTest.cs
+++ b/src/MSALWrapper.Test/AuthFlow/AuthFlowFactoryTest.cs
@@ -102,6 +102,15 @@ namespace MSALWrapper.Test
 
 #if PlatformWindows
         [Test]
+        public void IntegratedWindowsAuthentication_Only()
+        {
+            IEnumerable<IAuthFlow> subject = this.Subject(AuthMode.IWA);
+
+            subject.Should().HaveCount(1);
+            subject.First().GetType().Name.Should().Be(typeof(IntegratedWindowsAuthentication).Name);
+        }
+
+        [Test]
         public void Broker_Only()
         {
             this.MockIsWindows10Or11(true);
@@ -119,13 +128,14 @@ namespace MSALWrapper.Test
 
             IEnumerable<IAuthFlow> subject = this.Subject(AuthMode.Default);
 
-            subject.Should().HaveCount(2);
+            subject.Should().HaveCount(3);
 
             // BeEquivalentTo doesn't assert order for a list :(
             // so explicitly assert the first and second item names.
             var names = subject.Select(a => a.GetType().Name).ToList();
-            names[0].Should().Be(typeof(Broker).Name);
-            names[1].Should().Be(typeof(Web).Name);
+            names[0].Should().Be(typeof(IntegratedWindowsAuthentication).Name);
+            names[1].Should().Be(typeof(Broker).Name);
+            names[2].Should().Be(typeof(Web).Name);
         }
 
         [Test]
@@ -135,9 +145,45 @@ namespace MSALWrapper.Test
 
             IEnumerable<IAuthFlow> subject = this.Subject(AuthMode.Default);
 
-            subject.Should().HaveCount(1);
+            subject.Should().HaveCount(2);
             var names = subject.Select(a => a.GetType().Name).ToList();
-            names[0].Should().Be(typeof(Web).Name);
+            names[0].Should().Be(typeof(IntegratedWindowsAuthentication).Name);
+            names[1].Should().Be(typeof(Web).Name);
+        }
+
+        [Test]
+        public void Windows10Or11_All()
+        {
+            this.MockIsWindows10Or11(true);
+
+            IEnumerable<IAuthFlow> subject = this.Subject(AuthMode.All);
+
+            subject.Should().HaveCount(4);
+
+            // BeEquivalentTo doesn't assert order for a list :(
+            // so explicitly assert the first and second item names.
+            var names = subject.Select(a => a.GetType().Name).ToList();
+            names[0].Should().Be(typeof(IntegratedWindowsAuthentication).Name);
+            names[1].Should().Be(typeof(Broker).Name);
+            names[2].Should().Be(typeof(Web).Name);
+            names[3].Should().Be(typeof(DeviceCode).Name);
+        }
+
+        [Test]
+        public void Windows_All()
+        {
+            this.MockIsWindows10Or11(false);
+
+            IEnumerable<IAuthFlow> subject = this.Subject(AuthMode.All);
+
+            subject.Should().HaveCount(3);
+
+            // BeEquivalentTo doesn't assert order for a list :(
+            // so explicitly assert the first and second item names.
+            var names = subject.Select(a => a.GetType().Name).ToList();
+            names[0].Should().Be(typeof(IntegratedWindowsAuthentication).Name);
+            names[1].Should().Be(typeof(Web).Name);
+            names[2].Should().Be(typeof(DeviceCode).Name);
         }
 #endif
 
@@ -155,7 +201,7 @@ namespace MSALWrapper.Test
         [Platform("Win")]
         public void AllModes_Windows()
         {
-            this.MockIsWindows10Or11(true);
+            this.MockIsWindows10Or11(false);
 
             IEnumerable<IAuthFlow> subject = this.Subject(AuthMode.All);
 
@@ -166,6 +212,28 @@ namespace MSALWrapper.Test
                 .Should()
                 .BeEquivalentTo(new[]
                 {
+                    typeof(IntegratedWindowsAuthentication).Name,
+                    typeof(Web).Name,
+                    typeof(DeviceCode).Name,
+                });
+        }
+
+        [Test]
+        [Platform("Win")]
+        public void AllModes_Windows10Or11()
+        {
+            this.MockIsWindows10Or11(true);
+
+            IEnumerable<IAuthFlow> subject = this.Subject(AuthMode.All);
+
+            this.pcaWrapperMock.VerifyAll();
+            subject.Should().HaveCount(4);
+            subject
+                .Select(flow => flow.GetType().Name)
+                .Should()
+                .BeEquivalentTo(new[]
+                {
+                    typeof(IntegratedWindowsAuthentication).Name,
                     typeof(Broker).Name,
                     typeof(Web).Name,
                     typeof(DeviceCode).Name,
--- a/src/MSALWrapper.Test/AuthFlow/AuthFlowFactoryTest.cs
+++ b/src/MSALWrapper.Test/AuthFlow/AuthFlowFactoryTest.cs
@@ -104,6 +104,8 @@ namespace MSALWrapper.Test
         [Test]
         public void IntegratedWindowsAuthentication_Only()
         {
+            this.MockIsWindows(true);
+
             IEnumerable<IAuthFlow> subject = this.Subject(AuthMode.IWA);
 
             subject.Should().HaveCount(1);
@@ -124,6 +126,7 @@ namespace MSALWrapper.Test
         [Test]
         public void Windows10Or11_Defaults()
         {
+            this.MockIsWindows(true);
             this.MockIsWindows10Or11(true);
 
             IEnumerable<IAuthFlow> subject = this.Subject(AuthMode.Default);
@@ -141,6 +144,7 @@ namespace MSALWrapper.Test
         [Test]
         public void Windows_Defaults()
         {
+            this.MockIsWindows(true);
             this.MockIsWindows10Or11(false);
 
             IEnumerable<IAuthFlow> subject = this.Subject(AuthMode.Default);
@@ -154,6 +158,7 @@ namespace MSALWrapper.Test
         [Test]
         public void Windows10Or11_All()
         {
+            this.MockIsWindows(true);
             this.MockIsWindows10Or11(true);
 
             IEnumerable<IAuthFlow> subject = this.Subject(AuthMode.All);
@@ -172,6 +177,7 @@ namespace MSALWrapper.Test
         [Test]
         public void Windows_All()
         {
+            this.MockIsWindows(true);
             this.MockIsWindows10Or11(false);
 
             IEnumerable<IAuthFlow> subject = this.Subject(AuthMode.All);
@@ -201,6 +207,7 @@ namespace MSALWrapper.Test
         [Platform("Win")]
         public void AllModes_Windows()
         {
+            this.MockIsWindows(true);
             this.MockIsWindows10Or11(false);
 
             IEnumerable<IAuthFlow> subject = this.Subject(AuthMode.All);
@@ -222,6 +229,7 @@ namespace MSALWrapper.Test
         [Platform("Win")]
         public void AllModes_Windows10Or11()
         {
+            this.MockIsWindows(true);
             this.MockIsWindows10Or11(true);
 
             IEnumerable<IAuthFlow> subject = this.Subject(AuthMode.All);
@@ -276,6 +284,11 @@ namespace MSALWrapper.Test
         private void MockIsWindows10Or11(bool value)
         {
             this.platformUtilsMock.Setup(p => p.IsWindows10Or11()).Returns(value);
+        }
+
+        private void MockIsWindows(bool value)
+        {
+            this.platformUtilsMock.Setup(p => p.IsWindows()).Returns(value);
         }
     }
 }
--- a/src/MSALWrapper/AuthFlow/AuthFlowFactory.cs
+++ b/src/MSALWrapper/AuthFlow/AuthFlowFactory.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
             // and tries to auth silently if it finds an account in the cache.
             if (authMode.IsIWA())
             {
-                flows.Add(new IntegratedWindowsAuthentication(logger, clientId, tenantId, scopes, osxKeyChainSuffix, preferredDomain, pcaWrapper));
+                flows.Add(new IntegratedWindowsAuthentication(logger, clientId, tenantId, scopes, preferredDomain, pcaWrapper));
             }
 
             // This check silently fails on winserver if broker has been requested.
--- a/src/MSALWrapper/AuthFlow/AuthFlowFactory.cs
+++ b/src/MSALWrapper/AuthFlow/AuthFlowFactory.cs
@@ -46,8 +46,8 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
             List<IAuthFlow> flows = new List<IAuthFlow>();
 
             // We try IWA as the first auth flow as it works for any Windows version
-            // and tries to auth silently if it finds an account in the cache.
-            if (authMode.IsIWA())
+            // and tries to auth silently.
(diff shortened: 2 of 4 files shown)


src/MSALWrapper.Test/AuthFlow/AuthFlowFactoryTest.cs
```diff
@@ -135,9 +145,45 @@ public void Windows_Defaults()
...
+            this.MockIsWindows10Or11(false);
+
+            IEnumerable<IAuthFlow> subject = this.Subject(AuthMode.All);
+
+            subject.Should().HaveCount(3);
+
+            // BeEquivalentTo doesn't assert order for a list :(
+            // so explicitly assert the first and second item names.
             var names = subject.Select(a => a.GetType().Name).ToList();
-            names[0].Should().Be(typeof(Web).Name);
```
melindajohnson, line 140: Shouldn't `windows_all` also have broker mode?

shalinikhare27: Broker only works for Windows 10 and 11. If you see we have a test that checks flows that will queue up if windows version is 10 or 11.

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -0,0 +1,170 @@
...
+                {
+                    this.errors.Add(ex);
+                    this.logger.LogDebug($"IWA failed, 2FA is required.\n{ex.Message}");
+                }
+                catch (MsalUiRequiredException ex) when (
+                             ex.Classification == UiRequiredExceptionClassification.None
+                          && ex.Message.StartsWith("AADSTS500083", StringComparison.OrdinalIgnoreCase))
+                {
+                    this.errors.Add(ex);
+                    this.logger.LogDebug($"IWA failed, unable to verify token signature with identifier 'urn:federation:MSFT'.\n{ex.Message}");
```
Penguinwizzard, line 133: If this error message is user-exposed, what you do expect a user to do with it? There's not really much to go on if you don't already know the internals (in which case you're probably someone who's working on the tools, or who would rather see the full exception).

src/MSALWrapper/AuthFlow/AuthFlowFactory.cs
```diff
@@ -45,6 +45,13 @@ public class AuthFlowFactory
             // as it sets the order in which auth flows will be attempted.
             List<IAuthFlow> flows = new List<IAuthFlow>();
 
+            // We try IWA as the first auth flow as it works for any Windows version
+            // and tries to auth silently if it finds an account in the cache.
```
Penguinwizzard, line 49: nit: IWA is a silent flow, and doesn't work off of the MSAL cache, but rather your Windows login session state.

mvanchaa (approved):

src/MSALWrapper/AuthFlow/AuthFlowFactory.cs
shalinikhare27: I was referring to trying `AcquireTokenSilent` if we find account in cache. If account is null, then we try IWA. Let me know if I should state this a bit differently. :)

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
shalinikhare27: This is not user-exposed but collected for our telemetry. We queue a bunch of different auth flows depending on the user platform. When we get a null token, we capture the exception for our telemetry and then move on to the next auth flow in the queue.

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -0,0 +1,170 @@
...
+                {
+                    this.logger.LogWarning($"Msal Client Exception! Could not identify logged in user.\n{ex.Message}");
+                    this.errors.Add(ex);
+                }
+            }
+
+            return new AuthFlowResult(null, this.errors);
+        }
+
+        private IPCAWrapper BuildPCAWrapper(ILogger logger, Guid clientId, Guid tenantId, string osxKeyChainSuffix)
```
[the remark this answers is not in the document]
shalinikhare27: Great suggestions, Derek. Our team contemplated a lot between using abstract classes versus interfaces during the design review when we refactored our entire codebase. We went with interfaces then.
@kyle-rader - do you have any strong feels about this? If not, I will address it in one of the future PRs.

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -0,0 +1,170 @@
...
+                }
+            }
+            else
+            {
+                try
+                {
+                    var tokenResult = await TaskExecutor.CompleteWithin(
+                                    this.logger,
+                                    this.interactiveWindowsAuthTimeout,
+                                    "Get Token Interactive Windows Authentication",
```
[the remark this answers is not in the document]
shalinikhare27: Omg, yes I know. I kept hopping between `integrated` and `interactive` knowing that it's `integrated`. Just a few one offs where I have messed u. Fixed!

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -0,0 +1,170 @@
...
+                    this.errors.Add(ex);
+                }
+            }
+            else
+            {
+                try
+                {
+                    var tokenResult = await TaskExecutor.CompleteWithin(
+                                    this.logger,
+                                    this.interactiveWindowsAuthTimeout,
```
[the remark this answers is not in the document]
shalinikhare27: I understand. It was a sheer mistake and a few one offs here and there. The spelling is quite long and similar. :) I fixed it though.

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -0,0 +1,170 @@
...
+        private readonly string preferredDomain;
+        private readonly IList<Exception> errors;
+        private IPCAWrapper pcaWrapper;
+
+        #region Public configurable properties
+
+        /// <summary>
+        /// The silent auth timeout.
+        /// </summary>
+        private TimeSpan silentAuthTimeout = TimeSpan.FromMinutes(5);
```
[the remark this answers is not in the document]
shalinikhare27: Very good point. I tried to stay consistent with the timeout we set for silent auth in Broker.
@kyle-rader - Ideally, silent auth should be in about 300 ms as I heard in our demo today. So should we change the timeout limit to 5 seconds(more than ample time) in future PRs?

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -0,0 +1,170 @@
...
+                    tokenResult.SetAuthenticationType(AuthType.IntegratedWindowsAuthenticationFlow);
+
+                    return new AuthFlowResult(tokenResult, this.errors);
+                }
+                catch (MsalUiRequiredException ex) when (
+                             ex.Classification == UiRequiredExceptionClassification.BasicAction
+                          && ex.Message.StartsWith("AADSTS50076", StringComparison.OrdinalIgnoreCase))
+                {
+                    this.errors.Add(ex);
+                    this.logger.LogDebug($"IWA failed, 2FA is required.\n{ex.Message}");
```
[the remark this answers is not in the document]
shalinikhare27: Since we gather this only for our telemetry, do you have any specific suggestions?

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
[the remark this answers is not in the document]
kyle-rader: If an auth flow is doing a "silent auth token refresh" the web requests _can_ take as much as 5 seconds. IWA doesn't do this, though still depends on network traffic. But as @Penguinwizzard mentions - 5 minutes is, too generous. 6 seconds might be a more reasonable timeout for silent auth. @shalinikhare27 I think you can go ahead and change this to 5 seconds here, because IWA can only ever be silent, or fail, and if it takes more than 5 seconds, we probably want to move on.

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -0,0 +1,170 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.MSALWrapper.AuthFlow
+{
+#if NET472
+    using Microsoft.Identity.Client.Desktop;
+#endif
+    using System;
```
kyle-rader, line 9: We do not build for net472 ever now, so this should not be needed.

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        /// <summary>
+        /// The silent auth timeout.
+        /// </summary>
+        private TimeSpan silentAuthTimeout = TimeSpan.FromMinutes(5);
+
+        /// <summary>
+        /// The intercative windows auth flow timeout.
+        /// </summary>
+        private TimeSpan integratedWindowsAuthTimeout = TimeSpan.FromMinutes(15);
```
kyle-rader, line 36: Since IWA has no interactive path - we should only have 1 Timeout defined. Probably this one, so probably make this one 6 seconds and use it for both?

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -0,0 +1,170 @@
...
+        /// Initializes a new instance of the <see cref="IntegratedWindowsAuthentication"/> class.
+        /// </summary>
+        /// <param name="logger">The logger.</param>
+        /// <param name="clientId">The client id.</param>
+        /// <param name="tenantId">The tenant id.</param>
+        /// <param name="scopes">The scopes.</param>
+        /// <param name="osxKeyChainSuffix">The osx key chain suffix.</param>
+        /// <param name="preferredDomain">The preferred domain.</param>
+        /// <param name="pcaWrapper">Optional: IPCAWrapper to use.</param>
+        public IntegratedWindowsAuthentication(ILogger logger, Guid clientId, Guid tenantId, IEnumerable<string> scopes, string osxKeyChainSuffix = null, string preferredDomain = null, IPCAWrapper pcaWrapper = null)
```
kyle-rader, line 49: IWA doesn't exist on Windows so it doesn't make sense to have an `osxKeyChainSuffix` argument here. Could remove, and it where required for the cache pass null?

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -0,0 +1,170 @@
...
+                this.logger.LogDebug($"Using cached account '{account.Username}'");
+                try
+                {
+                    try
+                    {
+                        var tokenResult = await TaskExecutor.CompleteWithin(
+                            this.logger,
+                            this.silentAuthTimeout,
+                            "Get Token Silent",
+                            (cancellationToken) => this.pcaWrapper.GetTokenSilentAsync(this.scopes, account, cancellationToken),
```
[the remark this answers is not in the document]
kyle-rader: I agree - now that we have refactored to the notion of AuthFlows, we can clearly see a better abstraction. I'm good to let this duplication happen now, and pull it out in a future pull request. @shalinikhare27 can you log a bug for that?

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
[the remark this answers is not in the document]
kyle-rader: Same comment as before, now that we are here, the duplication is more visible, and yes we can address in a future PR. We can also log a bug for this change.

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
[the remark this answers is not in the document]
kyle-rader: Perhaps we can use `LogWarning` here to note that IWA failed with 2FA required and tell them, IWA can pass this requirement if you log into Windows with either a Smart Card or Windows Hello. This way it's shown to the user, as we go onto to do some interactive auth. This way they know more about why they are getting prompted and what to do.

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
kyle-rader: Is catching this error Microsoft specific? I think the first `catch` for ui required exception makes sense, because we can tell the user something that might help. I think this second catch should be an unconditional catch of the MsalUiRequiredException to catch all other forms, and simpel log the message directly, and add it to the error list. We can then track it in Telemetry and see how frequently this comes up. But I agree, it's not actionable to the user, so let's leave it as LogDebug.

src/MSALWrapper/AuthFlow/AuthFlowFactory.cs
kyle-rader: I think fine for now, we ack the opporuntity to refactor the MSAL cache silent attempt into it's own AuthFlow.

src/MSALWrapper/AuthFlow/AuthFlowFactory.cs
```diff
@@ -45,6 +45,13 @@ public class AuthFlowFactory
             // as it sets the order in which auth flows will be attempted.
             List<IAuthFlow> flows = new List<IAuthFlow>();
 
+            // We try IWA as the first auth flow as it works for any Windows version
+            // and tries to auth silently if it finds an account in the cache.
+            if (authMode.IsIWA())
```
kyle-rader, line 50: IWA only makes sense on Windows on below we are also checking if Windows for WAM. Do we need to check for windows here> or is the `IsIWA()` platform aware?

kyle-rader (changes requested):
Looking good, I only have a few small suggestions.

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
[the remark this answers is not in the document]
shalinikhare27: Yes, the bug is logged and added to shield backlog.

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -0,0 +1,170 @@
...
+                .WithLogging(
+                    this.LogMSAL,
+                    Identity.Client.LogLevel.Verbose,
+                    enablePiiLogging: false,
+                    enableDefaultPlatformLogging: true);
+
+            return new PCAWrapper(this.logger, clientBuilder.Build(), this.errors, tenantId, osxKeyChainSuffix);
+        }
+
+        private void LogMSAL(Identity.Client.LogLevel level, string message, bool containsPii)
```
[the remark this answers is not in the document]
shalinikhare27: @[Penguinwizzard](https://github.com/Penguinwizzard) We have created a work item to track this and will address it in future. For this PR, let's focus on the implementation. Sorry, I should mentioned before Derek. :) Really appreciate this feedback though. :)

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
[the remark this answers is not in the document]
shalinikhare27: Logged a bug and added to shield.

src/MSALWrapper/AuthFlow/AuthFlowFactory.cs
shalinikhare27: Perfect, bug has been logged to track this.

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
shalinikhare27: Good catch, removed!

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
shalinikhare27: I made the silent timeout (that uses cached accounts) and IWA timeout to be 6 seconds. I am leaving them separate as they still are trying different APIs behind the scenes and this makes it more explicit.

shalinikhare27: integratedWindowsAuthTimeout is now 6 seconds and used in both places.

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
shalinikhare27: Removed osxKeyChainSuffix and passing null to pcaWrapper.

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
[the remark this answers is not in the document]
shalinikhare27: Update the timeout to be 6 seconds.

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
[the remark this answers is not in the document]
shalinikhare27: Changing LogDebug to LogWarning so that users are get helpful information before we try to run an interactive prompt. This way they know how to avoid interactive prompts next time.

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
shalinikhare27: Updated the second block to catch any type of MSAL UI Exception as it's non-actionable for the users.

src/MSALWrapper/AuthFlow/AuthFlowFactory.cs
shalinikhare27: Added platformUtils.IsWindows() flag check for IWA. IsIWA() did check for windows platform but we should check it before queuing as well just like we do for Broker. Good catch!

Penguinwizzard (commented):
LGTM

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
Penguinwizzard: This error (AADSTS500083) isn't technically Microsoft-specific, but it's specific to a very particular situation (the trust relationship has been broken between the AD and AAD tenants involved) which is pretty uncommon. The changes made seem reasonable.

kyle-rader (approved):
Thanks for all the updates!

mvanchaa (approved):

goagain (approved):

PR #106, opened 2022-06-14
Send AuthFlow telemetry events to the App Insights

mvanchaa:
This is the first PR for the feature - Improving the AzureAuth telemetry instrumentation.

The final agenda is to make the following schema changes:

Drawing

Changes in this PR:
1. Added Lasso to MSALWrapper as we need `ITelemetryService` and `EventData` classes to send custom events of Auth Flows.
2. Changed `CommandMain` and `AuthFlowExecutor` classes to make use of `ITelemetryService`.
3. Populate and send `EventData` from `AuthFlowExecutor` class.

TODO:
1. Unit Tests (adding new and updating some old ones)
2. Add new attributes mentioned in the above schema to the main event `azureauth_command_azureauth`. I will create a new PR for this.
3. Fix Authentication type logic - (Currently, we capture authentication types as silent/interactive/devicecodeflow. These are not enough/accurate to capture whether a given authentication session was silent or not as the type - devicecodeflow can indicate a silent or interactive auth prompt. So there is a need for a flag to be added to the schema which indicates whether the authentication was silent or not.). I will create a new PR for this.

### DIFF
--- a/src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
+++ b/src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
@@ -126,6 +126,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
             result.TokenResult.Should().Be(this.tokenResult);
             result.Success.Should().BeTrue();
             result.Errors.Should().BeEmpty();
+            result.InteractivePromptsCount.Should().Be(0);
         }
 
         [Test]
@@ -144,6 +145,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
             result.Should().NotBeNull();
             result.Success.Should().BeFalse();
             result.Errors.Should().BeEmpty();
+            result.InteractivePromptsCount.Should().Be(0);
         }
 
         [Test]
--- a/src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
+++ b/src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
@@ -80,7 +80,8 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         public void ConstructorWith_Null_AuthFlows()
         {
             var logger = this.serviceProvider.GetService<ILogger<AuthFlowExecutor>>();
-            Action authFlowExecutor = () => new AuthFlowExecutor(logger, null, null);
+            var telemetryService = this.serviceProvider.GetService<ITelemetryService>();
+            Action authFlowExecutor = () => new AuthFlowExecutor(logger, telemetryService, null);
 
             // Assert
             authFlowExecutor.Should().Throw<ArgumentNullException>();
@@ -628,7 +629,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         private AuthFlowExecutor Subject(IEnumerable<IAuthFlow> authFlows)
         {
             var logger = this.serviceProvider.GetService<ILogger<AuthFlowExecutor>>();
-            return new AuthFlowExecutor(logger, null, authFlows);
+            return new AuthFlowExecutor(logger, this.telemetryServiceMock, authFlows);
         }
     }
 }
--- a/src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
+++ b/src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
@@ -15,6 +15,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
     using Microsoft.Extensions.Logging;
     using Microsoft.Identity.Client;
     using Microsoft.IdentityModel.JsonWebTokens;
+    using Microsoft.Office.Lasso.Interfaces;
     using Moq;
     using NLog.Extensions.Logging;
     using NLog.Targets;
@@ -28,6 +29,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         private MemoryTarget logTarget;
         private TokenResult tokenResult;
         private IEnumerable<IAuthFlow> authFlows;
+        private Mock<ITelemetryService> telemetryServiceMock;
 
         [SetUp]
         public void Setup()
@@ -38,6 +40,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
             loggingConfig.AddTarget(this.logTarget);
             loggingConfig.AddRuleForAllLevels(this.logTarget);
             this.authFlows = new List<IAuthFlow>();
+            this.telemetryServiceMock = new Mock<ITelemetryService>();
 
             // Setup Dependency Injection container to provide logger and out class under test (the "subject")
             this.serviceProvider = new ServiceCollection()
@@ -48,6 +51,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
                  loggingBuilder.SetMinimumLevel(Microsoft.Extensions.Logging.LogLevel.Trace);
                  loggingBuilder.AddNLog(loggingConfig);
              })
+             .AddSingleton(this.telemetryServiceMock.Object)
              .BuildServiceProvider();
 
             // Mock successful token result
@@ -83,10 +87,21 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         }
 
         [Test]
-        public void ConstructorWith_Valid_Arguments()
+        public void ConstructorWith_Null_TelemetryService()
         {
             var logger = this.serviceProvider.GetService<ILogger<AuthFlowExecutor>>();
             Action authFlowExecutor = () => new AuthFlowExecutor(logger, null, this.authFlows);
+
+            // Assert
+            authFlowExecutor.Should().Throw<ArgumentNullException>();
+        }
+
+        [Test]
+        public void ConstructorWith_Valid_Arguments()
+        {
+            var logger = this.serviceProvider.GetService<ILogger<AuthFlowExecutor>>();
+            var telemetryService = this.serviceProvider.GetService<ITelemetryService>();
+            Action authFlowExecutor = () => new AuthFlowExecutor(logger, telemetryService, this.authFlows);
 
             // Assert
             authFlowExecutor.Should().NotThrow<ArgumentNullException>();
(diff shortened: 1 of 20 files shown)


src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -17,15 +19,18 @@ public class AuthFlowExecutor : IAuthFlow
...
         /// Initializes a new instance of the <see cref="AuthFlowExecutor"/> class.
         /// </summary>
         /// <param name="logger">The logger.</param>
+        /// <param name="telemetryService">The telemetry service.</param>
         /// <param name="authFlows">The list of auth flows.</param>
-        public AuthFlowExecutor(ILogger logger, IEnumerable<IAuthFlow> authFlows)
+        public AuthFlowExecutor(ILogger logger, ITelemetryService telemetryService, IEnumerable<IAuthFlow> authFlows)
         {
             this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            this.telemetryService = telemetryService;
```
kyle-rader, line 33: Consider following the same pattern here that we expect non-null args and throw an exception if given a null arg.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -69,5 +76,11 @@ public async Task<AuthFlowResult> GetTokenAsync()
 
             return result;
         }
+
+        private void SendTelemetryEvent(AuthFlowResult attempt, string authFlowName)
+        {
+            // will be implemented in the next PR
+            return;
+        }
```
kyle-rader, line 84: I'm not sure we need the helper method here yet. I'd rather see us collect the metrics about this auth flow and send it first before pre-abstracting. 

I think in this PR, it's fine to go ahead and actually send the event and will make the whole change easier to reason about seeing the full change in the same diff. 

To send an event, you need an event name, and an instance of `EventData`. Consider creating your EventData instance at the start of the for loop - one for each authFlow attempt. You can then add values to that eventData, and at the end of the loop, send the event constructing an event name based off the authFlowName.

kyle-rader (changes requested):
Great, start - I don't think you need to break this change up and it will be easier to review the collection of the telemetry if we do adding the sending. Effectively this PR is 90% of the way to sending a new auth flow event, and doing the last 10% in a different PR makes the whole change harder to review.

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
```diff
@@ -86,7 +86,7 @@ public void ConstructorWith_Null_AuthFlows()
         public void ConstructorWith_Valid_Arguments()
         {
             var logger = this.serviceProvider.GetService<ILogger<AuthFlowExecutor>>();
-            Action authFlowExecutor = () => new AuthFlowExecutor(logger, this.authFlows);
+            Action authFlowExecutor = () => new AuthFlowExecutor(logger, null, this.authFlows);
```
shalinikhare27, line 89: Shouldn't this be non-null for valid args?

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
```diff
@@ -55,9 +55,9 @@ public void Setup()
         }
 
         [Test]
-        public void ConstructorWith_BothNullArgs()
+        public void ConstructorWith_AllNullArgs()
```
shalinikhare27, line 58: We probably need a test for valid telemetry service, but logger ad auth flows as null?

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
shalinikhare27: +1 to Kyle.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -48,9 +53,11 @@ public async Task<AuthFlowResult> GetTokenAsync()
                 this.logger.LogDebug($"Starting {authFlowName}...");
 
                 var attempt = await authFlow.GetTokenAsync();
+                this.SendTelemetryEvent(attempt, authFlowName);
+
                 if (attempt == null)
                 {
-                    var oopsMessage = $"Auth flow '{authFlow.GetType().Name}' returned a null AuthFlowResult.";
+                    var oopsMessage = $"Auth flow '{authFlowName}' returned a null AuthFlowResult.";
```
shalinikhare27, line 60: How is the `authFlowName` set?

shalinikhare27 (changes requested):
Waiting for some comment replies.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
mvanchaa: I thought of creating an instance of `EventData` inside each AuthFlow and then populating it inside the AuthFlow. When returning from any AuthFlow, this `EventData` instance will be returned along with the `AuthFlowResult`  like
```
return new AuthFlowResult(tokenResult, eventdata);
```

I will add all the changes inside one PR, I think that will make it easy to review.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
mvanchaa: This change is not highlighted because you added the same change I think and I merged your changes into this PR.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
mvanchaa: Got it!

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
mvanchaa: Yup, that's true. I will change it.

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
mvanchaa: Yes that's true.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
mvanchaa: @kyle-rader add changes here. Let me know if this approach doesn't seem good.

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -98,15 +108,18 @@ public async Task<AuthFlowResult> GetTokenAsync()
                             .GetTokenInteractiveAsync(this.scopes, account, cancellationToken),
                             this.errors)
                             .ConfigureAwait(false);
-                        tokenResult.SetAuthenticationType(AuthType.Interactive);
+                        tokenResult.SetAuthenticationType(AuthType.Interactive); // confirm this.
```
mvanchaa, line 111: @kyle-rader I was testing the code locally and I did not see an interactive prompt at this step. It seemed to have silently authenticate. Am I missing something here? 

I was testing on a machine with windows hello face enabled. I'm assuming it's because of that I couldn't see anything here. Else it would have prompted to take windows hello PIN. Am I right? 

My question here is, the authentication type could be `Interactive` or `Silent` depending on the machine configuration.  So which authentication type can be set here?

Also, I have added a new attribute - `no_of_interactive_prompts` to the schema. Whenever we invoke an interactive prompt, I'm incrementing the `interactivePromptsCount`. Am I right here?

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
kyle-rader: Generally I want to avoid having the auth logic itself know anything about Telemetry. I am _okay_ having the AuthFlowExector know about Telemetry given it's a meta class design to run multiple auth flows. But each auth flow I think should only return it's information, and the consumer can then decide what to send telemetry wise. The limits the responisbiltiy of every auth flow class, and forces knowledge about telemetry into 1 place. It also helps keep the telemetry schema per auth flow similar / the same.

kyle-rader (changes requested):
See my main comment on not having any specific auth flow (except the auth flow executor, know about telemetry. Having each auth flow have to handle telemetry on it's own, as now seen, requires changing most files, and over-burdens the responsibility of each auth flow, + adds a lot of duplication that can be handled centrally by the auth flow executor. 

I was little on the fence about even adding the telemetry service to auth flow executor, but given it's current role, I think that's okay and makes sense, since it is the main orchestrator of auth flows.

src/MSALWrapper/ExceptionExtensions.cs
```diff
@@ -66,4 +67,29 @@ public static IEnumerable<Exception> GetAllExceptions(this Exception exception)
             }
         }
     }
+
+    /// <summary>
+    /// Extracts correlation IDs from the list of exceptions.
+    /// </summary>
+    /// <param name="exceptions">List of exceptions from which correlation IDs are extracted.</param>
+    /// <returns>List of correlation IDs.</returns>
+    public static List<string> ExtractCorrelationIDsFromException(IList<Exception> exceptions)
```
kyle-rader, line 76: Discussed offline - move this method into ExceptionListToStringConverter since it's not a true _extension_ method to the Exception class.

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -23,6 +24,7 @@ public class Broker : IAuthFlow
         private readonly string promptHint;
         private readonly IList<Exception> errors;
         private IPCAWrapper pcaWrapper;
+        private int interactivePromptsCount;
```
kyle-rader, line 27: Discussed offline - because we do not have a specific question to answer by tracking prompt count, we don't need to collect this information right now. Generally speaking a run of an auth flow should have 0 (silent) or 1 (interactive) prompts and 1 is already more than desired. So we're going to remove this count.

kyle-rader (changes requested):
Summary (from offline discussion):
Break apart the overal change into 3:
1. Add an immutable `AuthFlowName` property to `AuthFlowResult`
2. Change the interface of AuthFlowExecutor to return both the first successful `AuthFlowResult` as well as the full list of AuthFlowResults.
3. In CommandMain - utilize the now available data to send a separate event for each `AuthFlowResult` returned.

PR #107, opened 2022-06-16
Add AuthFlowName to the AuthFlowResult

mvanchaa:

Based on the comments on PR #106 , this new PR is created.

Changes:
1. Add `AuthFlowName` to the `AuthFlowResult` Class.

In the next PRs following changes will be made:
1. Currently, we send one final result from `AuthFlowExecutor` after each `AuthFlow` is attempted until succeeded. This will be changed to return all `AuthFlowResult`s for each `AuthFlow` attempted instead of just one. This is done so that these results can be leveraged in the `AzureAuth` CLI and send telemetry from there. With this approach, `MSALWrapper` will not have any dependency on `Lasso`.
2. Make changes to send telemetry events from the `AuthFlowResult` s returned to the CLI.

### DIFF
--- a/src/MSALWrapper/AuthFlow/AuthFlowResult.cs
+++ b/src/MSALWrapper/AuthFlow/AuthFlowResult.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Authentication.MSALWrapper
         /// Initializes a new instance of the <see cref="AuthFlowResult"/> class with a null TokenResult and empty error list.
         /// </summary>
         public AuthFlowResult()
-            : this(null, null, null)
+            : this(null, null, string.Empty)
         {
         }
 
@@ -29,7 +29,7 @@ namespace Microsoft.Authentication.MSALWrapper
         {
             this.TokenResult = tokenResult;
             this.Errors = errors ?? new List<Exception>();
-            this.AuthFlowName = authFlowName;
+            this.AuthFlowName = authFlowName ?? throw new ArgumentNullException(nameof(authFlowName));
         }
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


src/MSALWrapper/AuthFlow/AuthFlowResult.cs
```diff
@@ -24,10 +24,12 @@ public AuthFlowResult()
         /// </summary>
         /// <param name="tokenResult">A <see cref="MSALWrapper.TokenResult"/>.</param>
         /// <param name="errors">A list of errors encountered while getting (or failing to get) the given token result. Will initialize a new empty List if null is given.</param>
-        public AuthFlowResult(TokenResult tokenResult, IList<Exception> errors)
+        /// <param name="authFlowName">The name of the authflow from which the AuthFlowResult is returned or cretaed.</param>
+        public AuthFlowResult(TokenResult tokenResult, IList<Exception> errors, string authFlowName)
         {
             this.TokenResult = tokenResult;
             this.Errors = errors ?? new List<Exception>();
+            this.AuthFlowName = authFlowName;
```
kyle-rader, line 32: To help prevent null ref exceptions, because C#, I think we should ensure this is not null. (I acknowledge that in the future we should enable this kind of checking in the compiler since a newer version of C# supports it). But for now,

```
this.AuthFlowName = authFloweName ?? throw new ArgumentNullException(nameof(authFlowName));

kyle-rader (changes requested):
1 small suggestion on nullability.

src/MSALWrapper/AuthFlow/AuthFlowResult.cs
mvanchaa: @kyle-rader understood. Updated the changes. :)

goagain (approved):

kyle-rader (approved):

goagain (approved):

goagain (approved):
I feel like there are too many unit test changes for `AuthFlowName`. In my option, writing several separated tests for validating the AuthFlowName would be better. Whatever, I can accept these changes.

PR #109, opened 2022-06-16
Send AuthFlowResults to the CLI

mvanchaa:

Based on the comments on PR #106  , this new PR is created.

Changes:
Currently, we send one final result from `AuthFlowExecutor` after each `AuthFlow` is attempted until succeeded. This is changed to return all `AuthFlowResult`s for each `AuthFlow` attempted instead of just one. This is done so that these results can be leveraged in the `AzureAuth` CLI and send telemetry from there. With this approach, `MSALWrapper` will not have any dependency on `Lasso`.

### DIFF
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -360,7 +360,7 @@ Allowed values: [all, web, devicecode]";
             try
             {
                 AuthFlowExecutor authFlowExecutor = this.AuthFlowExecutor();
-                AuthFlowResult succeededResult = null;
+                AuthFlowResult successfulResult = null;
                 IEnumerable<AuthFlowResult> results = null;
 
                 // When running multiple AzureAuth processes with the same resource, client, and tenant IDs,
@@ -397,7 +397,7 @@ Allowed values: [all, web, devicecode]";
                     try
                     {
                         results = authFlowExecutor.GetTokenAsync().Result;
-                        succeededResult = results.FirstOrDefault(result => result.Success == true);
+                        successfulResult = results.FirstOrDefault(result => result.Success == true);
                     }
                     finally
                     {
@@ -408,24 +408,25 @@ Allowed values: [all, web, devicecode]";
                 var errors = results.SelectMany(result => result.Errors).ToList();
                 this.eventData.Add("error_list", ExceptionListToStringConverter.Execute(errors));
 
-                if (succeededResult == null)
+                if (successfulResult == null)
                 {
                     this.logger.LogError("Authentication failed. Re-run with '--verbosity debug' to get see more info.");
                     return 1;
                 }
 
-                this.eventData.Add("auth_type", $"{succeededResult.TokenResult.AuthType}");
+                var tokenResult = successfulResult.TokenResult;
+                this.eventData.Add("auth_type", $"{tokenResult.AuthType}");
 
                 switch (this.Output)
                 {
                     case OutputMode.Status:
-                        this.logger.LogSuccess(succeededResult.TokenResult.ToString());
+                        this.logger.LogSuccess(tokenResult.ToString());
                         break;
                     case OutputMode.Token:
-                        Console.Write(succeededResult.TokenResult.Token);
+                        Console.Write(tokenResult.Token);
                         break;
                     case OutputMode.Json:
-                        Console.Write(succeededResult.TokenResult.ToJson());
+                        Console.Write(tokenResult.ToJson());
                         break;
                     case OutputMode.None:
                         break;
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -361,7 +361,7 @@ Allowed values: [all, web, devicecode]";
             {
                 AuthFlowExecutor authFlowExecutor = this.AuthFlowExecutor();
                 AuthFlowResult succeededResult = null;
-                List<AuthFlowResult> results = null;
+                IEnumerable<AuthFlowResult> results = null;
 
                 // When running multiple AzureAuth processes with the same resource, client, and tenant IDs,
                 // They may prompt many times, which is annoying and unexpected.
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -57,6 +57,7 @@ Allowed values: [all, web, devicecode]";
         private readonly IEnv env;
         private Alias authSettings;
         private IAuthFlow authFlow;
+        private AuthFlowExecutor authFlowExecutor;
 
         /// <summary>
         /// The maximum time we will wait to acquire a mutex around prompting the user.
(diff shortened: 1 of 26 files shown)


src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -33,9 +33,9 @@ public AuthFlowExecutor(ILogger logger, IEnumerable<IAuthFlow> authFlows)
         /// Get a auth flow result.
         /// </summary>
         /// <returns>The <see cref="Task"/>.</returns>
-        public async Task<AuthFlowResult> GetTokenAsync()
+        public async Task<List<AuthFlowResult>> GetTokenAsync()
```
kyle-rader, line 36: I think we want to make it obvious which AuthFlow Result is the successful one, while still returning the list. Only returning a List means our code up top has to now find the successful , or no successful attempt.

mvanchaa: Discussed offline to go with this approach.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -48,26 +48,28 @@ public async Task<AuthFlowResult> GetTokenAsync()
...
-                    result.AddErrors(attempt.Errors);
-
                     this.logger.LogDebug($"{authFlowName} success: {attempt.Success}.");
+                    resultList.Add(attempt);
                     if (attempt.Success)
                     {
-                        result.TokenResult = attempt.TokenResult;
                         break;
                     }
                 }
```
kyle-rader, line 69:
```
                if (attempt == null)
                {
                    var oopsMessage = $"Auth flow '{authFlow.GetType().Name}' returned a null AuthFlowResult.";
                    this.logger.LogDebug(oopsMessage);

                    attempt = new AuthFlowResult(null, null, authFlowName);
                    attempt.Errors.Add(new NullTokenResultException(oopsMessage));
                }

                resultList.Add(attempt);

                if (attempt.Success)
                {
                    this.logger.LogDebug($"{authFlowName} success: {attempt.Success}.");
                    break;
                }
```

kyle-rader (changes requested):
Requests:
* Simplify main loop logic to be more declarative (suggestion in comment)
* * Condense Tests

src/AzureAuth/CommandMain.cs
```diff
@@ -394,35 +396,36 @@ private int GetToken()
 
                     try
                     {
-                        result = authFlow.GetTokenAsync().Result;
+                        results = authFlowExecutor.GetTokenAsync().Result;
+                        succeededResult = results.FirstOrDefault(result => result.Success == true);
```
reillysiemens, line 400: I can't quite put my finger on why, but `succeededResult` sounds a little awkward to me while `successfulResult` sounds fine. 😆

src/AzureAuth/CommandMain.cs
```diff
@@ -394,35 +396,36 @@ private int GetToken()
...
 
-                if (!result.Success)
+                if (succeededResult == null)
                 {
                     this.logger.LogError("Authentication failed. Re-run with '--verbosity debug' to get see more info.");
                     return 1;
                 }
 
-                this.eventData.Add("auth_type", $"{result.TokenResult.AuthType}");
+                this.eventData.Add("auth_type", $"{succeededResult.TokenResult.AuthType}");
```
reillysiemens, line 417: By this point you know you've got a non-null `succeededResult` and that it has a `TokenResult` on it. You could make the lines below a bit shorter if you pulled it out. For example
```suggestion
                var tokenResult = succeededResult.TokenResult;
                this.eventData.Add("auth_type", $"{tokenResult.AuthType}");

```
Doesn't save you much for this one line, but it does make this one _and_ the following 3 a little cleaner.

src/AzureAuth/CommandMain.cs
mvanchaa: I will update it.

src/AzureAuth/CommandMain.cs
shalinikhare27, line 417: Is it guarenteed that a non-null TokenResult will always have a non-null AuthType? Or should we do a null check here?

mvanchaa: from the unit tests it looks like it will always have a default `silent` AuthType. Figuring it out how.

mvanchaa: Just learnt that the default value of an enum property is always 0. So AuthType for TokenResult is Silent by default.

shalinikhare27 (dismissed):

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
```diff
@@ -96,37 +97,50 @@ public void ConstructorWith_Valid_Arguments()
...
+            var authFlowResultList = new List<AuthFlowResult>();
+            authFlowResultList.Add(authFlowResult);
             authFlow1.Setup(p => p.GetTokenAsync()).ReturnsAsync(authFlowResult);
 
             // Act
-            var authFlow = this.Subject(new[] { authFlow1.Object });
-            var result = await authFlow.GetTokenAsync();
+            var authFlowExecutor = this.Subject(new[] { authFlow1.Object });
+            var result = await authFlowExecutor.GetTokenAsync();
+            var resultList = result.ToList();
```
reillysiemens, line 109: Just a thought, but maybe... `results` instead of `resultList`? The compiler knows what the type is.

src/AzureAuth/CommandMain.cs
kyle-rader: There is an interesting divergence here - previously we were not using `null` to indicate failed auth. We designed AuthFlowResult to represent a failed auth attempt, while itself not being null.

This is why it has the Success property. By now that we have a list of them, and are searching for one with success, if we don't find one, we are left with null. I almost want to say the final result should just be the Last authFlowResult in the list. 
I think this is fine. Given you can run the CLI in such a way that no auth flows are attempted, in which case, the result list will be empty.

Right enums by default cannot be null, Though we do have ensure ourselves that it's set correctly. (Which can assume is done here)

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
mvanchaa: Just wanted to differentiate the Enumerable result and List result so used `resultList`.

kyle-rader (approved):

reillysiemens (approved):

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
kyle-rader: Could clean up more in the future - you could tack on the `.ToList()` to the `GetTokenAsync()` so you only have 1 `results` variable and it is a List.

kyle-rader: Or an Array

PR #111, opened 2022-06-17
Send AuthFlow Events from the AzureAuth CLI

mvanchaa:

Based on the comments on PR #106 , this is the third PR created.

This PR must be reviewed only after reviewing #109 .

Changes:
1. Send telemetry events from the `AuthFlowResult` s returned to the CLI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Environment variable `AZUREAUTH_CACHE_FILE` and option `--cache` to support a custom cache location on Windows.
+- Added Integrated Windows Authentication functionality as one of the auth flows.
+
 ## [0.3.1] - 2022-06-07
 ### Fixed
 - Fixed a bug where the tenant and resource ids were swapped in the telemetry events.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Environment variable `AZUREAUTH_CACHE_FILE` and option `--cache` to support a custom cache location on Windows.
 - Added Integrated Windows Authentication functionality as one of the auth flows.
+- Removed the usage of AuthTypes. Interactive/Silent token fetching scenario is captured under the boolean property `IsSilent` in TokenResult Class.
+- Send custom telemetry events for each AuthFlow.
 
 ## [0.3.1] - 2022-06-07
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Environment variable `AZUREAUTH_CACHE_FILE` and option `--cache` to support a custom cache location on Windows.
 - Added Integrated Windows Authentication functionality as one of the auth flows.
-- Removed the usage of AuthTypes. Interactive/Silent token fetching scenario is captured under the boolean property `IsSilent` in TokenResult Class.
 - Send custom telemetry events for each AuthFlow.
 
 ## [0.3.1] - 2022-06-07
--- a/src/AzureAuth.Test/CommandMainTest.cs
+++ b/src/AzureAuth.Test/CommandMainTest.cs
@@ -620,7 +620,8 @@ invalid_key = ""this is not a valid alias key""
         /// Test to generate event data from an authflow result with token result and no errors.
         /// </summary>
         [Test]
-        public void TestGenerateEvent_From_AuthFlowResult_With_TokenResult_And_NullErrors()
+        public void TestGenerateEvent_From_AuthFlowResult_With_TokenResult_And_Null_Errors()
+
         {
             // TODO
             var tokenResultCorrelationID = Guid.NewGuid();
--- a/src/AzureAuth.Test/CommandMainTest.cs
+++ b/src/AzureAuth.Test/CommandMainTest.cs
@@ -13,6 +13,8 @@ namespace Microsoft.Authentication.AzureAuth.Test
     using Microsoft.Authentication.MSALWrapper;
     using Microsoft.Extensions.DependencyInjection;
     using Microsoft.Extensions.Logging;
+    using Microsoft.Identity.Client;
+    using Microsoft.IdentityModel.JsonWebTokens;
     using Microsoft.Office.Lasso.Interfaces;
     using Microsoft.Office.Lasso.Telemetry;
 
@@ -27,6 +29,7 @@ namespace Microsoft.Authentication.AzureAuth.Test
     /// </summary>
     internal class CommandMainTest
     {
+        private const string FakeToken = "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsInJoIjoieHh4IiwieDV0IjoieHh4Iiwia2lkIjoieHh4In0.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYXVkIjoiMTExMTExMTEtMTExMS0xMTExLTExMTEtMTExMTExMTExMTExIiwiaWF0IjoxNjE3NjY0Mjc2LCJuYmYiOjE2MTc2NjQyNzYsImV4cCI6MTYxNzY2ODE3NiwiYWNyIjoiMSIsImFpbyI6IllTQjBiM1JoYkd4NUlHWmhhMlVnYTJWNUlDTWtKVjQ9Iiwic2NwIjoidXNlcl9pbXBlcnNvbmF0aW9uIiwidW5pcXVlX25hbWUiOiJreXJhZGVyQG1pY3Jvc29mdC5jb20iLCJ1cG4iOiJreXJhZGVyQG1pY3Jvc29mdC5jb20iLCJ2ZXIiOiIxLjAifQ.bNc3QlL4zIClzFqH68A4hxsR7K-jabQvzB2EodgujQqc0RND_VLVkk2h3iDy8so3azN-964c2z5AiBGY6PVtWKYB-h0Z_VnzbebhDjzPLspEsANyQxaDX_ugOrf7BerQOtILWT5Vqs-A3745Bh0eTDFZpobmeENpANNhRE-yKwScjU8BDY9RimdrA2Z00V0lSliUQwnovWmtfdlbEpWObSFQAK7wCcNnUesV-jNZAUMrDkmTItPA9Z1Ks3NUbqdqMP3D6n99sy8DxQeFmbNQGYocYqI7QH24oNXODq0XB-2zpvCqy4T2jiBLgN_XEaZ5zTzEOzztpgMIWH1AUvEIyw";
         private const string InvalidTOML = @"[invalid TOML"; // Note the missing closing square bracket here.
         private const string CompleteAliasTOML = @"
 [alias.contoso]
@@ -60,6 +63,7 @@ invalid_key = ""this is not a valid alias key""
         private MemoryTarget logTarget;
         private Mock<ITokenFetcher> tokenFetcherMock;
         private Mock<IEnv> envMock;
+        private Mock<ITelemetryService> telemetryServiceMock;
 
         /// <summary>
         /// The setup.
@@ -85,6 +89,7 @@ invalid_key = ""this is not a valid alias key""
             this.tokenFetcherMock = new Mock<ITokenFetcher>(MockBehavior.Strict);
 
             this.envMock = new Mock<IEnv>(MockBehavior.Strict);
+            this.telemetryServiceMock = new Mock<ITelemetryService>(MockBehavior.Strict);
 
             // Environment variables should be null by default.
             this.envMock.Setup(env => env.Get(It.IsAny<string>())).Returns((string)null);
@@ -101,6 +106,7 @@ invalid_key = ""this is not a valid alias key""
                 .AddSingleton(this.fileSystem)
                 .AddSingleton<ITokenFetcher>(this.tokenFetcherMock.Object)
                 .AddSingleton<IEnv>(this.envMock.Object)
+                .AddSingleton<ITelemetryService>(this.telemetryServiceMock.Object)
                 .AddTransient<CommandMain>()
                 .BuildServiceProvider();
         }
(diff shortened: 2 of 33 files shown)


src/AzureAuth/CommandMain.cs
```diff
@@ -217,6 +222,46 @@ public static string PrefixedPromptHint(string promptHint)
...
+
+            var eventData = new EventData();
+            eventData.Add("auth_mode", result.AuthFlowName);
+            eventData.Add("success", result.Success);
+
+            var correlationIDs = new List<string>();
+
+            if (result.Errors.Any())
+            {
+                var error_messages = JsonSerializer.Serialize(result.Errors.Select(error => ExceptionsExtensions.ToFormattedString(error)));
```
mvanchaa, line 245: Is this readable? Or should I break it into two lines?

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -503,6 +509,167 @@ public void TestCacheFileOptionWithWindowsAbsolutePath()
...
+            // Act
+            var eventData = subject.GenerateEventData(authFlowResult);
+
+            // Assert
+            eventData.Properties.Should().NotContainKey("msal_correlation_ids");
+            eventData.Properties.Should().NotContainKey("error_messages");
+            eventData.Measures.Should().NotContainKey("token_validity_hours");
+            eventData.Properties.Should().NotContainKey("is_silent");
+
+            eventData.Properties.TryGetValue("auth_mode", out string authMode);
```
reillysiemens, line 546: Don't quote me on this, but from looking at the [Fluent Assertions docs](https://fluentassertions.com/dictionaries/) I _think_ you might be able to write this as
```csharp
eventData.Properties.Should().Contain("auth_mode", "AuthFlowName");
```

mvanchaa: done.

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -503,6 +509,140 @@ public void TestCacheFileOptionWithWindowsAbsolutePath()
...
```
kyle-rader, line 546: Shouldn't this be `authflow_name` ? This is where tragedy of "AuthMode" enum comes into play. Right now `auth_mode` is an enum that has values that represent different things and we have tried to store too much in it. 

I think we want to refactor how we use auth_mode. Now that we have the AuthFlowName property on the AuthFlow Result itself. All we want on the TokenResult is a bool - `Interactive`. And we should be able to delete the AuthMode enum all together, and instead of setting it to different values, it is only ever set to true or false.

kyle-rader (changes requested):
I've left one main comment so far about refactoring AuthMode away, and replacing it with a boolean Property on TokenResult. (Now that we have the type of auth flow tracked as the auth flow name).

src/AzureAuth.Test/CommandMainTest.cs
kyle-rader: Sorry I was referring to "auth Mode" above but was meaning to talk about "Auth Type".

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -503,6 +509,140 @@ public void TestCacheFileOptionWithWindowsAbsolutePath()
...
+            eventData.Properties.Should().Contain("is_silent", "True");
+            eventData.Properties.Should().Contain("error_messages", "[\"An MSAL Service Exception message\",\"An MSAL UI Required Exception message\"]");
+            eventData.Measures.Should().ContainKey("token_validity_hours");
+        }
+
+        /// <summary>
+        /// Test to generate event data from an authflow result with token result and no errors.
+        /// </summary>
+        [Test]
+        public void TestGenerateEvent_From_AuthFlowResult_With_TokenResult_And_NullErrors()
```
reillysiemens, line 623:
```suggestion
        public void TestGenerateEvent_From_AuthFlowResult_With_TokenResult_And_Null_Errors()

```

src/MSALWrapper/TokenResult.cs
```diff
@@ -32,6 +32,22 @@ public enum AuthType
         IntegratedWindowsAuthenticationFlow,
     }
 
+    /// <summary>
```
kyle-rader, line 35: My other main comment is about this - we should not need this, and instead can better encode the concept of interaction by using better types.

src/AzureAuth/CommandMain.cs
reillysiemens: This is fine, but I think this is the limit (for me anyway). 😅

src/MSALWrapper/TokenResult.cs
```diff
@@ -32,6 +32,22 @@ public enum AuthType
...
+        /// Checks if authType is silent.
+        /// </summary>
+        /// <param name="authType">The auth type.</param>
+        /// <returns>The <see cref="bool"/>.</returns>
+        public static bool IsSilent(this AuthType authType)
+        {
+            return authType == AuthType.Silent;
+        }
+    }
+
```
kyle-rader, line 50: My other main comment is about this - we should not need this, and instead can better encode the concept of interaction by using better types.

kyle-rader (changes requested):

src/AzureAuth/CommandMain.cs
```diff
@@ -218,6 +222,46 @@ public static string PrefixedPromptHint(string promptHint)
...
+
+            var eventData = new EventData();
+            eventData.Add("authflow", result.AuthFlowName);
+            eventData.Add("success", result.Success);
+
+            var correlationIDs = new List<string>();
+
+            if (result.Errors.Any())
+            {
+                var error_messages = JsonSerializer.Serialize(result.Errors.Select(error => ExceptionsExtensions.ToFormattedString(error)));
```
kyle-rader, line 245: I don't think we need JSON serialization here. We can use the existing ExceptionListToStringConverter class and method without modiciation to turn this list of errors, into a collecting string (with 1 error message per line) which is easy to parse in Kusto.

src/AzureAuth/CommandMain.cs
```diff
@@ -218,6 +222,46 @@ public static string PrefixedPromptHint(string promptHint)
...
+                eventData.Add("error_messages", error_messages);
+                correlationIDs = ExceptionListToStringConverter.ExtractCorrelationIDsFromException(result.Errors);
+            }
+
+            if (result.Success)
+            {
+                correlationIDs.Add(result.TokenResult.CorrelationID.ToString());
+                eventData.Add("token_validity_hours", result.TokenResult.ValidFor.Hours);
```
kyle-rader, line 253:
```suggestion
                eventData.Add("token_valid_for_minutes", result.TokenResult.ValidFor.TotalMinutes);

```

kyle-rader: Need to use the `TotalMinutes` otherwise it doesn't compute the entire time span into that unit.

kyle-rader (changes requested):

src/AzureAuth/CommandMain.cs
```diff
@@ -86,8 +89,9 @@ public CommandMain(CommandExecuteEventData eventData, ILogger<CommandMain> logge
         /// <param name="fileSystem">The file system.</param>
         /// <param name="env">The environment interface.</param>
         /// <param name="authFlow">An injected <see cref="IAuthFlow"/> (defined for testability).</param>
-        public CommandMain(CommandExecuteEventData eventData, ILogger<CommandMain> logger, IFileSystem fileSystem, IEnv env, IAuthFlow authFlow)
-            : this(eventData, logger, fileSystem, env)
+        /// <param name="telemetryService">The telemetry service.</param>
```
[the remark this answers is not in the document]
shalinikhare27: Interesting to see that stylecop did not catch this.

src/AzureAuth/CommandMain.cs
```diff
@@ -218,6 +222,47 @@ public static string PrefixedPromptHint(string promptHint)
...
+
+            var eventData = new EventData();
+            eventData.Add("authflow", result.AuthFlowName);
+            eventData.Add("success", result.Success);
+
+            var correlationIDs = new List<string>();
+
+            if (result.Errors.Any())
+            {
+                var error_messages = JsonSerializer.Serialize(result.Errors.Select(error => ExceptionsExtensions.ToFormattedString(error)));
```
shalinikhare27, line 245: Why are we serializing this since `ExceptionsExtensions.ToFormattedString(error)` does return a string?

src/MSALWrapper/ExceptionListToStringConverter.cs
```diff
@@ -6,46 +6,48 @@ namespace Microsoft.Authentication.MSALWrapper
...
-                                .Replace("\n", string.Empty)
-                                .Replace("\r", string.Empty)
-                                .Replace("\t", string.Empty);
-                return $"{ex.GetType()}: {message}";
-            }
-            else
+            foreach (Exception exception in exceptions)
             {
-                return "null";
+                if (exception.GetType() == typeof(MsalServiceException))
```
shalinikhare27, line 32: `MsalServiceException` and `MsalUiRequiredException` are the only exceptions we collect correlation IDs for? Is that on purpose since we do catch exceptions such as `MsalClientException`, etc.

src/AzureAuth/CommandMain.cs
```diff
@@ -218,6 +222,47 @@ public static string PrefixedPromptHint(string promptHint)
...
+                eventData.Add("error_messages", error_messages);
+                correlationIDs = ExceptionListToStringConverter.ExtractCorrelationIDsFromException(result.Errors);
+            }
+
+            if (result.Success)
+            {
+                correlationIDs.Add(result.TokenResult.CorrelationID.ToString());
+                eventData.Add("token_valid_for_minutes", result.TokenResult.ValidFor.TotalMinutes);
+
```
shalinikhare27, line 254: [nit]: extra line

src/AzureAuth/CommandMain.cs
```diff
@@ -397,18 +442,20 @@ private int GetToken()
                     try
                     {
                         results = authFlowExecutor.GetTokenAsync().Result.ToArray();
```
shalinikhare27, line 444: Why are we using array instead of the returned list?

src/MSALWrapper/ExceptionListToStringConverter.cs
mvanchaa: Yes, those are the only exceptions with correlation IDs.

src/AzureAuth/CommandMain.cs
mvanchaa: I changed it to use the previous method which takes the exceptions messages and converts them to line separated messages.

src/AzureAuth/CommandMain.cs
```diff
@@ -218,6 +222,46 @@ public static string PrefixedPromptHint(string promptHint)
             return $"{PromptHintPrefix}: {promptHint}";
         }
 
+        /// <summary>
+        /// Generates event data from the AuthFlowResult.
+        /// </summary>
+        /// <param name="result">The AuthFlowResult.</param>
+        /// <returns>The event data.</returns>
+        public EventData GenerateEventData(AuthFlowResult result)
```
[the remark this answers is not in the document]
mvanchaa: done

src/MSALWrapper/TokenResultExtensions.cs
```diff
@@ -9,15 +9,14 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
...
         /// <param name="result">The result.</param>
-        /// <param name="authType">The auth type.</param>
-        internal static void SetAuthenticationType(this TokenResult result, AuthType authType)
+        internal static void SetSilent(this TokenResult result)
         {
             if (result != null)
             {
-                result.AuthType = authType;
+                result.Silent = true;
             }
```
shalinikhare27, line 20: Super confused here, why is `result != null` means token is silent? You could have a non-null result with web as well which is interactive.

src/AzureAuth/CommandMain.cs
mvanchaa: We return IEnumerable<T>. So we can either convert the results to List or Array. Array is chosen here since we do not need to manipulate (add/update/remove) the results as we only need to access the contents of it.

src/MSALWrapper/TokenResultExtensions.cs
mvanchaa: This method is only used to set the tokenResult's `Silent` property to true. The default value of that property is false. 
This method is called from each authflow when we fetch the token silently. A non-null tokenResult will always be interactive by default.

mvanchaa: This is a replacement for the method - `SetAuthenticationStatus(result, AuthType)`. But in this PR, the `AuthType`s are removed. It is replaced by a boolean `Silent` which represents whether the token is fetched silently or not.

src/AzureAuth/CommandMain.cs
```diff
@@ -417,7 +464,8 @@ private int GetToken()
                 }
 
                 var tokenResult = successfulResult.TokenResult;
-                this.eventData.Add("auth_type", $"{tokenResult.AuthType}");
+                this.eventData.Add("silent", tokenResult.Silent);
```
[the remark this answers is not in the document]
mvanchaa: Silent is a boolean value. So if it's false, it would mean the token is not fetched silently.

src/AzureAuth/CommandMain.cs
```diff
@@ -444,6 +494,21 @@ private int GetToken()
...
+        private void SendAuthFlowTelemetryEvents(AuthFlowResult[] results)
+        {
+            if (results != null)
+            {
+                Parallel.ForEach(results, result =>
+                {
+                    var eventData = this.AuthFlowEventData(result);
+                    if (eventData != null)
+                    {
+                        this.telemetryService.SendEvent($"authflow_{result.AuthFlowName}", eventData);
```
shalinikhare27, line 506: Since we are using parallel foreach here, let's ensure that `SendEvent` is thread safe. For instance we have 2 threads simultaneously tried to send events then both of them should be able to. It shouldn't be that one succeeds and the other fails. It will result in loss of events in your telemetry.

src/AzureAuth/CommandMain.cs
```diff
@@ -444,6 +494,21 @@ private int GetToken()
             return 0;
         }
 
+        private void SendAuthFlowTelemetryEvents(AuthFlowResult[] results)
+        {
+            if (results != null)
+            {
+                Parallel.ForEach(results, result =>
```
shalinikhare27, line 501: Also, the results gathered are not that huge in number, so there might not be a lot of benefit to do parallel foreach here.  Sometimes creating overhead of creating multiple threads for a small collection can be larger than any benefit achieved by processing it in parallel. The processing of each result is not very expensive since you just create an event data object from a result object.

src/MSALWrapper/TokenResult.cs
```diff
@@ -87,9 +68,9 @@ public JsonWebToken JWT
         public TimeSpan ValidFor { get; internal set; }
 
         /// <summary>
-        /// Gets the auth type.
+        /// Gets a value indicating whether this token was acquired silently or not.
         /// </summary>
-        public AuthType AuthType { get; internal set; }
+        public bool Silent { get; internal set; }
```
shalinikhare27, line 73: I prefer having a bool property as a verb so that the intention is clear. Consider using `IsSilent`.
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members

src/MSALWrapper/TokenResultExtensions.cs
shalinikhare27: Okay! I saw it now. You are only using it when we get token using `AcquireTokenSilent` or IWA. I have been scrolling up and down to understand the code. A lot of new stuff in CommandMain is implemented below. Is there a reason why we did not add this specific work of change AuthType to Silent or not as a PR in itsself?

shalinikhare27 (changes requested):
Waiting to address some comments

src/MSALWrapper/TokenResultExtensions.cs
mvanchaa: Yea this was a part of the current feature. I thought of removing the AuthType in a different PR but I got feedback about it on this PR itself so I made changes here.

src/AzureAuth/CommandMain.cs
mvanchaa: That's a good question, @shalinikhare27. I think they are thread-safe. Because when this method is called, Lasso asynchronously sends event in a separate new process. @kyle-rader please correct me if I'm wrong.

src/AzureAuth/CommandMain.cs
mvanchaa: I think it's a good to have even though it doesn't improve the performance much here.

src/AzureAuth/CommandMain.cs
kyle-rader: The only part that may have thread safety concerns is the logger instance, though I believe NLog loggers are generally thread safe.

kyle-rader (changes requested):

src/AzureAuth/CommandMain.cs
```diff
@@ -397,18 +441,20 @@ private int GetToken()
...
-                var errors = results.SelectMany(result => result.Errors).ToArray();
-                this.eventData.Add("error_list", ExceptionListToStringConverter.Execute(errors));
-                this.eventData.Add("error_count", errors.Length);
-                this.eventData.Add("authflow_count", results.Length);
+                if (results != null)
+                {
+                    var errors = results.SelectMany(result => result.Errors).ToArray();
+                    this.eventData.Add("error_count", errors.Length);
+                    this.eventData.Add("authflow_count", results.Length);
+                }
```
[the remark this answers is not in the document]
mvanchaa: Fair point.

kyle-rader (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Environment variable `AZUREAUTH_CACHE_FILE` and option `--cache` to support a custom cache location on Windows.
 - Added Integrated Windows Authentication functionality as one of the auth flows.
+- Removed the usage of AuthTypes. Interactive/Silent token fetching scenario is captured under the boolean property `IsSilent` in TokenResult Class.
+- Send custom telemetry events for each AuthFlow.
```
kyle-rader, line 12:
```suggestion
- Send custom telemetry events for each AuthFlow.
```

kyle-rader (changes requested):
1 comment on change log - the ChangeLog is for end users, so changes in internal code or things not directly applicable to an end user understanding if this is a risk for them to pick up, can be left out.

shalinikhare27 (approved):

shalinikhare27 (approved):

reillysiemens (approved):

kyle-rader (approved):

PR #115, opened 2022-06-17
Updating install scripts to no longer create latest

Haard30:
As a part of this PR, following changes have been made in the install scripts (install.sh and install.ps1)

1. Separate the install logic into 2 functions - installing pre-0.4.0 version and post-0.4.0 version
2. Keep the current installation script logic in pre-0.4.0 
3. For post-0.4.0 :
- Update the install script to stop creating a latest directory and instead create a release version directory
- Add a ` -NoUpdatePath ` flag to help avoid modifications to the global $PATH when the flag is provided
-  Improve verbose output by specifying whether we are Appending/Removing from PATH
- Clean up previously added installation path (for currently installed azurauth) including default path, custom path
- Append our newly installed version's path to PATH (when` -NoUpdatePath ` is not provided)
4. Use setx instead of set-item command to signal PATH changes to the OS
5. Version determination logic to call respective functions
6. For install.ps1 - minor refactoring including inverting the null equality comparison for better coding standards 

Changes in release.yml
1. Update the release.yml to improve windows packaging i.e. removing the extra directory hierarchy from the windows archive

### DIFF
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -1,6 +1,6 @@
 # Enable a default -Verbose flag for debug output.
 [CmdletBinding()]
-Param()
+Param([switch] $NoUpdatePath)
 
 # Halt script execution at the first failed command.
 $script:ErrorActionPreference='Stop'
@@ -149,35 +149,39 @@ function Install-Post-0-4-0 {
     Write-Verbose "Removing ${zipFile}"
     Remove-Item -Force $zipFile
 
-    $registryPath = 'Registry::HKEY_CURRENT_USER\Environment'
-    $currentPath = (Get-ItemProperty -Path $registryPath -Name PATH -ErrorAction SilentlyContinue).Path
-    $azureauthSource = (get-command azureauth -ErrorAction SilentlyContinue).Source
-    $azureauthSourceParent = if($azureauthSource -ne $null) {
-        (get-item $azureauthSource).Directory.Parent.FullName }
-    
-    $newPath = "";
+    if(!$NoUpdatePath)
+    {
+        $registryPath = 'Registry::HKEY_CURRENT_USER\Environment'
+        $currentPath = (Get-ItemProperty -Path $registryPath -Name PATH -ErrorAction SilentlyContinue).Path
+        $azureauthSource = (get-command azureauth -ErrorAction SilentlyContinue).Source
+        $azureauthSourceParent = if($azureauthSource -ne $null) {
+            (get-item $azureauthSource).Directory.Parent.FullName }
+        
+        $newPath = "";
 
-    if (($currentPath -ne $null) -And ($currentPath.Contains($azureauthDirectory) -Or $currentPath.Contains($azureauthSourceParent))) {
-        $paths = $currentPath.Split(";")
-        $pathArr = @()
-        ForEach($path in $paths){
-            if(!(($path.Equals("")) -Or ($path.Contains($azureauthDirectory)) -Or (($azureauthSourceParent -ne $null) -And $azureauthSourceParent.Contains($path)))){
-                $pathArr += "${path}"
+        if (($currentPath -ne $null) -And ($currentPath.Contains($azureauthDirectory) -Or $currentPath.Contains($azureauthSourceParent))) {
+            $paths = $currentPath.Split(";")
+            $pathArr = @()
+            ForEach($path in $paths){
+                if(!(($path.Equals("")) -Or ($path.Contains($azureauthDirectory)) -Or (($azureauthSourceParent -ne $null) -And $azureauthSourceParent.Contains($path)))){
+                    $pathArr += "${path}"
+                }
+            }
+            $pathArr += "${versionPath}"
+            $newPath = $pathArr -join ";"
+        }
+        else {
+            Write-Verbose "Updating `$PATH to include ${versionPath}"
+            if ($currentPath -Match $null) {
+                $newPath = "${versionPath}"
+            } else {
+                $newPath = "${currentPath};${versionPath}"
             }
         }
-        $pathArr += "${versionPath}"
-        $newPath = $pathArr -join ";"
-    }
-    else {
-        Write-Verbose "Updating `$PATH to include ${versionPath}"
-        if ($currentPath -Match $null) {
-            $newPath = "${versionPath}"
-        } else {
-            $newPath = "${currentPath};${versionPath}"
-        }
+
+        Set-ItemProperty -Path $registryPath -Name PATH -Value $newPath 
     }
 
-    Set-ItemProperty -Path $registryPath -Name PATH -Value $newPath 
 
     Write-Output "Installed azureauth $version!"
 
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -151,7 +151,10 @@ function Install-Post-0-4-0 {
     Write-Verbose "Removing ${zipFile}"
     Remove-Item -Force $zipFile
 
-    if (!$NoUpdatePath) {
+    if ($NoUpdatePath) {
+        Write-Verbose "Not updating `$env:PATH"
+    }
+    else{
         $registryPath = 'Registry::HKEY_CURRENT_USER\Environment'
         $currentPath = (Get-ItemProperty -Path $registryPath -Name PATH -ErrorAction SilentlyContinue).Path
         $currentAzureauth = (get-command azureauth -ErrorAction SilentlyContinue).Source
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -153,17 +153,17 @@ function Install-Post-0-4-0 {
     if(!$NoUpdatePath) {
         $registryPath = 'Registry::HKEY_CURRENT_USER\Environment'
         $currentPath = (Get-ItemProperty -Path $registryPath -Name PATH -ErrorAction SilentlyContinue).Path
-        $azureauthSource = (get-command azureauth -ErrorAction SilentlyContinue).Source
-        $azureauthSourceParent = if($azureauthSource -ne $null) {
-            (get-item $azureauthSource).Directory.Parent.FullName }
+        $currentAzureauth = (get-command azureauth -ErrorAction SilentlyContinue).Source
+        $currentAzureauthParent = if($currentAzureauth -ne $null) {
+            (get-item $currentAzureauth).Directory.Parent.FullName }
         
         $newPath = "";
 
-        if (($currentPath -ne $null) -And ($currentPath.Contains($azureauthDirectory) -Or $currentPath.Contains($azureauthSourceParent))) {
+        if (($currentPath -ne $null) -And ($currentPath.Contains($azureauthDirectory) -Or $currentPath.Contains($currentAzureauthParent))) {
             $paths = $currentPath.Split(";")
             $pathArr = @()
             ForEach($path in $paths){
-                if(!(($path.Equals("")) -Or ($path.Contains($azureauthDirectory)) -Or (($azureauthSourceParent -ne $null) -And $azureauthSourceParent.Contains($path)))){
+                if(!(($path.Equals("")) -Or ($path.Contains($azureauthDirectory)) -Or (($currentAzureauthParent -ne $null) -And $currentAzureauthParent.Contains($path)))){
                     $pathArr += "${path}"
                 }
             }
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -113,7 +113,7 @@ function Install-Post-0-4-0 {
     } else {
         $Env:AZUREAUTH_INSTALL_DIRECTORY
     }
-    $extractedDirectory = ([System.IO.Path]::Combine($azureauthDirectory, $releaseName))
+    
     $targetDirectory = ([System.IO.Path]::Combine($azureauthDirectory, $version))
     $zipFile = ([System.IO.Path]::Combine($azureauthDirectory, $releaseFile))
 
@@ -146,9 +146,7 @@ function Install-Post-0-4-0 {
 
     Write-Verbose "Extracting ${zipFile} to ${targetDirectory}"
     Add-Type -AssemblyName System.IO.Compression.FileSystem
-    [System.IO.Compression.ZipFile]::ExtractToDirectory($zipFile, $azureauthDirectory)
-
-    Rename-Item $extractedDirectory $targetDirectory
+    [System.IO.Compression.ZipFile]::ExtractToDirectory($zipFile, $targetDirectory)
 
     Write-Verbose "Removing ${zipFile}"
     Remove-Item -Force $zipFile
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -8,17 +8,12 @@ $script:ErrorActionPreference='Stop'
 # We don't currently have good cross-platform options for determining the latest release version, so we require
 # knowledge of the specific target version, which the user should set as an environment variable.
 $version = $Env:AZUREAUTH_VERSION
-#$version = "0.3.1"
+
 if ([string]::IsNullOrEmpty($version)) {
     Write-Error 'No $AZUREAUTH_VERSION specified, unable to download a release'
 }
 
-
-
-function V1-Install-With-Latest{
-    param()
-
-    Write-Output "Using the Install with latest version"
+function Install-Pre-0-4-0{
 
     $repo = if ([string]::IsNullOrEmpty($Env:AZUREAUTH_REPO)) { 'AzureAD/microsoft-authentication-cli' } else { $Env:AZUREAUTH_REPO }
     $releaseName = "azureauth-${version}-win10-x64"
@@ -37,9 +32,9 @@ function V1-Install-With-Latest{
     Write-Verbose "Creating ${azureauthDirectory}"
     $null = New-Item -ItemType Directory -Force -Path $azureauthDirectory
 
-#    Write-Verbose "Downloading ${releaseUrl} to ${zipFile}"
-#    $client = New-Object System.Net.WebClient
-#    $client.DownloadFile($releaseUrl, $zipFile)
+    Write-Verbose "Downloading ${releaseUrl} to ${zipFile}"
+    $client = New-Object System.Net.WebClient
+    $client.DownloadFile($releaseUrl, $zipFile)
 
     # A running instance of azureauth can cause installation to fail, so we try to kill any running instances first.
     # We suppress taskkill output here because this is a best effort attempt and we don't want the user to see its output.
@@ -101,13 +96,9 @@ function V1-Install-With-Latest{
 
     Write-Output "Installed azureauth $version!"
 
-
 }
 
-function V2-No-Latest{
-    param()
-
-    Write-Output "Using the Install No Latest Version"
+function Install-Post-0-4-0{
 
     $repo = if ([string]::IsNullOrEmpty($Env:AZUREAUTH_REPO)) { 'AzureAD/microsoft-authentication-cli' } else { $Env:AZUREAUTH_REPO }
     $releaseName = "azureauth-${version}-win10-x64"
@@ -120,17 +111,16 @@ function V2-No-Latest{
         $Env:AZUREAUTH_INSTALL_DIRECTORY
     }
     $extractedDirectory = ([System.IO.Path]::Combine($azureauthDirectory, $releaseName))
-#    $latestDirectory = ([System.IO.Path]::Combine($azureauthDirectory, "latest"))
     $versionPath = ([System.IO.Path]::Combine($azureauthDirectory, $version))
     $zipFile = ([System.IO.Path]::Combine($azureauthDirectory, $releaseFile))
 
     Write-Verbose "Creating ${azureauthDirectory}"
     $null = New-Item -ItemType Directory -Force -Path $azureauthDirectory
 
-#   TODO uncomment downloading
-#    Write-Verbose "Downloading ${releaseUrl} to ${zipFile}"
-#    $client = New-Object System.Net.WebClient
-#    $client.DownloadFile($releaseUrl, $zipFile)
+
+    Write-Verbose "Downloading ${releaseUrl} to ${zipFile}"
+    $client = New-Object System.Net.WebClient
+    $client.DownloadFile($releaseUrl, $zipFile)
 
     # A running instance of azureauth can cause installation to fail, so we try to kill any running instances first.
     # We suppress taskkill output here because this is a best effort attempt and we don't want the user to see its output.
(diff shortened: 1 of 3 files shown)


install/install.ps1
```diff
@@ -8,87 +8,196 @@ $script:ErrorActionPreference='Stop'
 # We don't currently have good cross-platform options for determining the latest release version, so we require
 # knowledge of the specific target version, which the user should set as an environment variable.
 $version = $Env:AZUREAUTH_VERSION
+#$version = "0.3.1"
```
kyle-rader, line 11: This looks like an accidental commit that we wont want to keep around. Note that VS Code has a pretty nice interface for staging hunks (a chunk of code) within a file, line by line - which lets you not commit testing lines you might have locally.

install/install.ps1
```diff
@@ -8,87 +8,196 @@ $script:ErrorActionPreference='Stop'
...
-    Write-Verbose "Removing pre-existing extracted directory at ${extractedDirectory}"
-    Remove-Item -Force -Recurse $extractedDirectory
-}
+function V1-Install-With-Latest{
+    param()
 
-Write-Verbose "Extracting ${zipFile} to ${extractedDirectory}"
-Add-Type -AssemblyName System.IO.Compression.FileSystem
-[System.IO.Compression.ZipFile]::ExtractToDirectory($zipFile, $azureauthDirectory)
+    Write-Output "Using the Install with latest version"
```
reillysiemens, line 21: This isn't _quite_ accurate. We install whatever version was requested to be installed, but we _also_ update the `latest` directory. I would just not print this out.

install/install.ps1
```diff
@@ -8,87 +8,196 @@ $script:ErrorActionPreference='Stop'
...
-    # second to be enough.
-    Start-Sleep -Seconds 1
-}
 
-if (Test-Path -Path $extractedDirectory) {
-    Write-Verbose "Removing pre-existing extracted directory at ${extractedDirectory}"
-    Remove-Item -Force -Recurse $extractedDirectory
-}
+function V1-Install-With-Latest{
+    param()
```
reillysiemens, line 19: Do all PowerShell functions need a `param()`? If there are no parameters can this be left out? :thinking:

install/install.ps1
```diff
@@ -8,87 +8,196 @@ $script:ErrorActionPreference='Stop'
...
+#    }
+
+    Write-Verbose "Removing ${zipFile}"
+    Remove-Item -Force $zipFile
+
+    # Permanently add the latest directory to the current user's $PATH (if it's not already there).
+    # Note that this will only take effect when a new terminal is started.
+    $registryPath = 'Registry::HKEY_CURRENT_USER\Environment'
+    $currentPath = (Get-ItemProperty -Path $registryPath -Name PATH -ErrorAction SilentlyContinue).Path
+    if ($currentPath -NotMatch 'AzureAuth') {
```
kyle-rader, line 184: In disucssion we noted how the curernt if block basically does what we want in this case where we do want to update the path but azureauth.exe isn't on it yet. Something I did not call out however, is how in our future world, `AzureAuth` itself may never be in the path. Since the directory env var could be custom, and the directory we install is going to change to only be the version, and not the full release asset name.

install/install.ps1
```diff
@@ -8,87 +8,196 @@ $script:ErrorActionPreference='Stop'
...
-    # its own DLLs). The OS may take an indeterminate amount of time to clean those up, but so far we've observed 1
-    # second to be enough.
-    Start-Sleep -Seconds 1
-}
 
-if (Test-Path -Path $extractedDirectory) {
-    Write-Verbose "Removing pre-existing extracted directory at ${extractedDirectory}"
-    Remove-Item -Force -Recurse $extractedDirectory
-}
+function V1-Install-With-Latest{
```
reillysiemens, line 18: The `V1` vs. `V2` distinction is a little muddy. Let's just leave this as `Install-With-Latest` and make the opposite be `Install-Without-Latest`. I would also be happy to see this be named something like `Install-Pre-0-4-0` and `Install-Post-0-4-0` or similar if PowerShell allows a function name like that. Thoughts here, @kyle-rader?

install/install.ps1
```diff
@@ -8,87 +8,196 @@ $script:ErrorActionPreference='Stop'
...
+    $extractedDirectory = ([System.IO.Path]::Combine($azureauthDirectory, $releaseName))
+    $latestDirectory = ([System.IO.Path]::Combine($azureauthDirectory, "latest"))
+    $zipFile = ([System.IO.Path]::Combine($azureauthDirectory, $releaseFile))
+
+    Write-Verbose "Creating ${azureauthDirectory}"
+    $null = New-Item -ItemType Directory -Force -Path $azureauthDirectory
+
+#    Write-Verbose "Downloading ${releaseUrl} to ${zipFile}"
+#    $client = New-Object System.Net.WebClient
+#    $client.DownloadFile($releaseUrl, $zipFile)
```
reillysiemens, line 42: I think we still need these lines. Should these be commented out? 🤔

install/install.ps1
```diff
@@ -8,87 +8,196 @@ $script:ErrorActionPreference='Stop'
...
-$registryPath = 'Registry::HKEY_CURRENT_USER\Environment'
-$currentPath = (Get-ItemProperty -Path $registryPath -Name PATH -ErrorAction SilentlyContinue).Path
-if ($currentPath -NotMatch 'AzureAuth') {
-    Write-Verbose "Updating `$PATH to include ${latestDirectory}"
-    $newPath = if ($currentPath -Match $null) {
-        "${latestDirectory}"
+function V2-No-Latest{
+    param()
+
+    Write-Output "Using the Install No Latest Version"
```
reillysiemens, line 110: Ideally this is not something the user sees. We want this kind of thing to be totally transparent to the user. They don't care which installation process we use, only that when it's done the binary is on their `$PATH` and they can run `azureauth`.

install/install.ps1
```diff
@@ -8,87 +8,196 @@ $script:ErrorActionPreference='Stop'
...
+    $versionPath = ([System.IO.Path]::Combine($azureauthDirectory, $version))
+    $zipFile = ([System.IO.Path]::Combine($azureauthDirectory, $releaseFile))
+
+    Write-Verbose "Creating ${azureauthDirectory}"
+    $null = New-Item -ItemType Directory -Force -Path $azureauthDirectory
+
+#   TODO uncomment downloading
+#    Write-Verbose "Downloading ${releaseUrl} to ${zipFile}"
+#    $client = New-Object System.Net.WebClient
+#    $client.DownloadFile($releaseUrl, $zipFile)
```
reillysiemens, line 133: Ah, I see. You're just commenting it out now to avoid doing network requests every time you run and test?

install/install.ps1
```diff
@@ -8,87 +8,196 @@ $script:ErrorActionPreference='Stop'
...
+
+    # We use a directory junction here because not all Windows users will have permissions to create a symlink.
+    # We create this junction with cmd.exe's mklink because it has a stable interface across all active versions of Windows and Windows Server,
+    # while PowerShell's New-Item has breaking changes and doesn't have the -Target param in 4.0 (the default PowerShell on Win Server 2012).
+
+#    Write-Verbose "Linking ${latestDirectory} to ${extractedDirectory}"
+#    cmd.exe /Q /C "mklink /J `"$latestDirectory`" `"$extractedDirectory`"" > $null
+#    if (!$?) {
+#        Write-Error "Linking failed!"
+#    }
```
reillysiemens, line 175: This looks `TODONE` to me. 😆

install/install.ps1
```diff
@@ -8,87 +8,196 @@ $script:ErrorActionPreference='Stop'
...
+        Write-Verbose "Updating `$PATH to include ${versionPath}"
+        $newPath = if ($currentPath -Match $null) {
+            "${versionPath}"
+        } else {
+            "${currentPath};${versionPath}"
+        }
+        Set-ItemProperty -Path $registryPath -Name PATH -Value $newPath
+    }
+    else{
+        #Update the PATH variable
```
reillysiemens, line 194: From the conversation we just had it seems like this is the crux of what you're dealing with right now and that you're making good progress!

install/install.ps1
```diff
@@ -8,87 +8,183 @@ $script:ErrorActionPreference='Stop'
...
+
+    Write-Verbose "Removing ${zipFile}"
+    Remove-Item -Force $zipFile
+
+    $registryPath = 'Registry::HKEY_CURRENT_USER\Environment'
+    $currentPath = (Get-ItemProperty -Path $registryPath -Name PATH -ErrorAction SilentlyContinue).Path
+
+    # Permanently add the latest directory to the current user's $PATH (if it's not already there).
+    # Note that this will only take effect when a new terminal is started.
+    if (($currentPath -ne $null) -And !$currentPath.Contains($azureauthDirectory)) {
```
kyle-rader, line 157: A couple of comments from pairing:
1. It will be slightly easier to read the logic, I think, if we invert the if and else cases, so we do not have to invert each boolean contains check. First check if it's on the PATH already, if so build an updated newPath. Else - just append. 
2. We can factor out the actual Path set command, and scope the if branches to just assigning the new path var. 
3. We can move the azureauth get-Command test above the if, so that we can also look for it in the initial PATH testing (i.e. include whatever directory it may be installed in already if it is. Because in this case - if it's already on the PATH appending will not actually add the current version to the PATH since it is already shadowed.

kyle-rader (changes requested):

install/install.ps1
```diff
@@ -8,87 +8,183 @@ $script:ErrorActionPreference='Stop'
...
+        $newPath = "";
+        $azureauthSource = (get-command azureauth -ErrorAction SilentlyContinue).Source
+        ForEach($path in $paths){
+            if(!(($path.Equals("")) -Or ($path.Contains($azureauthDirectory)) -Or (($azureauthSource -ne $null) -And $azureauthSource.Contains($path)))){
+                if($newPath.Equals("")){
+                    $newPath = -join("${newPath}","${path}")
+                }
+                else{
+                    $newPath = -join("${newPath}",";","${path}")
+                }
```
kyle-rader, line 177: From discussion offline - see if we can copy elements that do not match one of the two dirs to a new array, and then append the version path to that array, and call join at the end to join the array on `;`, instead of string allocating inside the for loop.

kyle-rader (changes requested):

install/install.ps1
```diff
@@ -1,94 +1,203 @@
...
+
+    Write-Verbose "Downloading ${releaseUrl} to ${zipFile}"
+    $client = New-Object System.Net.WebClient
+    $client.DownloadFile($releaseUrl, $zipFile)
+
+    # A running instance of azureauth can cause installation to fail, so we try to kill any running instances first.
+    # We suppress taskkill output here because this is a best effort attempt and we don't want the user to see its output.
+    # Here, Get-Process is used to first determine whether there is an existing azureauth process. If there is, kill the existing process first.
+    $ProcessCheck = Get-Process -Name azureauth -ErrorAction SilentlyContinue -ErrorVariable ProcessError
+    if ($null -ne $ProcessCheck)
```
goagain, line 44: In very old school style C we use 
```C
if (NULL == something)
```
because if developer uses missed a `=` in the comparison the statement 
```C
if (something = NULL)
```
will assign `NULL` to the variable without compile errors. It does reduce the probability to make a mistake but it also reduce the readability as well. 
So now my question is in PowerShell, do we still need to reverse the order?

install/install.ps1
```diff
@@ -1,94 +1,203 @@
...
 
-Write-Output "Installed azureauth $version!"
+switch ($version) {
+     { $_ -in "v0.1.0","v0.2.0","v0.3.0","0.3.1" } {
+        Install-Pre-0-4-0
+    }
+    default {
+        Install-Post-0-4-0
+    }
+}
```
goagain, line 203: We don't need to really consider the installation script version. In the pull request #96 , installation has been detoured to the corresponded tag. There for we only need to consider the install script for current version.

install/install.ps1
```diff
@@ -1,94 +1,203 @@
...
+        if (($null -ne $currentPath) -And ($currentPath.Contains($azureauthDirectory) -Or $currentPath.Contains($currentAzureauthParent))) {
+            $paths = $currentPath.Split(";")
+            $pathArr = @()
+            ForEach($path in $paths){
+                if(!(($path.Equals("")) -Or ($path.Contains($azureauthDirectory)) -Or (($null -ne $currentAzureauthParent) -And $currentAzureauthParent.Contains($path)))){
+                    $pathArr += "${path}"
+                }
+                else {
+                    Write-Verbose "Removing '${path}' from `$env:PATH"
+                }
```
goagain, line 175: This code block has less readability than others. Think about make it simple or extract a pattern to an extra method.

goagain (changes requested):

install/install.ps1
goagain: @kyle-rader @reillysiemens , I think this part is worth to discuss.

install/install.ps1
reillysiemens: @goagain: This is not related to the specific C example you've given, but `$null` is intentionally placed on the left because apparently this is common PowerShell best practice. See this [Stack Overflow answer](https://stackoverflow.com/questions/58217913/should-null-be-on-the-left-side-of-the-equality-comparison-eq-with-arrays) and this [PowerShell doc](https://docs.microsoft.com/en-us/powershell/scripting/learn/deep-dives/everything-about-null?view=powershell-7.2#checking-for-null) for more info.

install/install.ps1
reillysiemens: @goagain: In #96 we changed our _documentation_, but we can't change any existing user installations that may have been done _before_ we changed that. We **must** consider the version in the installation script to avoid breaking the user's environment if they have a different version installed globally. This is not ideal and I would like for us to do this only for some transition period (maybe 6 months?).

install/install.ps1
reillysiemens: I am interested in a recommendation to make it easier to read. Is it mainly the consecutive `-Or` patterns that harms readability for you, @goagain? Extracting this to a separate function is possible, but tricky because of the number of local variables referenced.

install/install.ps1
```diff
@@ -1,94 +1,203 @@
...
 [CmdletBinding()]
-Param()
+Param([switch] $NoUpdatePath)
 
 # Halt script execution at the first failed command.
 $script:ErrorActionPreference='Stop'
 
 # We don't currently have good cross-platform options for determining the latest release version, so we require
 # knowledge of the specific target version, which the user should set as an environment variable.
+
```
shalinikhare27, line 10: [nit]: extra line

install/install.ps1
```diff
@@ -1,94 +1,203 @@
...
 $version = $Env:AZUREAUTH_VERSION
 if ([string]::IsNullOrEmpty($version)) {
     Write-Error 'No $AZUREAUTH_VERSION specified, unable to download a release'
 }
 
-$repo = if ([string]::IsNullOrEmpty($Env:AZUREAUTH_REPO)) { 'AzureAD/microsoft-authentication-cli' } else { $Env:AZUREAUTH_REPO }
-$releaseName = "azureauth-${version}-win10-x64"
-$releaseFile = "${releaseName}.zip"
-$releaseUrl = "https://github.com/${repo}/releases/download/${version}/$releaseFile"
+function Install-Pre-0-4-0 {
```
shalinikhare27, line 17: I see a new line after function, is that a powershell thing?

install/install.ps1
```diff
@@ -1,94 +1,203 @@
...
-{
-    Write-Verbose "Stopping any currently running azureauth instances"
-    taskkill /f /im azureauth.exe 2>&1 | Out-Null
-
-    # After killing the process it is still possible for there there to be locks on the files it was using (including
-    # its own DLLs). The OS may take an indeterminate amount of time to clean those up, but so far we've observed 1
-    # second to be enough.
-    Start-Sleep -Seconds 1
-}
+    Write-Verbose "Installing using pre-0.4.0 method"
```
shalinikhare27, line 18: Will we need to change the version name manually in this script? If yes, how do we remember to do it?

install/install.ps1
goagain: > @goagain: This is not related to the specific C example you've given, but `$null` is intentionally placed on the left because apparently this is common PowerShell best practice. See this [Stack Overflow answer](https://stackoverflow.com/questions/58217913/should-null-be-on-the-left-side-of-the-equality-comparison-eq-with-arrays) and this [PowerShell doc](https://docs.microsoft.com/en-us/powershell/scripting/learn/deep-dives/everything-about-null?view=powershell-7.2#checking-for-null) for more info.

Thank you. I realized this is the best practice in PowerShell.

install/install.ps1
reillysiemens: This version represents a specific point in time at which we radically changed the installation logic. We shouldn't need to change it when we release new versions. Ideally this goes away as well, we just need to leave some time for users to migrate to a post-0.4.0 installation (maybe 6 months?).

install/install.ps1
```diff
@@ -1,94 +1,203 @@
...
-        "${currentPath};${latestDirectory}"
+        $Env:AZUREAUTH_INSTALL_DIRECTORY
     }
-    Set-ItemProperty -Path $registryPath -Name PATH -Value $newPath
+
+    $targetDirectory = ([System.IO.Path]::Combine($azureauthDirectory, $version))
+    $zipFile = ([System.IO.Path]::Combine($azureauthDirectory, $releaseFile))
+
+    Write-Verbose "Creating ${azureauthDirectory}"
+    $null = New-Item -ItemType Directory -Force -Path $azureauthDirectory
```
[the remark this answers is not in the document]
reillysiemens: This is being used to suppress output. It is functionally the same as redirecting output to `$null` with
```powershell
thing-that-does-output > $null
```
We used this originally because this was shown in examples that suppressed the output of built-in PowerShell commandlets, but I'm not sure if we could replace it with `> $null`. Never tried. 😅

install/install.ps1
```diff
@@ -1,94 +1,203 @@
...
+        Write-Verbose "Stopping any currently running azureauth instances"
+        taskkill /f /im azureauth.exe 2>&1 | Out-Null
+
+        # After killing the process it is still possible for there there to be locks on the files it was using (including
+        # its own DLLs). The OS may take an indeterminate amount of time to clean those up, but so far we've observed 1
+        # second to be enough.
+        Start-Sleep -Seconds 1
+    }
+
+    if (Test-Path -Path $targetDirectory) {
```
shalinikhare27, line 142: The if-else statement from here might require more commenting or method extraction. I did not totally understand the decisions made here and why.

install/install.sh
```diff
@@ -50,55 +50,163 @@ release_name() {
...
+    verbose "Installing using post-0.4.0 method"
+
+    verbose "Creating ${azureauth_directory}"
+    mkdir -p $azureauth_directory
+
+    verbose "Downloading ${release_url} to ${tarball}"
+    if ! curl -sfL $release_url > $tarball; then
+        error "Failed to download azureauth ${version}"
+        exit 1
+    fi
```
shalinikhare27, line 151: What's the usage of this in PowerShell? I did search before adding this comment and did not find anything substantial.

shalinikhare27 (changes requested):
General questions.

reillysiemens:
If it helps reviewers to demonstrate that, at the very least, these changes don't break installation of pre-0.4.0 versions you can run the following.

#### PowerShell
```powershell
$versions = "v0.1.0", "v0.2.0", "v0.3.0", "0.3.1"
foreach ($version in $versions)
{
    iex "& { $(irm https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/users/haashah/azure_auth_install_bug/install/install.ps1) } -Verbose"
}
```

#### Bash
```shell
for version in "v0.1.0" "v0.2.0" "v0.3.0" "0.3.1"
do
    export AZUREAUTH_VERSION="${version}"
    curl -sL https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/users/haashah/azure_auth_install_bug/install/install.sh | sh
done
```

install/install.ps1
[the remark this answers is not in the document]
shalinikhare27: Why do we need to suppress it? Just curious!

install/install.ps1
shalinikhare27: Perfect! So there is a world on how the logic works pre 0.4.0 and post 0.4.0.

install/install.ps1
```diff
@@ -1,94 +1,203 @@
...
     }
-    Set-ItemProperty -Path $registryPath -Name PATH -Value $newPath
+
+    $targetDirectory = ([System.IO.Path]::Combine($azureauthDirectory, $version))
+    $zipFile = ([System.IO.Path]::Combine($azureauthDirectory, $releaseFile))
+
+    Write-Verbose "Creating ${azureauthDirectory}"
+    $null = New-Item -ItemType Directory -Force -Path $azureauthDirectory
+
+
```
shalinikhare27, line 123: [nit]: extra line

install/install.ps1
```diff
@@ -1,94 +1,203 @@
...
+        Write-Verbose "Updating `$env:PATH to include ${latestDirectory}"
+        $newPath = if ($null -eq $currentPath) {
+            "${latestDirectory}"
+        } else {
+            "${currentPath};${latestDirectory}"
+        }
+        setx PATH $newPath > $null
+    }
+
+    Write-Output "Installed azureauth $version!"
```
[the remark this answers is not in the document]
reillysiemens: Which line? 🤔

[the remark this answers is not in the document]
shalinikhare27: After line `Write-Output "Installed azureauth $version!"`

install/install.sh
reillysiemens: This is `/bin/sh` syntax, not PowerShell. In shells like `/bin/sh` (Bash and ZSH count here too) the `fi` statement closes a corresponding `if` (note that it's just `if` backwards!).

reillysiemens: I'm **not** a fan of this syntax. It's one of my least favorite parts of shell scripting.

shalinikhare27: Fun stuff! Learnt something.

install/install.ps1
Haard30: Sure, I will add some comments to help readability. Just wanted to confirm if it is the Test-Path blocks or the NoUpdatePath flag block that's requiring comment?

reillysiemens: Yes, which part is confusing? 🤔

reillysiemens: Ah, I see. You just wanted some more explanation from this point onward. I've added some comments. Hope those help.

install/install.ps1
```diff
@@ -12,83 +12,196 @@ if ([string]::IsNullOrEmpty($version)) {
...
+    $registryPath = 'Registry::HKEY_CURRENT_USER\Environment'
+    $currentPath = (Get-ItemProperty -Path $registryPath -Name PATH -ErrorAction SilentlyContinue).Path
+    if ($currentPath -NotMatch 'AzureAuth') {
+        Write-Verbose "Updating `$env:PATH to include ${latestDirectory}"
+        $newPath = if ($null -eq $currentPath) {
+            "${latestDirectory}"
+        } else {
+            "${currentPath};${latestDirectory}"
+        }
+        setx PATH $newPath > $null
```
kyle-rader, line 93: I do think it's worth commenting on why it's important to use `setx` here, since it handles signaling the env vars update to explorer.ese (the god parent of everything on windows apparently).

kyle-rader (changes requested):

install/install.ps1
reillysiemens: Done.

kyle-rader (approved):
👌

goagain (approved):

goagain:
🎉🎉🎉🎉🚀🚀🚀🚀🚀🚀

PR #127, opened 2022-07-14
Adding global timeout and timeout warnings

Haard30:
This PR includes:

- Adding a global timeout so that we can provide a reliable contract stating that our CLI will not run beyond a specific time ( From telemetry - 10 minutes should be a safe bet).
- Adding a `--timeout` option to override the global timeout.
- Providing timeout warnings as the the timeout approaches to let the user know that we are waiting on an interactive auth
- Unit test to validate timeout design and behavior of application.

Approach and design decisions:

- Running each auth flow (one at a time) on a separate task and poll that task at regular intervals to know the status of the task. At the same time, we are keeping track of timer for global timeout.
- If timeout occurs, we are returning a timeout error to have better understand our telemetry results.
- Made a thin wrapper around `Stopwatch` for management of timeout (`IStopwatch`). It is good to note that though we are not extensively changing the behavior of `Stopwatch`, this design helps with unit testing and having high code coverage for our tests. Moreover, this also helps `AuthFlowExecutor` to not have the responsibility to maintain the logic of when timeout will occur or what's the remaining time (though fairly simple but also extensible to complex design) and thus helping with Single Responsibility Principle.


mvanchaa (changes requested):

goagain (changes requested):
Generally, I don't think use a static class is a good idea. It makes projects coupling, and also there will be many inherit dependencies.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -73,5 +85,54 @@ public async Task<IEnumerable<AuthFlowResult>> GetTokenAsync()
...
+        /// Run the auth mode in a separate task and
+        /// poll to see if we hit global timeout before the auth flow completes.
+        /// </summary>
+        /// <param name="authFlow">Auth Flow that we are executing.</param>
+        /// <returns>AuthFlowResult containing Error if CLI times out; else actual result of the AuthFlow.</returns>
+        private async Task<AuthFlowResult> GetTokenAndPoll(IAuthFlow authFlow)
+        {
+            var flowResult = Task.Run(() => authFlow.GetTokenAsync());
+            var authFlowName = authFlow.GetType().Name;
+            var oneMinute = TimeSpan.FromMinutes(1);
```
goagain, line 99: Why one minutes. It seems like another magic number.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -73,5 +85,54 @@ public async Task<IEnumerable<AuthFlowResult>> GetTokenAsync()
...
+
+                if (GlobalTimeoutManager.HasTimedout())
+                {
+                    GlobalTimeoutManager.StopTimer();
+                    this.logger.LogWarning("AzureAuth has timed out!");
+                    AuthFlowResult timeoutResult = new AuthFlowResult(null, null, authFlow.GetType().Name);
+                    timeoutResult.Errors.Add(new TimeoutException($"The application has timed out while waiting on {authFlowName}"));
+                    /* Note that though the task running the auth flow will be killed once we return from this method,
+                     * the interactive auth prompt will be killed as we exit the application (possibly due to the way GC works).
+                     */
```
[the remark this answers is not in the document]
Haard30: Regarding static class - I agree with you that in most cases we should avoid. For this case though it made sense to me because of two reasons:
1. Having timeout related stuff in AuthFlowExecutor would violate the Single Responsibility Principle as then that class has to execute the auth flow and also do housekeeping work for timeout.
2. This is one of the instance where a class should be available globally because the timeout itself is 'global'. Static might be a good option that satisfies the need. The other way would be to have a normal class and pass instances and make it singleton in order to avoid being duplicated which is also fine - just a toss up between two approaches.  
Please let me know if Singleton approach looks better and I am missing something that could be very bad with this static class and we can revisit the design in that case.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
Haard30: Yes, I can move that with the other constant and that should provide a better context but in essence that is for waiting before we start giving warning to user

kyle-rader:
Summary from offline discussion:
- avoid global state classes - look at refactoring global timeout manager to implement an ITimeoutManager interface, which you can inject into the class(es) that need to isolate the dependency within exiting code and decouple that dependency from the concrete implementation.

src/MSALWrapper/TaskExecutor.cs
```diff
@@ -31,7 +31,7 @@ internal static async Task<T> CompleteWithin<T>(ILogger logger, TimeSpan timeout
             source.CancelAfter(timeout);
             try
             {
-                logger?.LogDebug($"{taskName} has {timeout.TotalMinutes} minutes to complete before timeout.");
+                // logger?.LogDebug($"{taskName} has {timeout.TotalMinutes} minutes to complete before timeout.");
```
mvanchaa, line 34: You can remove this line if it's not needed instead of commenting.

src/MSALWrapper/ITimeoutManager.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    using System;
+
+    /// <summary>
+    /// Timeout manager interface.
+    /// </summary>
+    public interface ITimeoutManager
+    {
+        /// <summary>
+        /// Determines if the has timed out.
```
mvanchaa, line 14:
```suggestion
        /// Determines if the timer has timed out.

```

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -73,5 +94,60 @@ public async Task<IEnumerable<AuthFlowResult>> GetTokenAsync()
...
+                    this.logger.LogWarning($"Waiting for {authFlowName} authentication. Please look for an interactive auth prompt.");
+                    this.logger.LogWarning($"Timeout in {this.timeoutManager.GetRemainingTime():hh\\:mm\\:ss}");
+                }
+
+                if (this.timeoutManager.HasTimedout())
+                {
+                    this.timeoutManager.StopTimer();
+                    this.logger.LogWarning("AzureAuth has timed out!");
+                    AuthFlowResult timeoutResult = new AuthFlowResult(null, null, authFlow.GetType().Name);
+                    timeoutResult.Errors.Add(new TimeoutException($"The application has timed out while waiting on {authFlowName}"));
```
kyle-rader, line 121:
```suggestion
                    timeoutResult.Errors.Add(new TimeoutException($"Global timeout hit during {authFlowName}"));

```

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -73,5 +94,60 @@ public async Task<IEnumerable<AuthFlowResult>> GetTokenAsync()
...
+            return await flowResult;
+        }
+
+        /// <summary>
+        /// Determines amount of time to wait before polling.
+        /// If global timeout is in less than default delayPeriodForPolling, we would want to wait for a period equal to
+        /// timeout rather than the delayPeriodForPolling.
+        /// </summary>
+        /// <returns>Time to wait before polling.</returns>
+        private TimeSpan DetermineDelayPeriod()
```
kyle-rader, line 140: Good extraction of a helper method.

src/MSALWrapper/Constants.cs
```diff
@@ -20,6 +20,11 @@ public static class Constants
         /// </summary>
         public static readonly Uri AadOAuthRedirectUri = new Uri("urn:ietf:wg:oauth:2.0:oob");
 
+        /// <summary>
```
kyle-rader, line 23: Given the main usage of this value is in the CLI project, this isn't actually a constant inherent to the MSALWrapper - I think we can define this on the CommandMain class itself.

src/MSALWrapper/GlobalTimeoutManager.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Microsoft.Authentication.MSALWrapper
+{
+    using System;
+    using System.Diagnostics;
+
+    /// <summary>
+    /// Manages global timer for the CLI.
+    /// </summary>
+    public class GlobalTimeoutManager : ITimeoutManager
```
kyle-rader, line 12:
```suggestion
    public class TimeoutManager : ITimeoutManager

```

src/MSALWrapper/GlobalTimeoutManager.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        private readonly Stopwatch globalTimer;
```
kyle-rader, line 14:
```suggestion
        private readonly Stopwatch timer;

```

src/MSALWrapper/GlobalTimeoutManager.cs
```diff
@@ -0,0 +1,70 @@
...
+        private readonly TimeSpan globalTimeout;
```
kyle-rader, line 15:
```suggestion
        private readonly TimeSpan timeout;

```

src/MSALWrapper/GlobalTimeoutManager.cs
kyle-rader: 3 naming suggestions for this class.

kyle-rader (changes requested):
I've left a few comments but my 1 big comment is this: Does the `TimeoutManager` class need to exist as an abstraction here? Or could AuthFlowExecutor simply take the TimeSpan value for it's timeout, and internally use a StopWatch directly? We don't really need to test `StopWatch` we can assume it works as expected.

I'm not saying to get rid of it - as the interface allows you to TDD it's usage within the authflowexecutor. I'm more calling out that it's worth acknowledging this abstraction is very thinnly wrapping the StopWatch in order to support Testing. And that's the kind of comment on design trade-offs I'd like to see called out in the Pull Request description ;).

src/AzureAuth/CommandMain.cs
```diff
@@ -120,6 +126,12 @@ public CommandMain(CommandExecuteEventData eventData, ITelemetryService telemetr
         [Option(PromptHintOption, "The prompt hint text for WAM prompts and web mode.", CommandOptionType.SingleValue)]
         public string PromptHint { get; set; }
 
+        /// <summary>
+        /// Gets or sets global Timeout.
+        /// </summary>
+        [Option(TimeoutOption, "Number of seconds for which the CLI should run", CommandOptionType.SingleValue)]
```
kyle-rader, line 132: I know we have some back and forth on whether to use seconds or minutes. Seconds is common, but here almost never make sense, and devs have an easier time thinking in minutes they wait. We are using a double - to allow for sub-minute granularity, and I thought by using doubles, @reillysiemens had agreed that interpreting as minutes was okay here.

I thought of this again because of this suggestion where I want to say "default is 10 minutes" but I don't want to say 10 minutes if the actual value to make that happen is actually `--timeout 600`. You could do "default is 600 seconds" But then I 'd want to add `(10 minutes)`. I'm very against making people do math for no reason :). 
```suggestion
        [Option(TimeoutOption, "The number of seconds the CLI is allowed to wait before timing out.\nDefault: 10 minutes.", CommandOptionType.SingleValue)]

```

kyle-rader (commented):
1 last comment on seconds vs minutes for the CLI option.

src/AzureAuth/CommandMain.cs
Haard30: @kyle-rader would it be better in that case to have the option interpret the number as minutes directly?

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
```diff
@@ -467,13 +469,14 @@ public async Task HasThreeAuthFlows_Returns_Null_TokenResult_With_Errors()
...
-            var authFlowExecutor = this.Subject(new[] { authFlow1.Object, authFlow2.Object, authFlow3.Object });
+            var authFlowExecutor = this.Subject(new[] { authFlow1.Object, authFlow2.Object, authFlow3.Object }, this.timeoutManager);
             var result = await authFlowExecutor.GetTokenAsync();
             var resultList = result.ToList();
 
             // Assert
             authFlow1.VerifyAll();
-            authFlow2.VerifyAll();
+
+            // authFlow2.VerifyAll();
```
reillysiemens, line 479: Should this be commented out? 🤔

Haard30: oops! no. I missed uncommenting it after making changes - will update it.

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
```diff
@@ -218,7 +220,7 @@ public async Task HasTwoAuthFlows_Returns_Null_TokenResult()
             authFlowResultList.Add(authFlowResult2);
 
             // Act
-            var authFlowExecutor = this.Subject(new[] { authFlow1.Object, authFlow2.Object });
+            var authFlowExecutor = this.Subject(new[] { authFlow1.Object, authFlow2.Object }, this.timeoutManager);
```
kyle-rader, line 223: The goal of having the `Subject` Method is that a change to the creation of the subject like this one allows to to update that one method with the new arg, and you should not have to update all these calls to Subject.

kyle-rader (changes requested):

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
```diff
@@ -54,12 +55,13 @@ public void Setup()
 
             // Mock successful token result
             this.tokenResult = new TokenResult(new JsonWebToken(TokenResultTest.FakeToken), Guid.NewGuid());
+            this.timeoutManager = new TimeoutManager(TimeSpan.FromSeconds(60));
```
reillysiemens, line 58: None of these tests take advantage of the `ITimeoutManager` abstraction. This is a real `TimeoutManager`, which means all of these tests are capped at 60s runtime (that's good anyway, but maybe not what you intend to express).

It's not clear to me what the `ITimeoutManager` abstraction is meant to help with. Maybe we don't need it?

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
```diff
@@ -742,16 +744,62 @@ public async Task HasThreeAuthFlows_Returns_Early_With_TokenResultAndErrors()
...
+                new TimeoutException("Timeout exception"),
+            };
+
+            var errors1 = new[]
+            {
+                new Exception("Exception 1"),
+            };
+
+            var authFlowResult1 = new AuthFlowResult(null, errors1, "authFlow1");
+            var authFlowResult2 = new AuthFlowResult(null, timeoutError, "authFlow2");
```
reillysiemens, line 761: You induce a timeout error by creating and injecting a result with a `TimeoutException` in it, but isn't it the `TimeoutManager` that causes that to happen? I might have misread this, but I think you're circumventing the very thing you intend to test here. Could you achieve a similar test by creating a specialized `ITimeoutManager` class? 🤔

reillysiemens (changes requested):
A lot of my suggestions here are naming-related and not blockers, but I do have some questions that I think we should answer before proceeding.

src/MSALWrapper/TimeoutManager.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <returns>True if CLI has timedout.</returns>
+        public bool HasTimedout()
+        {
+            return this.timeout <= this.timer.Elapsed;
+        }
+
+        /// <summary>
+        /// Starts the timer.
+        /// </summary>
+        public void StartTimer()
```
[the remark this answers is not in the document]
kyle-rader: +1 here, for any direct wrapper prefer using the same interface for consistency and familiarity.

src/MSALWrapper/TimeoutManager.cs
```diff
@@ -0,0 +1,70 @@
...
+            return this.timer.Elapsed;
+        }
+
+        /// <summary>
+        /// Get number of time remaining before CLI times out.
+        /// </summary>
+        /// <returns>Remaining time for timeout.</returns>
+        public TimeSpan GetRemainingTime()
+        {
+            return this.timeout.Subtract(this.timer.Elapsed);
```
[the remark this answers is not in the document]
Haard30: Yes I will update it to use `return this.timeout - this.timer.Elapsed;` My POV while coding was as we are subtracting timespans which are objects as compared to primary data structures, `Subtract()` might offer better readability. But operator overloading does work here to use `-`

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -73,5 +84,60 @@ public async Task<IEnumerable<AuthFlowResult>> GetTokenAsync()
...
+                await Task.WhenAny(Task.Delay(this.DetermineDelayPeriod()), flowResult);
+            }
+
+            return await flowResult;
+        }
+
+        /// <summary>
+        /// Determines amount of time to wait before polling.
+        /// If global timeout is in less than default delayPeriodForPolling, we would want to wait for a period equal to
+        /// timeout rather than the delayPeriodForPolling.
```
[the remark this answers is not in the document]
Haard30: The `if (this.timeoutManager.GetElapsedTime() >= this.timeToWaitBeforeWarning)` check is for the first iteration of the loop where we have not called the `Delay()` yet. So for the why of this method - It is used to determine polling interval. While our default is 30s, we don't always want to wait for that time. Two instances where this situation will arise are -
1. Initially we are warning user after 10 seconds so we will poll for that time if we have not yet passed 10s.
2. When timer reaches the end of timeout - we want to wait for only remaining time and not entire 30s. E.g - if remaining time is 20s we will wait for that time as compared to default 30s.
I will try to improve the comment on this so the why is better understood.. thanks :)

src/AzureAuth/CommandMain.cs
```diff
@@ -163,6 +169,12 @@ public CommandMain(CommandExecuteEventData eventData, ITelemetryService telemetr
         [Option(ConfigOption, "The path to a configuration file.", CommandOptionType.SingleValue)]
         public string ConfigFilePath { get; set; }
 
+        /// <summary>
+        /// Gets or sets global Timeout.
+        /// </summary>
+        [Option(TimeoutOption, "The number of minutes the CLI is allowed to wait before timing out.\nDefault: 10 minutes.", CommandOptionType.SingleValue)]
```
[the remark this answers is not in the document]
Haard30: Also, I am not sure if it was intended - but our helper text seems inconsistent with blank lines. 
image
Do we have a guideline of when we want to leave blank lines? If not, should I make them consistent with this?

src/MSALWrapper.Test/AuthFlow/AuthFlowExecutorTest.cs
```diff
@@ -751,7 +820,36 @@ private EquivalencyAssertionOptions<AuthFlowResult> ExcludeDurationTimeSpan(Equi
...
+        private class AlwaysTimesOutAuthFlow : IAuthFlow
+        {
+            public async Task<AuthFlowResult> GetTokenAsync()
+            {
+                await Task.Delay(TimeSpan.FromMinutes(100));
+                return null;
+            }
+        }
+
+        // This auth flow is used to wait before returning error.
```
kyle-rader, line 836: This class we don't need to add since we can easily mock this behavior directly in the test. because we are not relying on this to pass any time, we have mocked the stop watch - the mock behavior can immediately return the error.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -16,18 +16,28 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
     /// </summary>
     public class AuthFlowExecutor
     {
+        /// <summary>
+        /// The amount of time to wait before we start warning on stderr about waiting for auth.
+        /// </summary>
+        public static TimeSpan WarningInterval = TimeSpan.FromSeconds(10);
```
kyle-rader, line 22: I would not call this an `interval` because interval implies repetition. Consider `WarningDelay`? We only wait for this one time, before using the actual 30 sec interval below.

src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
```diff
@@ -73,5 +88,61 @@ public async Task<IEnumerable<AuthFlowResult>> GetTokenAsync()
...
+            var flowResult = Task.Run(() => authFlow.GetTokenAsync());
+            var authFlowName = authFlow.GetType().Name;
+
+            while (!flowResult.IsCompleted)
+            {
+                if (this.stopwatch.Elapsed() >= WarningInterval)
+                {
+                    var warningMessgae = $"Waiting for {authFlowName} authentication. Look for an auth prompt. " +
+                        $"\nTimeout in {this.stopwatch.Remaining():mm}m {this.stopwatch.Remaining():ss}s!";
+                    this.logger.LogWarning(warningMessgae);
```
kyle-rader, line 109: I would not worry about string concat here.

kyle-rader, line 109:
```suggestion
                    this.logger.LogWarning($"Waiting for {authFlowName} authentication. Look for an auth prompt.");
                    this.logger.LogWarning(Timeout in {this.stopwatch.Remaining():mm}m {this.stopwatch.Remaining():ss}s!");

```

kyle-rader (changes requested):
Looking pretty good, just a few remaining comments.

kyle-rader (dismissed):
Great work iterating on this!

mvanchaa (approved):

goagain (approved):

goagain:
Fantastic

reillysiemens (changes requested):
@Haard30: This is great work. My only remaining suggestion is that you add a `CHANGELOG.md` entry describing what's being added here.

reillysiemens (approved):
🚀🚢❗

melindajohnson (approved):

PR #128, opened 2022-07-18
Provide a way to disable user based authentication on non-user based environments

mvanchaa:
This PR contains changes to disable [Public Client Authentication](https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-client-applications) on environments where it's not applicable. This would help in disabling prompts on a machine which is not monitored by a user.


src/AzureAuth/CommandMain.cs
```diff
@@ -302,9 +302,34 @@ public int OnExecute()
             // Small bug in Lasso - Add does not accept a null IEnumerable here.
             this.eventData.Add("settings_scopes", this.authSettings.Scopes ?? new List<string>());
 
+            if (!this.IsAUserInteractiveEnv())
+            {
+                this.logger.LogDebug($"Skipping used based authentication as specified by the env variable(s) {EnvVars.DisableUserBasedAuthentication}/{EnvVars.CorextNonInteractive}.");
+                return 0;
```
reillysiemens, line 308: Is `0` the right thing to return? Does skipping interactivity mean `azureauth` is a no-op? 🤔 I thought that the idea was to try any modes that could proceed without interactivity but skip the rest.

reillysiemens (changes requested):
I've got some small suggestions here before merging. My biggest concern is around `Corext_Noninteractive` and whether we should respect that. If you removed that we could always add it back in later.

src/AzureAuth/CommandMain.cs
mvanchaa: The original idea was to try silent auth modes if interactivity is disabled. But the objective of the feature changed to disable any authentication that runs on behalf of a user on non-interactive machines as it's not applicable. Since non-interactive machines do not need to fetch the token silently from the cache, we won't be trying that scenario or any other silent modes.

src/AzureAuth/CommandMain.cs
```diff
@@ -302,9 +302,34 @@ public int OnExecute()
...
+            }
+
             return this.ClearCache ? this.ClearLocalCache() : this.GetToken();
         }
 
+        /// <summary>
+        /// Determines whether the given env is user interactive or not.
+        /// </summary>
+        /// <returns>A boolean to indicate user interactive env</returns>
+        public bool IsAUserInteractiveEnv()
```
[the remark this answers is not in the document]
kyle-rader: +1

src/AzureAuth/EnvVars.cs
```diff
@@ -23,6 +23,16 @@ internal static class EnvVars
         /// </summary>
         public static readonly string Cache = $"{EnvVarPrefix}_CACHE";
 
+        /// <summary>
+        /// Name of the env var used to disable user based authentication modes.
+        /// </summary>
+        public static readonly string DisableUserBasedAuthentication = $"{EnvVarPrefix}_DISABLE_USER_AUTH";
```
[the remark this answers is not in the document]
kyle-rader: Technically "No Interaction" still allows for IWA or Broker in some scenarios. But what we really want is "there is no user - so don't try authing the user at all"

kyle-rader (changes requested):
A few questions to address:
1. How are we going to track this in telemetry? We should be able to see when the CLI early exits because this was set. I think including the value of the env var or simply the bool of the computed "IsInteractive" should be included on the main command event. This will let us find in-correct / unexpected uses and also filter them out from other queries so they don't skew stats.
2. CHANGELOG update.
3. Discussed offline - for this initial feature add, leave off the corext var as we are still discussing that.

src/AzureAuth/EnvVars.cs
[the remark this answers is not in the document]
mvanchaa: That was the original idea to just disable interactivity. Since the non-interactive machines do not need any kind of authentication, specifically user based authentication, we disable user based auth modes which is all modes right now. When we introduce confidential client flows, the non-interactive machines can then use those modes.

src/AzureAuth/CommandMain.cs
mvanchaa: Ignore my previous comment. I just understood what you meant.

mvanchaa: I agree with your question on returning `0` as the error code. I will return `1` to fail it out loud.

src/AzureAuth/EnvVars.cs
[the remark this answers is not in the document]
mvanchaa: Made changes to use env variable `DISABLE_INTERACTIVITY`.

src/AzureAuth/EnvVars.cs
```diff
@@ -23,6 +23,16 @@ internal static class EnvVars
...
 
+        /// <summary>
+        /// Name of the env var used to disable user based authentication modes.
+        /// </summary>
+        public static readonly string DisableUserBasedAuthentication = $"{EnvVarPrefix}_DISABLE_USER_AUTH";
+
+        /// <summary>
+        /// Name of the env var used to disable user based authentication modes. This serves the same purpose as the env variable `DisableUserAuthModes`.
+        /// </summary>
+        public static readonly string CorextNonInteractive = $"Corext_NonInteractive";
```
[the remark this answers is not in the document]
mvanchaa: As discussed offline, removed the usage of this env var.

src/AzureAuth/CommandMain.cs
```diff
@@ -347,9 +347,31 @@ public int OnExecute()
             // Small bug in Lasso - Add does not accept a null IEnumerable here.
             this.eventData.Add("settings_scopes", this.authSettings.Scopes ?? new List<string>());
 
+            if (this.InteractivityDisabled())
+            {
+                this.logger.LogCritical($"Skipping used based authentication as specified by the env variable : {EnvVars.DisableInteractivity}.");
```
shalinikhare27, line 352:
```suggestion
                this.logger.LogCritical($"Skipping user based authentication as specified by the environment variable : {EnvVars.DisableInteractivity}.");

```

mvanchaa: oops!

src/AzureAuth/CommandMain.cs
```diff
@@ -347,9 +347,34 @@ public int OnExecute()
             // Small bug in Lasso - Add does not accept a null IEnumerable here.
             this.eventData.Add("settings_scopes", this.authSettings.Scopes ?? new List<string>());
 
+            if (this.InteractivityDisabled())
+            {
+                this.logger.LogCritical($"Skipping user based authentication as specified by the environment variable : {EnvVars.DisableInteractivity}.");
```
reillysiemens, line 352: Strictly speaking this might not be true if the alternative environment variable is being respected. You could just not display any environment variable names and write
```suggestion
                this.logger.LogCritical("Interactive authentication is disabled.");

```

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -644,6 +644,63 @@ public void TestGenerateEvent_From_AuthFlowResult_With_TokenResult_And_Null_Erro
             eventData.Measures.Should().ContainKey("duration_milliseconds");
         }
 
+        /// <summary>
+        /// Test to validate user interactive env.
+        /// </summary>
+        [Test]
+        public void InteractivityIsNotDisabledIfCorextIsSetToOne()
```
kyle-rader, line 651: This test name does not match the assertion being made in the test.

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -644,6 +644,63 @@ public void TestGenerateEvent_From_AuthFlowResult_With_TokenResult_And_Null_Erro
             eventData.Measures.Should().ContainKey("duration_milliseconds");
         }
 
+        /// <summary>
+        /// Test to validate user interactive env.
+        /// </summary>
```
kyle-rader, line 649: These XML Summaries should not be required by StyleCop - so unless they describe extra context the test name can't give, I would prefer leaving them off.

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -644,6 +644,63 @@ public void TestGenerateEvent_From_AuthFlowResult_With_TokenResult_And_Null_Erro
...
+            CommandMain subject = this.serviceProvider.GetService<CommandMain>();
+            this.envMock.Setup(e => e.Get("Corext_NonInteractive")).Returns("1");
+            subject.InteractivityDisabled().Should().BeTrue();
+        }
+
+        /// <summary>
+        /// Test to validate user interactive env.
+        /// </summary>
+        [Test]
+        public void InteractivityIsNotDisabledIfCorextIsNotSetToOneOrTrue()
```
kyle-rader, line 662:
```suggestion
        public void InteractivityIsEnabledIfCorextIsNotSetToOneOrTrue()

```

kyle-rader: Prefer in the naming to only use "IsEanbled" or "IsDisabled" to make it easier to read.

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -644,6 +644,63 @@ public void TestGenerateEvent_From_AuthFlowResult_With_TokenResult_And_Null_Erro
...
+        [Test]
+        public void InteractivityIsDisabledIfEnvVarIsSetToNonEmpty()
+        {
+            CommandMain subject = this.serviceProvider.GetService<CommandMain>();
+            this.envMock.Setup(e => e.Get("AZUREAUTH_DISABLE_INTERACTIVITY")).Returns("1");
+            subject.InteractivityDisabled().Should().BeTrue();
+
+            this.envMock.Setup(e => e.Get("AZUREAUTH_DISABLE_INTERACTIVITY")).Returns("non-empty-string");
+            subject.InteractivityDisabled().Should().BeTrue();
+        }
```
kyle-rader, line 681: This Test and the previous are of the same exact structure - and you are testing input X => output Y. For sets of X and Y. This is a good opportunity for writing a single data driven test using `TestCase`

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -644,6 +644,63 @@ public void TestGenerateEvent_From_AuthFlowResult_With_TokenResult_And_Null_Erro
...
+        /// </summary>
+        [Test]
+        public void InteractivityIsDisabledIfEnvVarIsSetToNonEmpty()
+        {
+            CommandMain subject = this.serviceProvider.GetService<CommandMain>();
+            this.envMock.Setup(e => e.Get("AZUREAUTH_DISABLE_INTERACTIVITY")).Returns("1");
+            subject.InteractivityDisabled().Should().BeTrue();
+
+            this.envMock.Setup(e => e.Get("AZUREAUTH_DISABLE_INTERACTIVITY")).Returns("non-empty-string");
+            subject.InteractivityDisabled().Should().BeTrue();
```
kyle-rader, line 680: Within a given test - it's best practice not to re-use the subject more than once, in order to test the behavior under isolation. This could be done with a `TestCase` based test.

src/AzureAuth/CommandMain.cs
```diff
@@ -347,9 +347,35 @@ public int OnExecute()
...
+            var disableInteractivity = this.env.Get(EnvVars.DisableInteractivity);
+            var corextNonInteractive = this.env.Get(EnvVars.CorextNonInteractive);
+
+            if (!string.IsNullOrEmpty(disableInteractivity) ||
+                string.Equals("1", corextNonInteractive))
+            {
+                return true;
+            }
+
+            return false;
```
kyle-rader, line 376: Generally when you have an `if` conditional that returns `true` or `false` you can simplify to return the boolean expression instead of using an `if` with multiple `returns`.

src/AzureAuth/CommandMain.cs
```diff
@@ -347,9 +347,35 @@ public int OnExecute()
...
+            }
+
             return this.ClearCache ? this.ClearLocalCache() : this.GetToken();
         }
 
+        /// <summary>
+        /// Determines whether the given env is user interactive or not.
+        /// </summary>
+        /// <returns>A boolean to indicate user interactive env</returns>
+        public bool InteractivityDisabled()
```
kyle-rader, line 365:
```suggestion
        public bool PCADisabled()
```
I would use the langauge PCA internally since that's what we really mean. Offline conversation about naming here, switching to "AZUERAUTH_NO_USER"

src/AzureAuth/EnvVars.cs
```diff
@@ -23,6 +23,16 @@ internal static class EnvVars
         /// </summary>
         public static readonly string Cache = $"{EnvVarPrefix}_CACHE";
 
+        /// <summary>
+        /// Name of the env var used to disable user based authentication modes.
+        /// </summary>
+        public static readonly string DisableInteractivity = $"{EnvVarPrefix}_DISABLE_INTERACTIVITY";
```
kyle-rader, line 29:
```suggestion
        public static readonly string DisableInteractivity = $"{EnvVarPrefix}_NO_USER";

```

kyle-rader (changes requested):

src/AzureAuth.Test/CommandMainTest.cs
mvanchaa: I was getting an error so included them. Suppressed the rule for test projects.

reillysiemens (approved):
This looks great to me. Thanks for addressing all the feedback, @mvanchaa! 🎉

shalinikhare27 (dismissed):

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -644,6 +644,37 @@ public void TestGenerateEvent_From_AuthFlowResult_With_TokenResult_And_Null_Erro
             eventData.Measures.Should().ContainKey("duration_milliseconds");
         }
 
+        [Test]
```
kyle-rader, line 647: The `[Test]` is not needed when using `TestCase`

kyle-rader: (Not a blocker)

kyle-rader (approved):

Haard30 (approved):

reillysiemens (approved):

PR #137, opened 2022-08-12
Upgrade CompleteWithin to forcefully timeout

kyle-rader:
Sometimes, the broker flow can hang in a way that doesn't respect cancellation tokens. I've reproed this for 1 day, but then not the next, likely due to machine/OS state.

The basic effect was that AcquireTokenSilent was hung, and the broker wasn't responding, but we didn't timeout that task and fall back to web.

This update uses the same technic we use in the global timeout to race a Task.Delay against the actual task, and manually kill that main task if it does not win.

I've tested this by injecting a wait for 7 seconds in the broker silent token call (not added to the PR), and confirmed it falls back to web from there.

### DIFF
--- a/src/MSALWrapper/AuthFlow/Broker.cs
+++ b/src/MSALWrapper/AuthFlow/Broker.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
         /// <summary>
         /// The silent auth timeout.
         /// </summary>
-        private TimeSpan silentAuthTimeout = TimeSpan.FromSeconds(6);
+        private TimeSpan silentAuthTimeout = TimeSpan.FromSeconds(20);
 
         /// <summary>
         /// The interactive auth timeout.
--- a/src/MSALWrapper/AuthFlow/Broker.cs
+++ b/src/MSALWrapper/AuthFlow/Broker.cs
@@ -9,6 +9,7 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
     using System;
     using System.Collections.Generic;
     using System.Threading.Tasks;
+
     using Microsoft.Extensions.Logging;
     using Microsoft.Identity.Client;
 
@@ -29,7 +30,7 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
         /// <summary>
         /// The silent auth timeout.
         /// </summary>
-        private TimeSpan silentAuthTimeout = TimeSpan.FromMinutes(5);
+        private TimeSpan silentAuthTimeout = TimeSpan.FromSeconds(6);
 
         /// <summary>
         /// The interactive auth timeout.
--- a/src/MSALWrapper/AuthFlow/DeviceCode.cs
+++ b/src/MSALWrapper/AuthFlow/DeviceCode.cs
@@ -31,7 +31,7 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
         /// <summary>
         /// The silent auth timeout.
         /// </summary>
-        private TimeSpan silentAuthTimeout = TimeSpan.FromMinutes(5);
+        private TimeSpan silentAuthTimeout = TimeSpan.FromSeconds(15);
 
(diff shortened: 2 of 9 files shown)


kyle-rader:
I will add a changelog entry for this as a bugfix.

src/MSALWrapper/TaskExecutor.cs
```diff
@@ -31,7 +31,17 @@ internal static async Task<T> CompleteWithin<T>(ILogger logger, TimeSpan timeout
             source.CancelAfter(timeout);
             try
             {
-                return await getTask(source.Token).ConfigureAwait(false);
+                Task<T> mainTask = getTask(source.Token);
+                Task wrapperTask = Task.Run(() => mainTask);
```
Haard30, line 35: Can we test if this works without needing ` wrapperTask `

```
                Task wrapperTask = Task.Run(() => mainTask);
                await Task.WhenAny(Task.Delay(timeout), mainTask);
```

kyle-rader: I had to had `wrapperTask` because the `Task.WhenAny` takes an enumerable of `Task<T>` which != `Task`. 
`mainTask` is our `Task<T>` that we want to return after awaiting while the `wrapperTask` gives us a non-generic `Task` to race the delay against.

Haard30: Got it!

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -29,7 +30,7 @@ public class Broker : IAuthFlow
         /// <summary>
         /// The silent auth timeout.
         /// </summary>
-        private TimeSpan silentAuthTimeout = TimeSpan.FromMinutes(5);
+        private TimeSpan silentAuthTimeout = TimeSpan.FromSeconds(6);
```
kyle-rader, line 33: Looking up telemetry to see what a good value for this is.

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -29,7 +30,7 @@ public class Broker : IAuthFlow
         /// <summary>
         /// The silent auth timeout.
         /// </summary>
-        private TimeSpan silentAuthTimeout = TimeSpan.FromMinutes(5);
+        private TimeSpan silentAuthTimeout = TimeSpan.FromSeconds(20);
```
Haard30, line 33: Should we also extend our initial `WarningDelay` in `AuthFlowExecutor` to `20 seconds` or more so that we provide enough time for silent auth to timeout before we warn for auth prompt

Haard30 (dismissed):

src/MSALWrapper/AuthFlow/DeviceCode.cs
```diff
@@ -31,7 +31,7 @@ public class DeviceCode : IAuthFlow
         /// <summary>
         /// The silent auth timeout.
         /// </summary>
-        private TimeSpan silentAuthTimeout = TimeSpan.FromMinutes(5);
+        private TimeSpan silentAuthTimeout = TimeSpan.FromSeconds(15);
```
mvanchaa, line 34: Why do we have different timeout for each flow?

Haard30: AFAIK, Kyle might have used telemetry data for each silent `authflow` and used the numbers accordingly

reillysiemens (dismissed):
TIL that `Task` and `Task<T>` are different, which gives me all sorts of complex feels. I'd love to find a better solution in the future, but for now this appears to work, and it just is what it is. 🤷‍♂️

src/MSALWrapper/AuthFlow/Broker.cs
reillysiemens: That seems reasonable, but we could probably do it in a subsequent PR.

reillysiemens (approved):

melindajohnson (approved):

PR #138, opened 2022-08-12
Launch broker auth prompts in the foreground

melindajohnson:
**Current State**

With the broker auth prompt hanging in the background, it gives a false impression to the user that the console app is hung, and there is no obvious way of finding the WAM dialog box. Using Alt + Tab doesn't show the WAM dialog box and this also gives an impression that there was no auth prompt. If a window handle is not given, MSAL currently uses the center of the desktop, which can cause this problem.

**Scenario to support**

 If the broker auth prompt is launched to the foreground of the calling process, users can interact with it immediately and sign in instead of being unaware that an auth prompt is waiting for their interaction.
MSAL has also made the WithParentActivityOrWindow() API [mandatory in their latest version.](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/wam#parent-window-handles)

**Solution**

Get the parent window handle of the process calling auth and pass that into MSAL.
Note : MSAL will be providing a helper for this in their library in a newer version - AzureAD/microsoft-authentication-library-for-dotnet#3590.

**Testing**
Manual validation in Windows has been done to make sure the broker auth prompt appears in front of the calling process.

### DIFF
--- a/src/MSALWrapper/AuthFlow/Broker.cs
+++ b/src/MSALWrapper/AuthFlow/Broker.cs
@@ -161,7 +161,7 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
                 {
                     HeaderText = this.promptHint,
                 })
-                .WithParentActivityOrWindow(() => this.consoleWindowHandleProvider); 
+                .WithParentActivityOrWindow(() => this.consoleWindowHandleProvider);
 
 #if NETFRAMEWORK
             clientBuilder.WithWindowsBroker();
--- a/src/MSALWrapper/AuthFlow/Broker.cs
+++ b/src/MSALWrapper/AuthFlow/Broker.cs
@@ -25,28 +25,6 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
         private readonly IList<Exception> errors;
         private IPCAWrapper pcaWrapper;
 
-        private enum GetAncestorFlags
-        {
-            /// <summary>
-            /// Retrieves the parent window. This does not include the owner, as it does with the GetParent function.
-            /// </summary>
-            GetParent = 1,
-
-            /// <summary>
-            /// Retrieves the root window by walking the chain of parent windows.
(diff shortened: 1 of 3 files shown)


reillysiemens:
It seems to me that AzureAD/microsoft-authentication-library-for-dotnet#3590 is indicating that they'll add code very similar to what you've written here so that we won't need to do this in the future, is that right?

If so, I think it would be good to include a comment in the code linking to that issue, indicating that we might be able to simplify this at a later point.

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -139,6 +158,19 @@ public async Task<AuthFlowResult> GetTokenAsync()
             return new AuthFlowResult(null, this.errors, this.GetType().Name);
         }
 
+        [DllImport("kernel32.dll")]
+        private static extern IntPtr GetConsoleWindow();
+
+        [DllImport("user32.dll", ExactSpelling = true)]
+        private static extern IntPtr GetAncestor(IntPtr windowsHandle, GetAncestorFlags flags);
```
reillysiemens, line 165: In the corresponding MSAL issue there were some additional comments above this function. Should we replicate those for completeness? 🤔

melindajohnson: added the comments.

reillysiemens (dismissed):
LGTM. 🚢

reillysiemens:
Hmm... I've just approved, but it wouldn't hurt to add a `CHANGELOG.md` entry describing this fix.

shalinikhare27 (dismissed):
Glad we confirmed with the MSAL team that this implementation will take a while before it's out. Even better that you added a comment that we can simplify this further in future.

reillysiemens (approved):

shalinikhare27 (approved):

mvanchaa (approved):

PR #142, opened 2022-08-29
AzureAuth: --resource should not be required if passing scopes

goagain:
When `--scope` fields are provided, the `--resource` field can be omitted.


src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -333,6 +334,21 @@ public void TestEvaluateOptionsWithoutAliasMissingTenant()
...
+            subject.Scopes = new string[] { ".default" };
+
+            subject.EvaluateOptions().Should().BeTrue();
+        }
+
+        /// <summary>
+        /// The test to evaluate options without alias missing required options.
+        /// </summary>
+        [Test]
+        public void TestEvaluateOptionsWithOverridedScopes()
```
mvanchaa, line 351: Method name doesn't match with the implementation. I think the above method and this method names are mismatched?

Haard30, line 351: We can also include a test where both scopes and resource are provided. In that case, `scope` field will be considered and `resource` will be ignored as in the code

goagain: My brain My brain was eaten by zombies.

mvanchaa: ROFL!! XD

goagain: Sure, Added a new test case!

Haard30 (approved):
May be, if it qualifies, we can also add a `CHANGELOG.md` entry for this bug fix

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -23,6 +23,7 @@ namespace Microsoft.Authentication.AzureAuth.Test
     using NLog.Extensions.Logging;
     using NLog.Targets;
     using NUnit.Framework;
+    using static System.Formats.Asn1.AsnWriter;
```
melindajohnson, line 26: Did this get added by mistake or am I missing something? 🤔

mvanchaa:
> May be, if it qualifies, we can also add a `CHANGELOG.md` entry for this bug fix

This doesn't sound like a bug to me. It's a change in functionality, so probably not a bug?

src/AzureAuth.Test/CommandMainTest.cs
shalinikhare27: Let's remove this as it's not used anywhere.

goagain: Accidently

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -342,12 +343,46 @@ public void TestEvaluateOptionsWithoutAliasMissingRequiredOptions()
...
                 "The --client field is required.",
                 "The --tenant field is required.",
             });
         }
 
+        /// <summary>
+        /// The test to evaluate options without resource but with scopes.
+        /// </summary>
+        [Test]
+        public void TestEvaluateOptionsWithOverridedScopes()
```
shalinikhare27, line 355: Let's add tests for the 2 other combos as well:
1. Neither resource or scope is given
2. Resource is given but no scopes

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -342,12 +342,60 @@ public void TestEvaluateOptionsWithoutAliasMissingRequiredOptions()
...
+        {
+            CommandMain subject = this.serviceProvider.GetService<CommandMain>();
+            subject.Resource = null;
+            subject.Client = "e19f71ed-3b14-448d-9346-9eff9753646b";
+            subject.Tenant = "9f6227ee-3d14-473e-8bed-1281171ef8c9";
+            subject.Scopes = new string[] { ".default" };
```
kyle-rader, line 361: From a raw data perspective this works, however - this is would not work for real.

Here you would need to pass `<resource ID/URL>/.default`. We add the `/.default` by default if only passing the resource. But if passing the scope, you have to pass the full scope. 

While this doesn't impact this test functionally, it does impact it from a documentation perspective.

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -342,12 +342,60 @@ public void TestEvaluateOptionsWithoutAliasMissingRequiredOptions()
...
+        /// The test to evaluate options with both resource and scopes.
+        /// </summary>
+        [Test]
+        public void TestEvaluateOptionsWithResourceAndScopes()
+        {
+            CommandMain subject = this.serviceProvider.GetService<CommandMain>();
+            subject.Resource = "f0e8d801-3a50-48fd-b2da-6476d6e832a2";
+            subject.Client = "e19f71ed-3b14-448d-9346-9eff9753646b";
+            subject.Tenant = "9f6227ee-3d14-473e-8bed-1281171ef8c9";
+            subject.Scopes = new string[] { ".default" };
```
kyle-rader, line 390: Same here, let's provide a valid scope example.

src/AzureAuth/CommandMain.cs
```diff
@@ -382,12 +382,18 @@ public bool PCADisabled()
         private bool ValidateOptions()
         {
             bool validOptions = true;
-            if (string.IsNullOrEmpty(this.authSettings.Resource))
+
+            if (string.IsNullOrEmpty(this.authSettings.Resource) && (this.authSettings.Scopes == null || this.authSettings.Scopes.Count == 0))
```
kyle-rader, line 386: Can we use the `?` operator here like we do in the conditional below?

src/AzureAuth/CommandMain.cs
```diff
@@ -382,12 +382,18 @@ public bool PCADisabled()
...
             {
-                this.logger.LogError($"The {ResourceOption} field is required.");
+                this.logger.LogError($"The {ResourceOption} field or the {ScopeOption} field is required.");
                 validOptions = false;
             }
 
+            if (!string.IsNullOrEmpty(this.authSettings.Resource) && this.authSettings.Scopes?.Count > 0)
+            {
+                this.logger.LogWarning($"The {ScopeOption} field is provided, so the {ResourceOption} field will be ignored.");
```
kyle-rader, line 394: Consider
```suggestion
                this.logger.LogWarning($"The {ScopeOption} option was provided with the {ResourceOption} option. Only {ScopeOption} will be used.");

```

kyle-rader (changes requested):

src/AzureAuth/CommandMain.cs
goagain: I tried the `?` operator, unfortunately, the answer is `NO`.

For example, when I use 
```csharp
this.authSettings.Scopes?.Count
```
when `this.authSettings.Scopes` is null, this expression `this.authSettings.Scopes?.Count` will be also null value rather than 0.

goagain: That is to say, if the list is null,
```csharp
this.authSettings.Scopes?.Count == 0
```
will always be false. That's very counterintuitive.

src/AzureAuth.Test/CommandMainTest.cs
kyle-rader: Did those tests not already exist?

src/AzureAuth/CommandMain.cs
kyle-rader: In this case - I would expect the line below doing 
```csharp
this.authSettings.Scopes?.Count > 0
```
to have the same potential runtime issue of comparing a `null` to `0`.

You could avoid this in both cases by extracting the value into a local variable above both statements like so:
```csharp
int scopesCount = this.authSettings.Scopes?.Count ?? 0;
```
And then you can clean up both comparisons in each conditional.

kyle-rader (changes requested):

src/AzureAuth/CommandMain.cs
goagain: Good idea. Thank you @kyle-rader.

src/AzureAuth.Test/CommandMainTest.cs
goagain: Those tests are already covered.

src/AzureAuth.Test/CommandMainTest.cs
reillysiemens: @goagain: I think what @kyle-rader was suggesting here and in the previous comment as well was that the example should contain something more like `f0e8d801-3a50-48fd-b2da-6476d6e832a2/.default`.

reillysiemens: Or to provide something that isn't `f0e8d801-3a50-48fd-b2da-6476d6e832a2/.default` at all, but rather another valid Azure scope.

reillysiemens (commented):
@goagain: These changes look good to me save for Kyle's comments about the scopes. I think this PR will also need to be updated after the latest changes in `main`.

src/AzureAuth.Test/CommandMainTest.cs
goagain: I just found `scopes` in all existing unit test were `.default`. Do I need to change all of them to `f0e8d801-3a50-48fd-b2da-6476d6e832a2/.default`?

reillysiemens:
@goagain: I want to approve this PR, but because you have not recently merged `main` (I usually rebase in cases like this, but that's a personal preference) the `CHANGELOG.md` is not up-to-date. This change should be noted **after** the `0.5.1` release now.

goagain:
> @goagain: I want to approve this PR, but because you have not recently merged `main` (I usually rebase in cases like this, but that's a personal preference) the `CHANGELOG.md` is not up-to-date. This change should be noted **after** the `0.5.1` release now.

Thank you @reillysiemens, I moved it to unreleased section.

reillysiemens (approved):

kyle-rader (approved):

PR #143, opened 2022-09-02
Increase the initial warning delay to 20 seconds

Haard30:
With recent changes, we have increased the timeout period for `Broker` silent `authflow` to `20 seconds`. However, we first warn the user to look for an auth prompt after `10 seconds`. So there is rough edge case where the silent mode might still be running and we warn the user to look for an prompt. 
As a part of this PR, I am increasing the initial `WarningDelay` in `AuthFlowExecutor` to `20 seconds` so that we provide enough time for silent auth to timeout before we warn for an auth prompt.

### DIFF
--- a/src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
+++ b/src/MSALWrapper/AuthFlow/AuthFlowExecutor.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
         /// <summary>
         /// The amount of time to wait before we start warning on stderr about waiting for auth.
         /// </summary>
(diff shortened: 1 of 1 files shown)


reillysiemens (approved):

melindajohnson (approved):

goagain (approved):
Is there a data as evidence?

Haard30:
> Is there a data as evidence?

Rick were you talking about the timeouts? Yes - the timeouts of `15 seconds`, `20 seconds` are from telemetry data. The initial `WarningDelay` of `20 seconds` is just based on the max silent auth timeout of `20 seconds`.

PR #145, opened 2022-09-08
Send auth flow events before early exiting when no success events exist

kyle-rader:
We found a bug where `azureauth_command_azureauth` events have no corresponding `authflow` events because the overall invocation has no successes, and we early exist before sending the individual events that contain the valuable `error_messages`.


goagain (dismissed):

shalinikhare27 (dismissed):

reillysiemens:
@kyle-rader: Did you want to do some extra validation/address linting issues before merging this?

Also, is that a lint we want to keep around? I think it's just mad about the extra newline you have. If that's the case, personally I'd be fine with turning that off. I often add extra newlines to separate logical blocks for clarity.

Haard30 (approved):

reillysiemens (approved):

mvanchaa (approved):

shalinikhare27 (approved):

PR #150, opened 2022-09-21
Use GitHub version Lasso as a submodule

goagain:
Use GitHub version Lasso as a submodule
https://github.com/microsoft/Lasso


goagain:
close #150 
We no longer need OSS Lasso.

PR #151, opened 2022-09-28
User/mstannah/iwa update

melindajohnson:
Refactoring IWA AuthFlow to call IWA after GetTokenSilent.
Previously GetTokenIntegratedWindowsAuthenticationAsync would only get called if there were no accounts found in cache.
But we want it to be called immediately if GetTokenSilent call fails.


kyle-rader:
@melindajohnson - Let's update the change log.

Haard30 (approved):

PR #153, opened 2022-09-28
Iwa timeoutfix and newrelease

melindajohnson:
While testing the new release 0.5.2 we uncovered that IWA calls might need more than 6 seconds to succeed at least for the first time.
We also saw s MSAL Client exception "WS-Trust endpoint not found in metadata document." which means you are not on Corp net and need to turn on your VPN. Adding warning logs in this PR to reflect this error.

This PR also has updates to the Changelog and README files for the new release 0.5.3 along with the fixes.

### DIFF
--- a/src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
+++ b/src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
@@ -109,6 +109,10 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
             {
                 this.logger.LogWarning($"Msal Client Exception! (Not expected)\n{ex.Message}");
                 this.errors.Add(ex);
+                if (ex.Message.StartsWith("WS-Trust endpoint not found", StringComparison.OrdinalIgnoreCase))
+                {
+                    this.logger.LogWarning($"IWA only works on Corp Net, please turn on VPN.");
+                }
             }
             catch (NullReferenceException ex)
             {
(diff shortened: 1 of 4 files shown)


shalinikhare27 (approved):

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -109,6 +109,10 @@ public async Task<AuthFlowResult> GetTokenAsync()
             {
                 this.logger.LogWarning($"Msal Client Exception! (Not expected)\n{ex.Message}");
                 this.errors.Add(ex);
+                if (ex.Message.StartsWith("WS-Trust endpoint not found", StringComparison.OrdinalIgnoreCase))
```
kyle-rader, line 112: Does the message really start with this or is there other content in there? Have you debugged this to see the actual message while your VPN is off?

Haard30: Here's the terminal o/p when VPN is off
```
PS C:\> azureauth --mode=iwa --alias ado --config c:\data\azureauth.toml
Msal Client Exception! (Not expected)
WS-Trust endpoint not found in metadata document.
```

mvanchaa: Since we are expecting the exact message, I think we don't need to ignore case. :)

melindajohnson: Yes, Checked with MSALs error message. But to be on the safe side, if MSAL were to update it with a prefix, I am updating this to check if it contains the same content.

Haard30 (approved):

mvanchaa (approved):

PR #154, opened 2022-09-29
Upgrade to net6.0 on Windows

kyle-rader:
# Problem
net5.0 has now passed end of life. Our Mac and Linux builds are already on net6.0 but the Windows build has remained on net5.0-windows10.xxx to maintain current WAM support. 

## Problem 2
MSAL.NET has not yet added a net6-windows target and it will check this target _at runtime_ and fail if you aren't using that version.

# Workaround
We can target a new windows specific TFM (Target Framework Moniker) to build our Windows version of AzureAuth with net6 but manually use use the net5.0-windows10.xxx version of the MSAL.NET dll (`Microsoft.Identity.Client.dll`).

This works by updating our `PackageReference` to generate the special `PkgPACKAGE_NAME` varaibles (which we use heavily in Office) and then create a separate reference tag for MSAL where we point explicitly to the DLL in the package we want to use for that namespace.

## Trimming and Testing
To help fully test I added publish scripts for both mac and windows (these work on both arm and intel macs + windows).

This is helpful because they build in release mode, publish, and trim - just like our CI. So the output in `./dist` is the actual binary, (unsigned) we would be shipping. This helped uncover new trimming errors, and lead to adding 2 AssemblyRoots to ensure those Dlls are not trimmed away. This is only needed for the Windows build. 

# Testing
- [x] Windows: Build and Test
  - [x] Clear & Auth
     - [x] IWA
     - [x] Broker 
     - [x] Web
  - [x] Auth silently with AT refresh (Web mode)
  - [x] Both on and Off VPN (IWA fails off VPN as expected)
- [x] Mac: Build and Test (Both ARM `arm64` and Intel `x86_64)
   - [x] Tests
   - [x] refresh auth with web mode silently
   - [x] clear
   - [x] auth in web mode
   - [x] auth in device code mode


src/AzureAuth.Test/AzureAuth.Test.csproj
```diff
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup Condition="$([MSBuild]::IsOSPlatform('Windows'))">
-    <TargetFramework>net5.0-windows10.0.17763.0</TargetFramework>
-    <SupportedOSPlatformVersion>7</SupportedOSPlatformVersion>
+    <TargetFramework>net6.0-windows</TargetFramework>
```
bgavrilMS, line 4: TLDR; use `net6.0-windows10.0.19041.0` - see https://learn.microsoft.com/en-us/dotnet/standard/frameworks#os-version-in-tfms 


There's a difference between "netX-windows" and "netX-windows10.0.Y.Z". The first allows a windows app to use WinForms. The second also allows to call into Windows COM APIs, which we need for our older WAM interop.

kyle-rader: That was the magic version! Figuring that out from the docs would have taken me a lot longer! Thanks @bgavrilMS !

src/MSALWrapper/MSALWrapper.csproj
```diff
@@ -38,13 +38,24 @@
...
+
+  <ItemGroup Condition="$([MSBuild]::IsOSPlatform('Windows')) == false">
+    <!-- Not Windows: simple package reference of 4.47 -->
+    <PackageReference Include="Microsoft.Identity.Client" Version="4.47.1" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net472' ">
-    <PackageReference Include="Microsoft.Identity.Client.Desktop" Version="4.43.1" />
+  <ItemGroup Condition="$([MSBuild]::IsOSPlatform('Windows'))">
+    <!-- Windows: Explicitly continue using the net5 version of MSAL -->
```
reillysiemens, line 51: Will you link to https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3682 here, @kyle-rader?

goagain:
Seems like changing csproj file doesn't trigger dotnet tests. Should we add it?

kyle-rader:
IWA and Broker still work, but Web mode is now broken. @bgavrilMS is there a way to explicitly do system web browser and not embedded web browser?

kyle-rader:
> Seems like changing csproj file doesn't trigger dotnet tests. Should we add it?

@goagain we do already have `src/**` in the trigger paths, it may not have run since I am in Draft mode?

pmaytak:
> IWA and Broker still work, but Web mode is now broken. @bgavrilMS is there a way to explicitly do system web browser and not embedded web browser?
> 
> EDIT: This is due to Assembly trimming. I'll have to add some trimming hints.

@kyle-rader To use OS browser add `WithUseEmbeddedWebView(false)` to AcquireTokenInteractive call.

More info [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/MSAL.NET-uses-web-browser) and [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/System-Browser-on-.Net-Core#system-browser-experience).

src/AzureAuth.Test/AzureAuth.Test.csproj
pmaytak: Yep, the precedence rules and availability of different components per target is confusing. 
Aside: With our [new WAM implementation](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/WAM), you can just target plain net6.

reillysiemens (dismissed):

bin/mac_publish.sh
```diff
@@ -0,0 +1,42 @@
...
+        ;;
+    *)
+        error "Unsupported OS '${os_name}'"
+        exit 1
+        ;;
+esac
+#------------------------------------------
+
+# Publish
+dotnet publish "$GIT_DIR/src/AzureAuth/AzureAuth.csproj" --self-contained true -r "$runtime" -c release -o $DIST $*
```
goagain, line 42:
```suggestion
dotnet publish "$GIT_DIR/src/AzureAuth/AzureAuth.csproj" --self-contained true -r "$runtime" -c release -o $DIST $*

```
A new line of end may be helpful.

bin/win_publish.cmd
```diff
@@ -0,0 +1,4 @@
+@ECHO OFF
+
+rmdir /S /Q %~dp0\dist
+dotnet publish %~dp0\..\src\AzureAuth\AzureAuth.csproj --self-contained true -r win10-x64 -c release -o %~dp0\..\dist %*
```
goagain, line 4:
```suggestion
dotnet publish %~dp0\..\src\AzureAuth\AzureAuth.csproj --self-contained true -r win10-x64 -c release -o %~dp0\..\dist %*

```
same

goagain (approved):
Tested all modes on Windows and Mac.

Haard30 (approved):
Tested all modes on Windows

PR #155, opened 2022-09-29
User/mstannah/ nouser should not disable IWA authmode

melindajohnson:
When users set env variables to disable user interaction for auth in their automated processes, we currently disable all auth modes.
Instead, we would still need to run just IWA even if that env variable is set, since there would be no user interaction required for IWA and automated processes will be able to auth successfully.

### DIFF
--- a/src/AzureAuth.Test/CommandMainTest.cs
+++ b/src/AzureAuth.Test/CommandMainTest.cs
@@ -725,29 +725,46 @@ invalid_key = ""this is not a valid alias key""
         [TestCase("non-empty-string", false)]
         [TestCase("true", false)]
         [TestCase("", false)]
-        public void TestPCA_IsDisabledOnCorextEnvVar(string corextNonInteractive, bool expected)
+        public void InteractiveAuth_IsDisabledOnCorextEnvVar(string corextNonInteractive, bool expected)
         {
             CommandMain subject = this.serviceProvider.GetService<CommandMain>();
             this.envMock.Setup(e => e.Get("Corext_NonInteractive")).Returns(corextNonInteractive);
-            subject.PCADisabled().Should().Be(expected);
+            subject.InteractiveAuthDisabled().Should().Be(expected);
         }
 
         [TestCase("1", true)]
         [TestCase("non-empty-string", true)]
         [TestCase("true", true)]
         [TestCase("", false)]
-        public void TestPCA_IsDisabledOnNoUserEnvVar(string noUser, bool expected)
+        public void InteractiveAuth_IsDisabledOnNoUserEnvVar(string noUser, bool expected)
         {
             CommandMain subject = this.serviceProvider.GetService<CommandMain>();
             this.envMock.Setup(e => e.Get("AZUREAUTH_NO_USER")).Returns(noUser);
-            subject.PCADisabled().Should().Be(expected);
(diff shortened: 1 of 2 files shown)


src/AzureAuth/CommandMain.cs
```diff
@@ -234,6 +234,20 @@ public static string PrefixedPromptHint(string promptHint)
...
 
+        /// <summary>
+        /// Get the correct CombinedAuthMode depending on env variables to disable interactive auth modes.
+        /// </summary>
+        /// <returns>AuthModes.</returns>
+        public AuthMode GetCombinedAuthMode()
+        {
+            if (this.InteractiveAuthDisabled())
+            {
+                return AuthMode.IWA;
```
kyle-rader, line 245: Referencing the actual `AuthMode.IWA` enum variant is only possible on Windows, since the enum is conditionally compiled. The IsEnabled method exists on both platforms is hardcoded to return false on non-Windows platforms.

kyle-rader: I like managing the modes in this method, but it will have to be platform specific. You could return a specific mode if windows, and no mode (0 as this enum) if Mac.

src/AzureAuth/CommandMain.cs
```diff
@@ -217,7 +217,7 @@ public Alias TokenFetcherOptions
             get { return this.authSettings; }
         }
 
-        private AuthMode CombinedAuthMode => this.AuthModes.Aggregate((a1, a2) => a1 | a2);
+        private AuthMode CombinedAuthMode => this.GetCombinedAuthMode();
```
kyle-rader, line 220: Rather than leave this method and have it call a new one, we can change the implementation of this method directly. Change the the `=>` to a `{` and pood, now you are have a normal method.

kyle-rader (changes requested):
I like this direction, just have a couple tweaks to recommend, and we need to take care of platform sematics.

src/AzureAuth/CommandMain.cs
melindajohnson: Updated to return platform specific auth mode

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -750,21 +750,34 @@ public void InteractiveAuth_IsEnabledIfEnvVarsAreNotSet()
...
+        [Platform("Win")] // Only valid on Windows
         public void GetCombinedAuthMode_withInteractiveAuthDisabled(string noUser)
         {
             CommandMain subject = this.serviceProvider.GetService<CommandMain>();
             this.envMock.Setup(e => e.Get("AZUREAUTH_NO_USER")).Returns(noUser);
-            subject.GetCombinedAuthMode().Should().Be(AuthMode.IWA);
+            subject.CombinedAuthMode.Should().Be(AuthMode.IWA);
         }
 
+        [Platform("Win")] // Only valid on Windows
```
kyle-rader, line 763: If these are in an `#if PlatformWindows` then they are not compiled on other platforms, so we should not also need the Platform win attributes.

kyle-rader (approved):
1 non-blocking suggestion.

goagain (approved):

Haard30 (approved):

kyle-rader (approved):

PR #157, opened 2022-10-05
Script to trigger (and wait) an Azure Devops release from GitHub workflow

mvanchaa:
This script will be triggered from the as a part of the [release workflow](https://github.com/AzureAD/microsoft-authentication-cli/actions/workflows/release.yml).

Functionality:
It will trigger an Azure DevOps pipeline from GitHub workflow and waits for it to be finished. 
Once the ADO release is done, it will either fail or continue the GitHub workflow depending on the ADO build status.

NOTE:
There is no timeout for wait. ADO pipelines have a timeout, but the timer starts only *after* the release has started. It doesn't count the wait time in queue - [Jobs in Azure Pipelines and TFS - Azure Pipelines | Microsoft Learn](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#timeouts). Since queue wait time is technically unknown, we will let the script wait indefinitely. But the good news is the current average queue wait time doesn't seem to exceed more than an hour. 

Testing:
Tested on my local machine and from a private GitHub repository workflow.

Pending:
Organization and project repo secrets are added. Adding the pipeline ID as a repo secret is pending and I have created a task to keep track of it. So, this will be done once the pipeline is created.


bin/release.py
```diff
@@ -0,0 +1,125 @@
...
+    }
+    return release_metadata
+
+
+def release_status_match(release, expected_status_list):
+    """Returns True if status of all the environments of the release match any of the expected statuses"""
+    # Each release can have one or more environments (stages).
+    release_env_statuses = [environment.status for environment in release.environments]
+    # return all(status in expected_status_list for status in release_env_statuses)
+    for status in release_env_statuses:
```
mvanchaa, line 62: This logic can be written in one line (commented in line 61). It seemed not easy to understand at first glance. But I'm open to suggestions.

bin/release.py
```diff
@@ -0,0 +1,124 @@
+"""A script which triggers and wait for an azure devops to complete a release"""
```
shalinikhare27, line 1:
```suggestion
"""A script which triggers and waits for Azure DevOps to complete a release"""

```

bin/release.py
```diff
@@ -0,0 +1,124 @@
...
+from msrest.authentication import BasicAuthentication
+
+FAILED_STATUSES = ["canceled", "partiallySucceeded", "rejected"]
+COMPLETED_STATUSES = ["succeeded", "canceled", "partiallySucceeded", "rejected"]
+
+
+def create_ado_connection(organization, ADO_PAT) -> Connection:
+    """Returns an ADO connection to call the ADO Rest APIs."""
+
+    organization_url = "https://dev.azure.com/{0}".format(organization)
```
goagain, line 16:
```suggestion
    organization_url = f"https://dev.azure.com/{organization}“
```

bin/release.py
goagain, line 1:
```python
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.
```
Do we need to add copyright in Python scripts?

bin/release.py
```diff
@@ -0,0 +1,125 @@
...
+def get_release_definition(project, pipeline_name, release_client):
+    """Returns the ADO definition for the given pipeline"""
+    # Get all release definitions in a given project
+    project_release_definitions = release_client.get_release_definitions(project)
+
+    # Filter release definitions with the given pipeline name
+    pipeline_release_definitions = []
+    for definition in project_release_definitions.value:
+        if definition.name == pipeline_name:
+            pipeline_release_definitions.append(definition)
```
reillysiemens, line 33: Consider a list comprehension for simple filtering like this.

```suggestion
    pipeline_release_definitions = [d for d in project_release_definitions.value if d.name == pipeline_name]
```

bin/release.py
```diff
@@ -0,0 +1,125 @@
...
+
+    if pipeline_release_definitions is None or len(pipeline_release_definitions) == 0:
```
reillysiemens, line 35: Python will allow you to handle both of these conditions in a single check by writing
```suggestion
    if not pipeline_release_definitions:
```
Both `None` and a `list` of length `0` are considered ["falsy"](https://docs.python.org/3/library/stdtypes.html#truth-value-testing).

More importantly though, it's not actually possible for `pipeline_release_definitions` to be `None`. You defined it just above. It will always be a `[]` in the worst case. Was this just a cautious C#-style check? 🤔

bin/release.py
```diff
@@ -0,0 +1,125 @@
...
+        error_message = f"Pipeline named {pipeline_name} not found in project {project}"
+        raise Exception(error_message)
```
reillysiemens, line 37: I think you can safely combine these lines. The intermediate variable `error_message` is only used immediately afterwards.
```suggestion
        raise Exception(f"Pipeline named {pipeline_name} not found in project {project}")
```

bin/release.py
```diff
@@ -0,0 +1,125 @@
...
+
+    if len(pipeline_release_definitions) > 1:
+        error_message = (
+            f"More than 1 Pipeline named {pipeline_name} found in project {project}"
+        )
+        raise Exception(error_message)
```
reillysiemens, line 43: Same here.

```suggestion
        raise Exception(
            f"More than 1 Pipeline named {pipeline_name} found in project {project}"
        )
```

bin/release.py
```diff
@@ -0,0 +1,125 @@
...
+    return pipeline_release_definitions[0]
+
+
+def populate_release_metadata(project, pipeline_name, release_client):
+    """Returns release metadata with required information"""
+    release_definition = get_release_definition(project, pipeline_name, release_client)
+    release_metadata = {
+        "definitionId": release_definition.id,
+    }
+    return release_metadata
```
reillysiemens, line 54: I _think_ if you remove the trailing comma then Black won't put this dictionary on multiple lines. You can have a readable 2-liners that way. 😆 
```suggestion
    release_definition = get_release_definition(project, pipeline_name, release_client)
    return {"definitionId": release_definition.id}
```

bin/release.py
```diff
@@ -0,0 +1,124 @@
...
+import time
+from azure.devops.connection import Connection
+from msrest.authentication import BasicAuthentication
+
+FAILED_STATUSES = ["canceled", "partiallySucceeded", "rejected"]
+COMPLETED_STATUSES = ["succeeded", "canceled", "partiallySucceeded", "rejected"]
+
+
+def create_ado_connection(organization, ADO_PAT) -> Connection:
+    """Returns an ADO connection to call the ADO Rest APIs."""
```
shalinikhare27, line 15: [nit]
```suggestion
    """Returns an ADO connection to call the ADO REST APIs."""

```

bin/release.py
```diff
@@ -0,0 +1,124 @@
...
+
+    organization_url = "https://dev.azure.com/{0}".format(organization)
+    credentials = BasicAuthentication("", ADO_PAT)
```
shalinikhare27, line 18: How long does this PAT live?

bin/release.py
```diff
@@ -0,0 +1,125 @@
...
+    return True
+
+
+def wait_for_release(release_client, project, release_id):
+    """Wait for the azure devops release to finish"""
+    release = release_client.get_release(project, release_id)
+
+    # polling interval is set in accordance with the rate limits specified here:
+    # https://learn.microsoft.com/en-us/azure/devops/integrate/concepts/rate-limits?view=azure-devops
+    polling_interval_secs = 30
```
reillysiemens, line 74: No reason we can't just use the full word here.
```suggestion
    polling_interval_seconds = 30

```

bin/release.py
```diff
@@ -0,0 +1,124 @@
...
+
+    if pipeline_release_definitions is None or len(pipeline_release_definitions) == 0:
+        error_message = f"Pipeline named {pipeline_name} not found in project {project}"
+        raise Exception(error_message)
+
+    if len(pipeline_release_definitions) > 1:
+        error_message = (
+            f"More than 1 Pipeline named {pipeline_name} found in project {project}"
+        )
+        raise Exception(error_message)
```
shalinikhare27, line 43: Shouldn't this be a separate error message than the one above? If we have more than one pipeline names, then returning an error message saying none found is confusing.

bin/release.py
```diff
@@ -0,0 +1,125 @@
...
+    """Creates and waits for an azure devops release to be finished"""
+    connection = create_ado_connection(organization, ADO_PAT)
+    release_client = connection.clients.get_release_client()
+    release_metadata = populate_release_metadata(project, pipeline_name, release_client)
+
+    triggered_release = release_client.create_release(release_metadata, project)
+    release_url = f"https://dev.azure.com/{organization}/{project}/_releaseProgress?_a=release-pipeline-progress&releaseId={triggered_release.id}"
+    print(
+        f"Successfully triggered a release. Waiting for the release to be completed.\nMore details on the release can be found here: {release_url}"
+    )
```
reillysiemens, line 95: Strings placed on multiple lines _without_ a `,` in Python will be considered the same string, so you can make this a little more readable like so.

```suggestion
    print(
        "Successfully triggered a release. Waiting for the release to be completed.\n"
        f"More details on the release can be found here: {release_url}"
    )

```

bin/release.py
```diff
@@ -0,0 +1,125 @@
...
+
+    completed_release = wait_for_release(release_client, project, triggered_release.id)
+    if release_status_match(completed_release, FAILED_STATUSES):
+        raise Exception("Azure Devops release failed!")
+
+    print("Azure Devops release succeeded!")
```
reillysiemens, line 101:
```suggestion
    print("Azure DevOps release succeeded!")

```

bin/release.py
```diff
@@ -0,0 +1,125 @@
...
+
+    triggered_release = release_client.create_release(release_metadata, project)
+    release_url = f"https://dev.azure.com/{organization}/{project}/_releaseProgress?_a=release-pipeline-progress&releaseId={triggered_release.id}"
+    print(
+        f"Successfully triggered a release. Waiting for the release to be completed.\nMore details on the release can be found here: {release_url}"
+    )
+
+    completed_release = wait_for_release(release_client, project, triggered_release.id)
+    if release_status_match(completed_release, FAILED_STATUSES):
+        raise Exception("Azure Devops release failed!")
```
reillysiemens, line 99:
```suggestion
        raise Exception("Azure DevOps release failed!")

```

reillysiemens (changes requested):
I have some suggestions here, but they're mostly superficial. Overall this is good work, @mvanchaa!

bin/release.py
```diff
@@ -0,0 +1,124 @@
...
+    while not release_status_match(release, COMPLETED_STATUSES):
+        time.sleep(polling_interval_secs)
+        release = release_client.get_release(project, release_id)
+    return release
+
+
+def create_and_wait_for_azure_devops_release(
+    organization, project, pipeline_name, ADO_PAT
+):
+    """Creates and waits for an azure devops release to be finished"""
```
shalinikhare27, line 86: The definition above this does exactly that - `wait for the azure devops release to finish`. Can we update the description to understand the difference between the two steps? :)

bin/release.py
mvanchaa: Yes Shalini! I didn't get you. I created two different error messages for both scenarios.

bin/release.py
mvanchaa: We can define its lifetime until one year.

shalinikhare27 (commented):
Added some comments.

reillysiemens:
Oh, and I would call this `bin/linux_release.py` instead of `bin/release.py`. I think that's more specific.

bin/release.py
shalinikhare27: How does that adhere to our security policies? 1 years is TOO long!

bin/release.py
reillysiemens: Isn't it a separate message? In both cases we `raise` an exception causing an early exit. This error will never be hit if the other one is, so no one will see both. We should have exactly one pipeline with this name, so `0` and `>1` are both different errors.

bin/release.py
mvanchaa: Sorry I was not clear. I meant anywhere between a day to one year. It's upto us to define the lifetime between the range.

bin/release.py
```diff
@@ -0,0 +1,125 @@
...
+    # Wait until the release have a complete status.
+    while not release_status_match(release, COMPLETED_STATUSES):
+        time.sleep(polling_interval_secs)
+        release = release_client.get_release(project, release_id)
+    return release
+
+
+def create_and_wait_for_azure_devops_release(
+    organization, project, pipeline_name, ADO_PAT
+):
```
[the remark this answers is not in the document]
mvanchaa: I capitalized it because it's in abbreviated form. But I agree.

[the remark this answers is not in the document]
reillysiemens: Oh, I see! That makes sense, but it's uncommon to see that. Non-global Python variable names are almost always lowercase, regardless of their normal English capitalization.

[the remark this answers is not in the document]
mvanchaa: ah got it!

mvanchaa:
> Oh, and I would call this `bin/publish_linux_packages.py` instead of `bin/release.py`. I think that's more specific.

kyle-rader (changes requested):

bin/trigger_azure_pipelines.py
```diff
@@ -0,0 +1,97 @@
...
+    build = build_client.get_build(project, build_id)
+
+    # polling interval is set in accordance with the rate limits specified here:
+    # https://learn.microsoft.com/en-us/azure/devops/integrate/concepts/rate-limits?view=azure-devops
+    polling_interval_seconds = 30
+
+    # Wait until the build have a complete status.
+    while build.result not in COMPLETED_STATUSES:
+        time.sleep(polling_interval_seconds)
+        build = build_client.get_build(project, build_id)
```
Haard30, line 46: Not at all a blocker but just curious - Could there be any scenario of a hung build and this script waits till the build is done?

Haard30 (approved):
Looks awesome to me.

bin/trigger_azure_pipelines.py
mvanchaa: Good question, Haard! There is a note I have added in the description about the same scenario.  
There is a timeout set for OE build pipelines. So, the build once it's triggered, ADO enforces a timeout. This make sures we don't wait forever on the build. 
However, if the build hasn't started and is waiting in the queue to get started, there is no timeout while waiting in the queue. In such case, we can manually cancel the ADO build and it will not have any impact because it hasn't started.

bin/trigger_azure_pipelines.py
```diff
@@ -0,0 +1,97 @@
...
+def create_ado_connection(organization, ado_pat) -> Connection:
+    """Returns an ADO connection to call the ADO REST APIs."""
+
+    return Connection(
+        base_url=f"https://dev.azure.com/{organization}",
+        creds=BasicAuthentication("", ado_pat),
+    )
+
+
+def build_status_match(build, expected_status_list):
```
melindajohnson, line 28: This doesn't seem to be used anywhere. Do we still need it?

mvanchaa: Thanks Melinda! Yup this is not needed.

bin/trigger_azure_pipelines.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # Wait until the build have a complete status.
+    while build.result not in COMPLETED_STATUSES:
+        time.sleep(polling_interval_seconds)
+        build = build_client.get_build(project, build_id)
+    return build
+
+
+def trigger_azure_pipeline_and_wait_until_its_completed(
+    organization, project, pipeline_id, ADO_PAT
```
reillysiemens, line 44: This `ADO_PAT` argument should be all lowercase.

bin/trigger_azure_pipelines.py
```diff
@@ -0,0 +1,90 @@
...
+    # 1. Read env vars.
+    try:
+        # ADO PAT (Azure DevOps Personal Access Token) with "Build" scope.
+        # More information here - https://learn.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate?view=azure-devops&tabs=Windows#create-a-pat
+        ado_pat = os.environ["AZURE_DEVOPS_BUILD_PAT"]
+        organization = os.environ["ADO_ORGANIZATION"]
+        project = os.environ["ADO_PROJECT"]
+        pipeline_id = os.environ["ADO_AZUREAUTH_LINUX_PIPELINE_ID"]
+
+    except KeyError as exc:
```
kyle-rader, line 75: nit: generally `ex` is used as the short form of `exception`. The `c` isn't really needed.

reillysiemens: In my Python experience both are fine and frequently seen.

reillysiemens: This code is actually borrowed from me and I'm the one who used `exc`. 😆

bin/trigger_azure_pipelines.py
```diff
@@ -0,0 +1,90 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""A script which triggers and waits for Azure DevOps to complete a build"""
+
+import os
+import sys
+import time
+from ast import parse
```
reillysiemens, line 9: I don't think you actually use this import. I would remove it.

reillysiemens (approved):

melindajohnson (approved):

bin/trigger_azure_pipelines.py
kyle-rader: Noted, same comment, not a blocker :)

kyle-rader (approved):
Thanks for all the updates here!

PR #159, opened 2022-10-21
Always set the output of AzureAuth to utf-8 encoding

kyle-rader:
Right now, if we try to output JSON or a status message containing Unicode characters, the JSON output can be malformed. This is an issue for some tools that use the authenticated user for got configuration.

### DIFF
--- a/src/AzureAuth/Program.cs
+++ b/src/AzureAuth/Program.cs
@@ -4,6 +4,8 @@
 namespace Microsoft.Authentication.AzureAuth
 {
     using System;
+    using System.Text;
+
     using McMaster.Extensions.CommandLineUtils;
     using Microsoft.Office.Lasso;
     using Microsoft.Office.Lasso.Telemetry;
@@ -15,6 +17,10 @@ namespace Microsoft.Authentication.AzureAuth
     {
         private static void Main(string[] args)
         {
+            // Use UTF-8 output encoding.
+            // This will impact the NLog Console Target as well as any other Console usage.
+            Console.OutputEncoding = Encoding.UTF8;
+
             CommandLineApplication app = new CommandLineApplication<CommandMain>();
 
             // We always instantiate and depend on telemetry services, but these defaults
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Upgrade the Windows build to use net6 now that net5 has reached end of life.
+- Set console output encoding to `utf-8` explicitly.
 
 ## [0.5.4] - 2022-09-29
 ### Fixed


goagain:
I love this PR, but it seems like too large to review. Does it come from dotnet6 branch? or should we merge it after dotnet6 branch gets merged?

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -86,8 +83,8 @@ private enum GetAncestorFlags
         public async Task<AuthFlowResult> GetTokenAsync()
         {
             IAccount account = await this.pcaWrapper.TryToGetCachedAccountAsync(this.preferredDomain)
-                ?? Identity.Client.PublicClientApplication.OperatingSystemAccount;
-            this.logger.LogDebug($"Using cached account '{account.Username}'");
+                 ?? PublicClientApplication.OperatingSystemAccount;
+            this.logger.LogDebug($"Using cached account '{account?.Username}'");
```
mvanchaa, line 87: If the account is null, would it just print "Using cached account "?

src/AzureAuth/Program.cs
```diff
@@ -15,6 +17,10 @@ public class Program
     {
         private static void Main(string[] args)
         {
+            // Use UTF-8 output encoding.
+            // This will impact the NLog Console Target as well as any other Console usage.
+            Console.OutputEncoding = Encoding.UTF8;
```
mvanchaa, line 22: Just curious, where/how is the Console being passed or used after this statement?

mvanchaa: Also, is it possible to write unit tests for this?

mvanchaa:
I'm not sure if I'm asking the right question. Does this mean the token contains Unicode characters and hence the JSON output or the status message contains those Unicode characters? If so, should MSAL also fix this?

goagain (dismissed):
Tested all modes on Windows and Mac.

src/AzureAuth/Program.cs
kyle-rader: Great questions!
1. The `Console` is a built in static class and this property we are changing dictates the encoding the `Console` uses for things like `Console.Write` and `Console.WriteLine`. This impacts everything globally that writes using the console, which for this particular bug - is the JSON output other programs are trying to read. To avoid writing tokens in logs, we only use `Console.Write` when printing the token/json. 
2. This is a change which I would argue doesn't warrant a unit test. It could have an integration test, but we have no good way of writing integration tests at this time, because to fully test this - you _must authenticate_ which we can't do in CI, again, at this time. This is not say unit testing encoding behavior isn't valuable. On the contrary, I _highly_ recommend unit testing encoding related behavior, it can be a very tricky space. However, here - we aren't adding any behavior. The `Console` already knows how to write in utf-8 and I'm not interested in unit testing the C# standard library :)

Haard30 (dismissed):

src/AzureAuth/Program.cs
mvanchaa: Got it! Thanks Kyle!

shalinikhare27 (approved):

mvanchaa (dismissed):

src/MSALWrapper/AuthFlow/Broker.cs
kyle-rader: There are ticks `'` so it prints `Using cached account ''`

kyle-rader: This could be slightly improved to say <NO ACCOUNT> or something. What do you think?

kyle-rader:
>

kyle-rader:
> I'm not sure if I'm asking the right question. Does this mean the token contains Unicode characters and hence the JSON output or the status message contains those Unicode characters? If so, should MSAL also fix this?

Great question! The token itself only ever uses alphanumeric characters and is always valid ascii. It's the User Display name that can have any unicode character in it. So what's happening is a user with unicode chars in their display name is having those chars botched while reading from AzureAuth in a Perl script.

src/MSALWrapper/AuthFlow/Broker.cs
mvanchaa: Looks good to me :) I just saw web and devicecode code is printing it at all if the account is empty.

mvanchaa:
> > I'm not sure if I'm asking the right question. Does this mean the token contains Unicode characters and hence the JSON output or the status message contains those Unicode characters? If so, should MSAL also fix this?
> 
> Great question! The token itself only ever uses alphanumeric characters and is always valid ascii. It's the User Display name that can have any unicode character in it. So what's happening is a user with unicode chars in their display name is having those chars botched while reading from AzureAuth in a Perl script.

ah got it! That's an interesting scenario I never thought of.

goagain (approved):

Haard30 (approved):

PR #160, opened 2022-10-21
Use System Web Browser for Web mode to avoid cache poisoning

kyle-rader:
With the build system using AzureAuth directly the last source of cache poisoning is using web mode when pre-warming ado auth for Nuget. We can avoid cache poisoning by going back to system web browser when web auth is needed. Falling back to web should rarely happen now that IWA and broker are both used before web. By using system web browser instead of embedded web view - we won't risk cache poisoning since the system web browser should always be able to pass the device management CA policy properly. 

This does mean that the prompt hint will not show if being prompted with the web browser. Which is tragic. But we also expect to see it far less.

### DIFF
--- a/src/MSALWrapper/PCAWrapper.cs
+++ b/src/MSALWrapper/PCAWrapper.cs
@@ -59,9 +59,7 @@ namespace Microsoft.Authentication.MSALWrapper
         /// <summary>
         /// Gets a value indicating whether or not to use the system web browser for web mode prompts. Default: false.
         /// </summary>
-        public bool UseSystemWebBrowser { get; private set; } = false;
-
-        private bool UseEmbeddedWebView => !this.UseSystemWebBrowser;
+        public bool UseEmbeddedWebView { get; private set; } = false;
 
         /// <inheritdoc/>
         public IPCAWrapper WithPromptHint(string promptHint)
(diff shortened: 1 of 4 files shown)


src/MSALWrapper/PCAWrapper.cs
```diff
@@ -78,6 +91,7 @@ public async Task<TokenResult> GetTokenInteractiveAsync(IEnumerable<string> scop
                 {
                     Title = this.PromptHint,
                 })
+                .WithUseEmbeddedWebView(this.UseEmbeddedWebView)
```
reillysiemens, line 94: Above this we make a call to `.WithEmbeddedWebViewOptions()`. Is that still necessary if we've written this code to never use the embedded web view?

reillysiemens (commented):
Is there an MSAL bug about the unreliability of embedded web view in combination with conditional access policies? If not, should we open one? That seems to me like the sort of thing that should be resolved, and I'd like to link to it in a comment.

src/MSALWrapper/PCAWrapper.cs
kyle-rader: In theory that is also how we are supposed to set the prompt hint for WAM, but I'm not sure that's actually the case. I could have sworn we saw WAM prompts using the prompt hint in testing. But now I'm not seeing WAM use the prompt hint. I think we can open a follow up bug to figure that out.

Haard30 (approved):
Tested on Windows.

kyle-rader:
> Is there an MSAL bug about the unreliability of embedded web view in combination with conditional access policies? If not, should we open one? That seems to me like the sort of thing that should be resolved, and I'd like to link to it in a comment.

Discussed offline, unlikely that this is a priority for MSAL.NET as WAM is the future.

PR #162, opened 2022-10-24
Handling invalid AZUREAUTH_CONFIG

Haard30:
If an invalid configuration file is specified via the `AZUREAUTH_CONFIG` environment variable, then the user is given a stack trace. In contrast, if the `--config` flag is used then the user gets a proper message which says "The file '<filename>.toml' does not exist.". This is because the config flag is checked and the error is handled with [FileExists] options check, while if we use `AZUREAUTH_CONFIG` environment variable, we have to specifically catch that error.

Adding a catch statement to make sure we catch `FileNotFound` exception and provide similar experience.

Also added a test case to verify the code works as expected.

### DIFF
--- a/src/AzureAuth.Test/CommandMainTest.cs
+++ b/src/AzureAuth.Test/CommandMainTest.cs
@@ -263,6 +263,26 @@ invalid_key = ""this is not a valid alias key""
         }
 
         /// <summary>
+        ///  The test to evaluate options when config file does not exist.
+        /// </summary>
+        [Test]
+        public void TestEvaluateOptionsConfigFileDoesNotExist()
+        {
+            string configFile = RootPath("does_not_exists_config.toml");
+
+            CommandMain subject = this.serviceProvider.GetService<CommandMain>();
+            subject.AliasName = "contoso";
+            subject.ConfigFilePath = null;
+
+            // Specify config via env var
+            this.envMock.Setup(e => e.Get("AZUREAUTH_CONFIG")).Returns(configFile);
+
+            subject.EvaluateOptions().Should().BeFalse();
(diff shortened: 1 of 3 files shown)


reillysiemens (dismissed):

goagain (approved):

goagain (changes requested):

mvanchaa (dismissed):

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -262,6 +262,26 @@ public void TestEvaluateOptionsProvidedInvalidConfig()
...
+            CommandMain subject = this.serviceProvider.GetService<CommandMain>();
+            subject.AliasName = "contoso";
+            subject.ConfigFilePath = null;
+
+            // Specify config via env var
+            this.envMock.Setup(e => e.Get("AZUREAUTH_CONFIG")).Returns(configFile);
+
+            subject.EvaluateOptions().Should().BeFalse();
+            this.logTarget.Logs.Should().ContainMatch($"The file '{configFile}' does not exist.*");
+            this.envMock.Verify();
```
kyle-rader, line 282:
```suggestion
            this.envMock.VerifyAll();

```

kyle-rader (changes requested):
1 suggestion to use VerifyAll instead of Verify. This way any future additions do not require adding more calls to Verify. (I'm only 60% sure on this, I don't recall the exact behave differences, only that Verify is typically what you want to setup a custom verification on the mock, while VerifyAll ensures all things that were mocked, were called, and nothing un-mocked was called.

kyle-rader (approved):

goagain (approved):

PR #165, opened 2022-10-27
Upgrade MSAL to 4.47.2 and opt-in native broker mode

goagain:
Change MSAL NuGet package version to 4.47.2 in `MSALWrapper` and opt-in native broker mode.
To do a benchmark for new native broker mode, this feature may contain a benchmark project and introduce a 3rd party benchmark framework.

# Review
After editing csproj file, an auto format was applied. To review this PR, please the option `ignore whitespace`.

# Benchmark and Result
There is no significant performance difference. However, by using native broker, we no longer need to reference to a .net5.0 DLL.

## Run Benchmark
The Benchmark project introduces a 3rd party framework [BenchmarkDotNet](https://benchmarkdotnet.org/). To use this benchmark project, set the `MSALWrapper.Benchmark` as the startup project, and change to release mode. Then the benchmark project will compile itself again and run series profiles then print the result in console.

## Benchmark Environment
```
BenchmarkDotNet=v0.13.2, OS=Windows 11 (10.0.22621.674)
Intel Xeon W-2235 CPU 3.80GHz, 1 CPU, 12 logical and 6 physical cores
.NET SDK=6.0.402
  [Host]     : .NET 6.0.10 (6.0.1022.47605), X64 RyuJIT AVX2
  DefaultJob : .NET 6.0.10 (6.0.1022.47605), X64 RyuJIT AVX2
```

## Result
|                 Method |     Mean |    Error |   StdDev |
|----------------------- |---------:|---------:|---------:|
|  NativeBrokerBenchmark | 19.65 ms | 0.378 ms | 0.353 ms |
| ManagedBrokerBenchmark | 19.69 ms | 0.339 ms | 0.317 ms |

## Remarks
```
// * Legends *
  Mean   : Arithmetic mean of all measurements
  Error  : Half of 99.9% confidence interval
  StdDev : Standard deviation of all measurements
  1 ms   : 1 Millisecond (0.001 sec)
```

# Package Size
|   Version Name   | Extracted folder size|   
|----------------- |---------------------:|
|  Native Broker   |        62 MB        |
|  GitHub 0.6.0    |        152 MB        |


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Upgrade the Windows build to use net6 now that net5 has reached end of life.
 - Set console output encoding to `utf-8` explicitly.
+- Upgrade MSAL to 4.47.2.
```
Haard30, line 14: I think you have got the older version of master. Current changelog has 0.6.0 release in it, so this will go in Unreleased section above 0.6.0

Haard30 (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 # Unreleased
+### Changed
```
kyle-rader, line 8:
```suggestion
## Changed
```

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+- Upgrade MSAL to 4.47.2.
+- Opt-in native WAM mode and Windows SDK is removed.
+- MSALWrapper now use package reference directly.
```
kyle-rader, line 11: This is slightly too much detail for the change log. The main target of the change log is a consumer of AzureAuth and they want to know quickly,  is anything breaking, are there any new features, is anything deprecated. Here, I think 1 line to note the update to MSAL 4.47 and opt-into native broker mode (which is expected to have no negative impact on AzureAuth's current functionality) is good.

src/AzureAuth/AzureAuth.csproj
```diff
@@ -1,64 +1,56 @@
...
-
-  <!-- Assemblies we need to manually mark as roots so they are not trimmed. -->
-  <ItemGroup Condition="$([MSBuild]::IsOSPlatform('Windows'))">
-    <!-- Needed for Web mode on Windows -->
-    <TrimmerRootAssembly Include="System.Private.CoreLib" />
-    <TrimmerRootAssembly Include="WinRT.Runtime" />
-  </ItemGroup>
-</Project>
\ No newline at end of file
+    <PropertyGroup Condition="$([MSBuild]::IsOSPlatform('Windows'))">
```
kyle-rader, line 3: I'm curious why this file has been reformatted. It's not the end of the world, but I would prefer to not change the whole file if not needed, as it really hides the actual diff in this PR.

kyle-rader (changes requested):

CHANGELOG.md
goagain: I would like to keep this change until we fix the format of CHANGELOG.md

src/MSALWrapper.Benchmark/BrokerBenchmark.cs
```diff
@@ -0,0 +1,117 @@
...
+    using Microsoft.Authentication.MSALWrapper.AuthFlow;
+    using BenchmarkDotNet.Running;
+
+    /// <summary>
+    /// <see cref="BrokerBenchmark"/> class is for benchmark comparison between C++ native broker (available after Microsoft.Identity.Client 4.44) and .Net managed broker.
+    /// </summary>
+    public class BrokerBenchmark
+    {
+        private readonly Guid clientID = Guid.Parse("872cd9fa-d31f-45e0-9eab-6e460a02d1f1"); // Visual Studio client id
+        private readonly Guid tenantID = Guid.Parse("72f988bf-86f1-41af-91ab-2d7cd011db47"); // Microsoft "MSIT" corp tenant
```
reillysiemens, line 20: I'm aware that these aren't especially secret values, but could these be CLI arguments rather than hardcoded?

reillysiemens (approved):

src/MSALWrapper.Benchmark/BrokerBenchmark.cs
goagain: Yes. I can add them by introducing CLI utils, but I wonder if it is worth to do because it will make the benchmark more complex.

kyle-rader (dismissed):

src/MSALWrapper.Benchmark/BrokerBenchmark.cs
kyle-rader: You could leave them as fake Guids (all zeros) to be obvious.

goagain: @kyle-rader I can't use empty GUID, otherwise I will get error
`00000000-0000-0000-0000-000000000000 is not a valid application identifier.`

reillysiemens (approved):

kyle-rader (approved):

Issue #166, opened 2022-10-28
Install fails if PATH variable is > 1024 characters.


seanadams9:
PS C:\Users\seadams> $env:AZUREAUTH_VERSION = '0.6.0'
PS C:\Users\seadams> [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
PS C:\Users\seadams> iex "& { $(irm https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/${env:AZUREAUTH_VERSION}/install/install.ps1) } -Verbose"
VERBOSE: Installing using post-0.4.0 method
VERBOSE: Creating C:\Users\seadams\AppData\Local\Programs\AzureAuth
VERBOSE: Downloading
https://github.com/AzureAD/microsoft-authentication-cli/releases/download/0.6.0/azureauth-0.6.0-win10-x64.zip to
C:\Users\seadams\AppData\Local\Programs\AzureAuth\azureauth-0.6.0-win10-x64.zip
VERBOSE: Extracting C:\Users\seadams\AppData\Local\Programs\AzureAuth\azureauth-0.6.0-win10-x64.zip to
C:\Users\seadams\AppData\Local\Programs\AzureAuth\0.6.0
VERBOSE: Removing C:\Users\seadams\AppData\Local\Programs\AzureAuth\azureauth-0.6.0-win10-x64.zip
VERBOSE: Appending 'C:\Users\seadams\AppData\Local\Programs\AzureAuth\0.6.0' to $env:PATH

**WARNING: The data being saved is truncated to 1024 characters.**
Installed azureauth 0.6.0!

setx truncates the path to 1024 characters, opening a new shell "azureauth" isn't found on the path
. https://superuser.com/questions/387619/overcoming-the-1024-character-limit-with-setx

reillysiemens:
Thanks for the report, @seanadams9. We'll look into this right away.

reillysiemens:
FYI, using `setx` was a deliberate choice because of these properties.
https://github.com/AzureAD/microsoft-authentication-cli/blob/ca9108b18a2b608d1b73c75a03ae07b3bfb19214/install/install.ps1#L197-L200

I'm not sure we were aware of or accounted for the truncation at the time...

reillysiemens:
[...]

PR #167, opened 2022-10-28
Remove setx and use WM_SETTINGCHANGE in install script

Haard30:
Bug fix for https://github.com/AzureAD/microsoft-authentication-cli/issues/166

`setx` truncated the `$PATH` when its size exceeded 1024 chars. As we always append azureauth installation, we were able to repro and catch the bug with install script using `setx`.

As a solution, I am removing the use of `setx` and using WM_SETTINGCHANGE to notify about environment variable changes.

Testing:
Test 1:
1. Remove any AzureAuth installations.
2. Put random values in PATH env var to increase size to 1024. 
3. Run modified install script in Powershell and verify that AzureAuth new version is appended to the PATH and not truncated.

Test2:
1. Follow first 2 steps of Test1.
2. Open `cmd.exe`. Run the install script.
3. Open new cmd.exe and verify that all new processes are notified of the env var change and that azureauth is usable.

### DIFF
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -12,6 +12,20 @@ if ([string]::IsNullOrEmpty($version)) {
     Write-Error 'No $AZUREAUTH_VERSION specified, unable to download a release'
 }
 
+# Send WM_SETTINGCHANGE after changing Environment variables.
+# Refer to https://gist.github.com/alphp/78fffb6d69e5bb863c76bbfc767effda
+function Send-SettingChange {
+    Add-Type -Namespace Win32 -Name NativeMethods -MemberDefinition @"
+    [DllImport("user32.dll", SetLastError = true, CharSet = CharSet.Auto)]
+    public static extern IntPtr SendMessageTimeout(IntPtr hWnd, uint Msg, UIntPtr wParam, string lParam, uint fuFlags, uint uTimeout, out UIntPtr lpdwResult);
+"@
+    $HWND_BROADCAST = [IntPtr] 0xffff;
+    $WM_SETTINGCHANGE = 0x1a;
+    $result = [UIntPtr]::Zero
+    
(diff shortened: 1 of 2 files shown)


install/install.ps1
```diff
@@ -194,12 +206,10 @@ function Install-Post-0-4-0 {
                 $newPath = "${currentPath};${targetDirectory}"
             }
         }
-        # Update the $PATH environment variable with any modifications made above. We use `setx` here instead of
-        # directly writing back the registry value because `setx` signals new processes to pick up on the environment
-        # variable changes.
-        setx PATH $newPath > $null
+        Set-ItemProperty -Path $registryPath -Name PATH -Value $newPath
+        Send-SettingChange
```
reillysiemens, line 210: Should we use `Set-ItemProperty` or `[Environment]::SetEnvironmentVariable`? Do you know if there's an important difference there, @Haard30?

Haard30: I found using the `Set-ItemProperty` well documented [here](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/set-itemproperty?view=powershell-7.2). I am not aware of any important difference but `[Environment]::SetEnvironmentVariable` directly uses .NET class `[Environment]` and its function.

kyle-rader: Those sound different, `[Enviornment]` using Dotnet's `Enviornment` under the hood implies only setting the env var for this process and not updating the Registry.

kyle-rader:
I think we just need to always run the pipeline, and unfortunately that disables external contributions directly through this PR experience. We could still take them, just logiscally, we would have to manually pull a branch from a fork, and reopen the PR directly.

install/install.ps1
Haard30: I found this doc on `[Environment]` [http://xahlee.info/powershell/environment_variables.html](http://xahlee.info/powershell/environment_variables.html)

kyle-rader (approved):

reillysiemens (approved):

PR #170, opened 2022-11-09
Remove `--cache` option and `osxSuffix`

goagain:
Option `--cache` was hidden and only for internal use. It should be removed now because all products should not rely on it anymore.
`OSXSuffix` is also no longer exist. It will be created at where it is used.

Fix for [https://github.com/AzureAD/microsoft-authentication-cli/issues/174](https://github.com/AzureAD/microsoft-authentication-cli/issues/174)


bgavrilMS (dismissed):

src/AzureAuth/CommandMain.cs
```diff
@@ -434,13 +434,15 @@ private bool ValidateOptions()
                 validOptions = false;
             }
 
+#if PlatformWindows
             if (!this.CacheFilePath.IsValidAbsoluteFilePath())
```
mvanchaa, line 438:
```suggestion
            if (!this.cacheFilePath.IsValidAbsoluteFilePath())

```

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 # Unreleased
+### Fixed
```
kyle-rader, line 8:
```suggestion
## Fixed
```
Sections should be at 2 level indent.

kyle-rader (changes requested):
1 comment so far, but still reviewing and formulating my thoughts. I don't like injecting `#ifdefs` arbistrarily into the middle of functions it adds a layer of complexity I'd like to avoid.

src/AzureAuth/CommandMain.cs
```diff
@@ -434,13 +434,15 @@ private bool ValidateOptions()
                 validOptions = false;
             }
 
+#if PlatformWindows
```
reillysiemens, line 437: I'm wary of using conditional compilation to get out of this issue. I think there might be a logic bug we can solve elsewhere. If we can, let's explore that angle before we merge this.

reillysiemens (changes requested):

src/AzureAuth/CommandMain.cs
goagain: Removed `--cache` and all its validation code.

CHANGELOG.md
goagain: I would like to keep it until we resolve the indent issue.

src/AzureAuth/CommandMain.cs
```diff
@@ -176,37 +174,17 @@ public CommandMain(CommandExecuteEventData eventData, ITelemetryService telemetr
...
             get
             {
-                // Check command parameter first.
-                if (!string.IsNullOrEmpty(this.cacheFilePath))
-                {
-                    return this.cacheFilePath;
-                }
-
-                // Check environment variable.
-                string envCacheFile = this.env.Get(EnvVars.Cache);
```
mvanchaa, line 194: I think we should still be able to provide cache file option through env variables.

kyle-rader: There's 2 reasons I don't think we should do that:
1. We no longer hve any specific use case for doing this.
2. The main motivation for this was cache file sharing -> this is not a recommended way to achieve SSO and should be discouraged. Consumers really should know nothing about the cache file - it's an implementation of AzureAuth and removing this option all together allows us to change that implementation as needed, at will, without breaking potential unknown dependencies.

src/AzureAuth/CommandMain.cs
```diff
@@ -176,37 +174,17 @@ public CommandMain(CommandExecuteEventData eventData, ITelemetryService telemetr
         public double Timeout { get; set; } = GlobalTimeout.TotalMinutes;
 
         /// <summary>
-        /// Gets or sets the cache file name. Only available on Windows.
+        /// Gets the cache file name. Only available on Windows.
         /// </summary>
-        [Option(CacheOption, "Override the default cache file location. This option is only available on Windows.", CommandOptionType.SingleValue, ShowInHelpText = false)]
-        [LegalFilePath]
         public string CacheFilePath
```
kyle-rader, line 179: Since this isn't going to be configurable, we can further simplify things. Instead of leaving this property here and passing it down to the PCA Wrapper, the PCAWrapper can now own this path creation itself.

kyle-rader (changes requested):

src/AzureAuth.Test/CommandMainTest.cs
```diff
@@ -220,7 +220,6 @@ public void TestEvaluateOptionsProvidedAliasWithEnvVarConfig()
             string configFile = RootPath("complete.toml");
             string clientOverride = "3933d919-5ba4-4eb7-b4b1-19d33e8b82c0";
             this.fileSystem.File.WriteAllText(configFile, CompleteAliasTOML);
-            this.envMock.Setup(env => env.Get(It.IsAny<string>())).Returns((string)null);
```
mvanchaa, line 223: Is this line removed by mistake?

goagain: It's not mistake. If a mock was set but never called, there would be an error.

kyle-rader (changes requested):
The osxKeyChain suffix plays the same role as the cache file on Mac, so we should also remove the osxKeychain suffix argument from all the same places as the cache file, and create it in the same place and time that we create the cache file.

src/MSALWrapper/PCAWrapper.cs
```diff
@@ -61,6 +63,19 @@ public PCAWrapper(ILogger logger, IPublicClientApplication pca, IList<Exception>
...
+        /// Gets the cache file name. Only available on Windows.
+        /// </summary>
+        /// <param name="tenantId">The tenant ID.</param>
+        /// <returns>The absolute path for cache file corresponding to the tenant.</returns>
+        public static string GetCacheFilePath(Guid tenantId)
+        {
+            // Use default cache file path.
+            string appData = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+            string absolutePath = Path.Combine(appData, ".IdentityService", $"msal_{tenantId}.cache");
+            return absolutePath;
```
kyle-rader, line 76:
```suggestion
            return Path.Combine(Environment.SpecialFolder.LocalApplicationData.ToString(), ".IdentityService", $"msal_{tenantId}.cache");

```

kyle-rader: This is where the bug actually was - the `Enviornment.GetFolderPath` method was returning null if the path doesn't exist. Technically this could happen on Windows too if someone deleted their AppData folder. We can remove that call because the constant always returns the actual value.

kyle-rader (changes requested):

src/MSALWrapper/PCAWrapper.cs
goagain: I tried it before.
`Environment.SpecialFolder.LocalApplicationData` is an enum type `SpecialFolder`. Use `ToString()` will return `LocalApplicationData`.

src/MSALWrapper/PCACache.cs
```diff
@@ -37,20 +38,26 @@ internal class PCACache
...
-        internal PCACache(ILogger logger, Guid tenantId, string cacheFilePath, string osxKeyChainSuffix = null)
+        internal PCACache(ILogger logger, Guid tenantId)
         {
             this.logger = logger;
-            this.osxKeyChainSuffix = string.IsNullOrWhiteSpace(osxKeyChainSuffix) ? $"{tenantId}" : $"{osxKeyChainSuffix}.{tenantId}";
+            this.osxKeyChainSuffix = $"{OSXKeyChainCategory}.{tenantId}";
 
-            if (string.IsNullOrWhiteSpace(cacheFilePath))
+            this.cacheFileName = $"msal_{tenantId}.cache";
+            this.cacheDir = this.CacheServiceFoler;
```
kyle-rader, line 47: Spelling - `CacheServiceFoler`

src/MSALWrapper/PCACache.cs
```diff
@@ -37,20 +38,26 @@ internal class PCACache
...
+        }
+
+        /// <summary>
+        /// Gets the absolute path of the cache folder. Only available on Windows.
+        /// </summary>
+        public string CacheServiceFoler
```
kyle-rader, line 53: Is there any need for this to be public?

kyle-rader (changes requested):
Looking good, only 2 small comments

src/MSALWrapper/PCACache.cs
goagain: 😊

src/MSALWrapper/PCACache.cs
goagain: Changed to private

src/MSALWrapper/PCAWrapper.cs
```diff
@@ -5,6 +5,7 @@ namespace Microsoft.Authentication.MSALWrapper
 {
     using System;
     using System.Collections.Generic;
+    using System.IO;
```
reillysiemens, line 8: Why did a library get added here when we've removed code? 🤔

reillysiemens (approved):

reillysiemens (dismissed):

src/MSALWrapper/PCACache.cs
```diff
@@ -37,20 +38,26 @@ internal class PCACache
...
             {
-                throw new ArgumentNullException($"{nameof(cacheFilePath)} should not be null or whitespace.");
+                string appData = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+                string absolutePath = Path.Combine(appData, ".IdentityService");
+                return absolutePath;
             }
-
-            this.cacheFileName = Path.GetFileName(cacheFilePath);
-            this.cacheDir = Directory.GetParent(cacheFilePath).FullName;
         }
```
kyle-rader, line 61: Last recommendation, Typically Properties don't require computation. I don't think there is a good reason to have this be a property instead of a function. We are effectively memoizing the value into a member of the class on construction - so I'd prefer if we make this a method, instead of a Property.

kyle-rader (changes requested):

kyle-rader (approved):

reillysiemens (approved):

PR #172, opened 2022-11-15
Parse inner exception messages

mvanchaa:
Currently, the telemetry doesn't capture any inner exception details of any exceptions thrown. Made changes to parse inner exception messages.

Changes in this PR:
- Moved `ExceptionListToStringConverter` class to `AzureAuth` package as it seemed more appropriate.
- Created a `SerializeException` class which takes fewer required properties of the `Exception` class. This is required because when `Exception` objects are JSON parsed, some properties throw unexpected exceptions. So, we take only required properties which are known for not throwing any unexpected exceptions and then parse the custom object. More details -
-- (JSON Serialization of `System.Type` instances is not supported right now and exceptions have certain fields which are `System.Type` instances. Sample exception when serializing unsupported exception fields:
`System.NotSupportedException: 'Serialization and deserialization of 'System.Type' instances are not supported and should be avoided since they can lead to security issues. Path: $.TargetSite.DeclaringType.`)

Testing:
- Unit tests.
- Tested on dev telemetry. Screenshot below.
image


Haard30 (dismissed):

Haard30 (commented):
Please ignore if you already tested but can we also manually repro getting nested exceptions and checking if this works as expected?

melindajohnson (approved):

mvanchaa:
> Please ignore if you already tested but can we also manually repro getting nested exceptions and checking if this works as expected?

Do you mean repro MSAL nested exceptions? I tested using regular nested exceptions though.

src/MSALWrapper/ExceptionListToStringConverter.cs
```diff
@@ -72,18 +72,18 @@ public static string Execute(IEnumerable<Exception> exceptions)
         /// <returns>The <see cref="string"/>.</returns>
         private static string SingleLineException(Exception ex)
         {
-            if (ex != null)
+            var messages = new List<string>();
+            while (ex != null)
             {
                 var message = ex.Message
```
goagain, line 78: Although it is a very simple code, mixed use of `message` and `messages` is still a little confusing. I recommend using `inner_message`.

goagain: You also need to consider [`AggregateException`](https://learn.microsoft.com/en-us/dotnet/api/system.aggregateexception?view=net-7.0)

mvanchaa: `inner_message` would mean parsing the inner exception message of a given exception. But we are not doing that and parsing the exception's message.
I have seen this as a common practice to not differentiate a list/array variable from a single object variable, just by using the plural form of the name.
But if you all think it's confusing, I can change it to may be `messageList` or something like that

Haard30:
> > Please ignore if you already tested but can we also manually repro getting nested exceptions and checking if this works as expected?
> 
> Do you mean repro MSAL nested exceptions? I tested using regular nested exceptions though.

Yes - I was just checking if can repro MSAL nested exception somehow to verify that the format is no different. But if you have verified with regular nested exceptions and as you also have a test, that should be fine as well :)

src/MSALWrapper/ExceptionListToStringConverter.cs
mvanchaa: @goagain that's a good point about AggregateExceptions. I will update this. I just saw there is a method that already exists in our code base to extract all exceptions from a given exception.

reillysiemens: I can't speak to `AggregateException`, but I agree with @mvanchaa's sentiment that this is common practice for differentiating between a list and elements of that list in a loop. Personally, I don't find it confusing and would support keeping the names as is.

reillysiemens: As for `messageList`, regardless of programming language I'm almost always against including the type name in a variable name, but especially in a statically typed language where intellisense or other helpful tooling makes it clear. I generally prefer to see containers named after what they contain rather than their what they contain **and** their type.

src/MSALWrapper/ExceptionListToStringConverter.cs
```diff
@@ -74,11 +74,19 @@ private static string SingleLineException(Exception ex)
         {
             if (ex != null)
```
reillysiemens, line 75: What if we flipped this and did
```csharp
if (ex == null) {
    return "null";
}
```
and _then_ put the remainder of the logic? The early return would place the `"null"` closer to the condition that causes it and you wouldn't need the `else` clause.

src/MSALWrapper/ExceptionListToStringConverter.cs
```diff
@@ -74,11 +74,19 @@ private static string SingleLineException(Exception ex)
...
-                return $"{ex.GetType()}: {message}";
+                IEnumerable<Exception> innerExceptions = ExceptionExtensions.GetAllExceptions(ex);
+                var messages = new List<string>();
+
+                foreach (Exception innerException in innerExceptions)
+                {
+                    var message = innerException.Message
+                                    .Replace("\n", string.Empty)
+                                    .Replace("\r", string.Empty)
+                                    .Replace("\t", string.Empty);
```
reillysiemens, line 85: I don't know that I ever bothered to look before now. Is there a reason we get rid of all these characters? Why do we make a single line exception? 🤔

mvanchaa: We wanted to single line exception to have just one value in the telemetry for all the exception messages. Otherwise, each exception message needs to be associated with a unique key in the telemetry.

I guess getting rid of the tabs and lines is to have trimmed exception messages and make them easy to read in telemetry?

kyle-rader: Yes, this is about making it clear how many exceptions are getting logged. We could alternatively format them as a JSON array or something but that would take longer.

src/MSALWrapper.Test/ExceptionListToStringConverterTest.cs
```diff
@@ -224,5 +224,37 @@ public void ExtractCorrelationIDs_FromMSALExceptionList_WithOneNullCorrelationID
...
+            List<Exception> exceptionList = new ()
+            {
+                new Exception("This is the first exception"),
+                new MsalServiceException("2", "This is the second exception", new MsalClientException("3", "This is the inner exception of the second exception")),
+            };
+            var exceptionString = ExceptionListToStringConverter.Execute(exceptionList);
+
+            exceptionString.Should().Be("System.Exception: This is the first exception\n" +
+                "Microsoft.Identity.Client.MsalServiceException: This is the second exception\n" +
+                "Microsoft.Identity.Client.MsalClientException: This is the inner exception of the second exception");
```
kyle-rader, line 240: The expectation right now is that each top level exception we catch is on it's own line. So with this format we would lose the connection that an inner exception was actually an inner exception of the previous, and not a new top level exception we caught. Does that make sense? Perhaps we should move to formatting the exception lists to JSON so that inner exceptions can be recursively nested as needed?

kyle-rader (changes requested):
Some design thoughts to consider. Should we format the exception as JSON? (which is parsable in Kusto)

src/MSALWrapper.Test/ExceptionListToStringConverterTest.cs
mvanchaa: Yes Kyle! I thought about this scenario and in my head, I justified that the inner exceptions come right after the original exception, but you are right! I will try formatting the exceptions to a JSON array/list and see how it goes.

mvanchaa: I just remembered why we didn't do it before. Earlier we thought of serializing the exceptions, but serialization of 'System.Type' instances is not supported right now and exceptions have certain fields which are 'System.Type' instances. 

So, I will only extract messages from these exceptions and not anything else while serializing.

mvanchaa: Sample exception when serializing unsupported exception fields:
`System.NotSupportedException: 'Serialization and deserialization of 'System.Type' instances are not supported and should be avoided since they can lead to security issues. Path: $.TargetSite.DeclaringType.'`

mvanchaa: Nvm, I tried Newtonsoft Json library and that doesn't seem to have any issues.

src/AzureAuth/ExceptionListToStringConverter.cs
```diff
@@ -51,39 +52,58 @@ public static List<string> ExtractCorrelationIDsFromException(IList<Exception> e
...
+
+            return JsonSerializer.Serialize(customExceptions);
         }
+    }
 
+    /// <summary>
+    /// CustomException class with only select properties of Exception class.
+    /// </summary>
+    internal class CustomException
+    {
```
mvanchaa, line 81: I'm open to suggestions for this class name. :)

src/AzureAuth/ExceptionListToStringConverter.cs
```diff
@@ -51,39 +52,58 @@ public static List<string> ExtractCorrelationIDsFromException(IList<Exception> e
...
-            if (ex != null)
+            if (exception?.Message != null)
             {
-                var message = ex.Message
-                                .Replace("\n", string.Empty)
-                                .Replace("\r", string.Empty)
-                                .Replace("\t", string.Empty);
-                return $"{ex.GetType()}: {message}";
+                var singleLineMessage = exception.Message
+                                    .Replace("\n", string.Empty)
```
kyle-rader, line 91: Now that we are formatting the whole list as a JSON array, we don't need to replace newlines, but it will still make querying easier if we replace `\r` with nothing. This effectively says, translate `\r\n` -> `\n` so we preserve the newline, but in a way that will make querying or splitting on them easy if needed. We can leave `\t` in the message.

src/AzureAuth/ExceptionListToStringConverter.cs
```diff
@@ -51,39 +52,58 @@ public static List<string> ExtractCorrelationIDsFromException(IList<Exception> e
         }
 
         /// <summary>
-        /// Executes the convertion of exceptions to a string.
+        /// Executes the convertion of exceptions to a JSON string.
```
kyle-rader, line 55:
```suggestion
        /// Converts a list of exceptions to a JSON string.

```

src/AzureAuth/ExceptionListToStringConverter.cs
```diff
@@ -51,39 +52,58 @@ public static List<string> ExtractCorrelationIDsFromException(IList<Exception> e
...
             }
-            else
+
+            if (exception?.InnerException != null)
             {
-                return "null";
+                this.InnerException = new CustomException(exception.InnerException);
             }
         }
+
```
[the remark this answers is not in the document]
kyle-rader: Similarly, it would be helpful if each exception JSON object had an AAD error code if one was present in the message. These are the strings starting with `AADSTS` ([listed here](https://learn.microsoft.com/en-us/azure/active-directory/develop/reference-aadsts-error-codes) ). So we could look pull that out as well and make a property for it if preset. This will make bucketing these MSAL exceptions far far easier in App Insights.

kyle-rader (changes requested):
Some comments on extending (and renaming) the CustomException, as this will greatly improve our ability to query and bucket exceptions :) (fyi @melindajohnson )

src/AzureAuth/ExceptionListToStringConverter.cs
mvanchaa: That's perfect!

src/AzureAuth/ExceptionListToStringConverter.cs
mvanchaa: @kyle-rader I was just trying to understand the difference between `\r` and `\n`. Looks like `\r` is a new line character on mac, `\n` is a new line character on linux and `\r\n` is on windows. So should I just leave all of them as is along with `\t`?

kyle-rader: `\r` is the "Carriage return" (symbolic of a typewriter moving back to the start o the line or returning the carriage back to the starting position). In Computers this is like back to start of the current line, opposed to a new line or "line feed" - also from type writers literally feeding the paper up to make a new line.

Today, `\r` is only a valid line ending on it's own on obscure systems.  Mac and Linux both use `\n` while Windows still uses `\r\n`. (Also seen as `LF` pronounced L-F and `CRLF`, pronounced "curl-if").

mvanchaa: Thanks, @kyle-rader!

src/AzureAuth/ExceptionListToStringConverter.cs
```diff
@@ -0,0 +1,133 @@
...
+            }
+
+            return JsonSerializer.Serialize(customExceptions);
+        }
+    }
+
+    /// <summary>
+    /// CustomException class with only select properties of Exception class.
+    /// </summary>
+    internal class SerializableException
```
reillysiemens, line 80: I think this can be a `private` class _on_ the `ExceptionListToStringConverter` itself rather than just in the same namespace.

src/AzureAuth/ExceptionListToStringConverter.cs
```diff
@@ -0,0 +1,133 @@
...
+            if (exception != null)
+            {
+                if (exception.Message != null)
+                {
+                    this.Message = exception.Message.Replace("\r", string.Empty);
+                }
+
+                if (exception.InnerException != null)
+                {
+                    this.InnerException = new SerializableException(exception.InnerException);
```
reillysiemens, line 97: If we can, let's document that this is a recursive call. It wasn't immediately obvious to me at a glance that that's how this was implemented.

src/AzureAuth.Test/ExceptionListToStringConverterTest.cs
```diff
@@ -0,0 +1,407 @@
...
+        [Test]
+        public void EmptyExceptionList()
+        {
+            IEnumerable<Exception> exceptions = new List<Exception>();
+
+            // Act
+            string result = ExceptionListToStringConverter.Execute(exceptions);
+
+            // Assert
+            result.Should().BeNull();
```
kyle-rader, line 23: Rather than `null` - I think it will be helpful to have the empty set be represented by the string representation of the empty list i.e. `"[]"`. The reason for this is consistency in the data when querying. If the value is null - then they key I think is actually not included on the event which will make querying harder. If it's an empty list, this will make querying easier i think, because we can do list operations on `error_messages` and not worry about it existing or not. For instance we can easily query for where `error_messages` has a count of 0, or 1, etc. Or search where error_messages contains one with some text, etc.

src/AzureAuth.Test/ExceptionListToStringConverterTest.cs
```diff
@@ -0,0 +1,407 @@
...
+        [Test]
+        public void NullExceptionList()
+        {
+            IEnumerable<Exception> exceptions = null;
+
+            // Act
+            string result = ExceptionListToStringConverter.Execute(exceptions);
+
+            // Assert
+            result.Should().BeNull();
```
kyle-rader, line 35: I think this case should throw - if given a null list to begin with, that likely indicates a bug. I don't think our callers have a valid scenario where the list is actually null and I don't think we want that to be supported.

src/AzureAuth/ExceptionListToStringConverter.cs
```diff
@@ -0,0 +1,133 @@
...
+    /// The exception list to string converter.
+    /// </summary>
+    public static class ExceptionListToStringConverter
+    {
+        /// <summary>
+        /// Extracts correlation IDs from the list of exceptions.
+        /// </summary>
+        /// <param name="exceptions">List of exceptions from which correlation IDs are extracted.</param>
+        /// <returns>List of correlation IDs.</returns>
+        public static List<string> ExtractCorrelationIDsFromException(IList<Exception> exceptions)
```
reillysiemens, line 22: Maybe we don't need to extract these correlation IDs in exactly this way? I think we lose some information by doing so.

src/AzureAuth.Test/ExceptionListToStringConverterTest.cs
```diff
@@ -0,0 +1,407 @@
...
+            };
+
+            // Act
+            string result = ExceptionListToStringConverter.Execute(exceptions);
+            string expectedResult = @"[" +
+                                        @"{""Message"":""This is an exception""," +
+                                        @"""InnerException"":null," +
+                                        @"""ExceptionType"":""System.Exception""," +
+                                        @"""AADErrorCode"":null}" +
+                                    @"]";
```
kyle-rader, line 53: All of your string literals can be simplified - when using the `@` string literal symbol, you can do multi-line strings right in the code, so you do not need every line to be it's own complete string. IDK if this gets optimized away by the compiler or not (the string concatenation), but it will be easier to read and edit if it's 1 big string literal.

kyle-rader (changes requested):
A couple more comments on error handling, and multi-line strings in the tests :)

src/AzureAuth.Test/ExceptionListToStringConverterTest.cs
reillysiemens: Taking this one step further, we could actually not do string comparisons for these tests at all, but do comparisons against deserialized objects to validate that our JSON is correct.

kyle-rader: Tl;Dr discussion offline - actually let's not do string comparisons for our assertions, let's assert that we do have deserializable JSON.

src/AzureAuth/ExceptionListToStringConverter.cs
```diff
@@ -0,0 +1,133 @@
...
+    /// </summary>
+    internal class SerializableException
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SerializableException"/> class.
+        /// </summary>
+        /// <param name="exception">Exception.</param>
+        public SerializableException(Exception exception)
+        {
+            if (exception != null)
```
reillysiemens, line 88: I think you can simplify the body of this method by doing some early returns instead of having such deep nesting.

src/AzureAuth/ExceptionListToStringConverter.cs
reillysiemens: After further consideration, does this even need to be a class? Could it just be a private method instead? 🤔

mvanchaa: Good point, but I don't think so, Tucker. Because the JSON serialization libraries work on a class. If we write a method, we will have to go by our own serialization method as well.

src/AzureAuth/CommandMain.cs
```diff
@@ -251,27 +251,19 @@ public EventData AuthFlowEventData(AuthFlowResult result)
...
             {
                 var error_messages = ExceptionListToStringConverter.Execute(result.Errors);
                 eventData.Add("error_messages", error_messages);
-                correlationIDs = ExceptionListToStringConverter.ExtractCorrelationIDsFromException(result.Errors);
             }
 
             if (result.Success)
             {
-                correlationIDs.Add(result.TokenResult.CorrelationID.ToString());
+                eventData.Add("msal_correlation_ids", result.TokenResult.CorrelationID.ToString());
```
kyle-rader, line 262: This key should become singular since we are only adding the 1 correlation id from the successful flow.

src/AzureAuth/ExceptionListToStringConverter.cs
```diff
@@ -14,89 +14,86 @@ namespace Microsoft.Authentication.AzureAuth
...
+            /// <summary>
+            /// Initializes a new instance of the <see cref="SerializableException"/> class.
+            /// </summary>
+            /// <param name="exception">Exception.</param>
+            public SerializableException(Exception exception)
             {
+                if (exception == null)
+                {
+                    return;
+                }
```
kyle-rader, line 48: Does this need to return null, or is handing this a null value an invalid state? Can we throw instead?

kyle-rader (changes requested):
This is really coming along nicely! Those tests are no far more readable and ensure what we send will be deserializable! I've just left 2 more comments to address :)

src/AzureAuth/ExceptionListToStringConverter.cs
mvanchaa: Yea, since we are calling recursively on inner exceptions we need this.

mvanchaa: Or may be that check can be done before the recursive call

mvanchaa: @kyle-rader I have addressed the new comments :)

src/AzureAuth/ExceptionListToStringConverter.cs
```diff
@@ -0,0 +1,137 @@
...
+            public SerializableException(Exception exception)
+            {
+                if (exception == null)
+                {
+                    throw new ArgumentNullException(nameof(exception));
+                }
+
+                if (exception.Message != null)
+                {
+                    this.Message = exception.Message.Replace("\r", string.Empty);
```
kyle-rader, line 52: This data sanitization is only being done in the constructor, but this is public property that can be reset. As a general guideline, I would move this kind of cleaning into the `set` method of the property so that the logic is "owned" by the property itself. That way the property always has `x` removed no matter where it is set from.

Looking back it took

kyle-rader: A second comment on this, it took me a minute to remember why we are only removing `\r` here, I recalled it's because we want to only have \n in the messages to make them easier to parse. I can't remember why this was still desired after the move to use JSON serialization though. I'm not able to think of a reason why we would need to be able to always split the `Message` on newlines, in Kusto, since each Message is going to be in it's own object.

kyle-rader (changes requested):
Small comment on data consistency.

src/AzureAuth/ExceptionListToStringConverter.cs
kyle-rader: Ah re-reading, I had said we don't need to strip them, but we can force them to always be LF, which is what this does. So only pay attention to my first note here about data consistency, cleaning in the setter :)

kyle-rader (approved):
@mvanchaa great work iterating on this!!

shalinikhare27 (approved):

reillysiemens (approved):

src/AzureAuth/ExceptionListToStringConverter.cs
```diff
@@ -0,0 +1,146 @@
...
+            /// <summary>Gets or Sets AADErrorCode.</summary>
+            public string AADErrorCode { get; set; }
+
+            /// <summary>Gets or Sets CorrelationIds.</summary>
+            public string CorrelationId { get; set; }
+
+            /// <summary>
+            /// Sets the error code.
+            /// </summary>
+            /// <param name="exception">exception from which error code is extracted</param>
```
goagain, line 103:
```suggestion
            /// <param name="exception">exception from which error code is extracted.</param>
```

goagain: Resolve it in a different PR.

PR #176, opened 2022-12-10
Add unlit Azure DevOps PAT client classes

reillysiemens:
This PR adds a `PatClient` class to the `Microsoft.Authentication.AdoPat` namespace. It also adds supporting interfaces and classes, namely
- a `PatClientException` for raising a subset of error conditions
- a corresponding `TokensHttpClientWrapper` class, which wraps `TokensHttpClient` from [`Microsoft.VisualStudio.Services.Client` 19.212.0-preview](https://www.nuget.org/packages/Microsoft.VisualStudio.Services.Client/19.212.0-preview)
- an `ITokensHttpClientWrapper` interface to make testing easier
- tests in the `Microsoft.Authentication.AdoPat.Test` namespace

There will need to be a handful of follow up PRs to fully implement this feature. I consider this phase 1 of 4. Until all of the phases are complete this code won't be "lit" or "live" in the application and should not factor into production builds.

This entity relationship diagram shows the rough layout of parts needed to make this work. This PR covers `PatClient` and `TokensHttpClientWrapper`, so there's still work to be done for the other pieces of the puzzle.

```mermaid
erDiagram
    CLI ||..|| PatManager : uses
    PatCache }|..|{ MSALCacheWrapper : uses
```

I've marked this PR as a draft for now because I'd like to get feedback while I work on the other phases. If it's well received, however, it should be safe to merge it, and I can lift the draft status. There are some questions marked as `XXX` or `TODO` that I'm particularly interested in getting feedback on, so keep an eye out for those.

### DIFF
--- a/src/AdoPat.Test/PatClientTest.cs
+++ b/src/AdoPat.Test/PatClientTest.cs
@@ -14,47 +14,63 @@ namespace Microsoft.Authentication.AdoPat.Test
 
     public class PatClientTest
     {
+        private const bool AllOrgs = false;
+        private const string DisplayName = "Test PAT";
+        private const string Scope = "test.scope";
+
+        // This is a test token. A real value would be a much longer string.
+        private const string Token = "Test Token";
+
+        // The Unix Epoch is used as an obviously fake test time which occurs in the past and cannot accidentally be valid.
+        private readonly DateTime unixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+
+        // This list of accounts uses dummy data, not real accounts.
+        private readonly List<Guid> targetAccounts = new List<Guid> { new Guid("b7b59161-cd70-46e9-aca5-883f24060eb1") };
+
+        // This is a dummy authorization ID, not valid in any real contexts.
+        private readonly Guid authorizationId = new Guid("ee0c5586-a96f-4a44-b1d9-8613028b1078");
+
         [Test]
         public async Task CreatePatAsync()
         {
             // Arrange
-            var displayName = "Test PAT";
-            var scope = "test.scope";
-            var validTo = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
-            var allOrgs = false;
             var patTokenCreateRequest = new PatTokenCreateRequest
             {
-                DisplayName = displayName,
-                Scope = scope,
-                ValidTo = validTo,
-                AllOrgs = allOrgs,
+                DisplayName = DisplayName,
+                Scope = Scope,
+                ValidTo = this.unixEpoch.AddDays(7),
+                AllOrgs = AllOrgs,
             };
             var patToken = new PatToken
             {
-                DisplayName = displayName,
-                Scope = scope,
-                TargetAccounts = new List<Guid> { new Guid("b7b59161-cd70-46e9-aca5-883f24060eb1") },
-                ValidTo = validTo,
-                ValidFrom = DateTime.UtcNow,
-                AuthorizationId = new Guid("ee0c5586-a96f-4a44-b1d9-8613028b1078"),
-                Token = "A real token would be much, much longer.",
+                DisplayName = DisplayName,
+                Scope = Scope,
+                TargetAccounts = this.targetAccounts,
+                ValidTo = this.unixEpoch.AddDays(7),
+                ValidFrom = this.unixEpoch,
+                AuthorizationId = this.authorizationId,
+                Token = Token,
             };
-            var expectedResult = new PatTokenResult { PatToken = patToken, PatTokenError = SessionTokenError.None };
+            var patTokenResult = new PatTokenResult
+            {
+                PatToken = patToken,
+                PatTokenError = SessionTokenError.None,
+            };
 
-            var client = new Mock<ITokensHttpClientProvider>(MockBehavior.Strict);
-            client.Setup(thcp => thcp.CreatePatAsync(
+            var client = new Mock<ITokensHttpClientWrapper>(MockBehavior.Strict);
+            client.Setup(c => c.CreatePatAsync(
                 It.IsAny<PatTokenCreateRequest>(),
                 It.IsAny<object>(),
                 It.IsAny<CancellationToken>()))
-            .ReturnsAsync(expectedResult);
+            .ReturnsAsync(patTokenResult);
 
             var patClient = new PatClient(client.Object);
 
             // Act
-            var patTokenResult = await patClient.CreatePatAsync(patTokenCreateRequest);
+            var renewedPatToken = await patClient.CreatePatAsync(patTokenCreateRequest);
 
             // Assert
-            patTokenResult.Should().BeEquivalentTo(expectedResult);
+            renewedPatToken.Should().BeEquivalentTo(patToken);
         }
 
         [Test]
@@ -64,8 +80,8 @@ namespace Microsoft.Authentication.AdoPat.Test
             var emptyPage = new PagedPatTokens(continuationToken: string.Empty, patTokens: new List<PatToken> { });
             var expectedTokens = new HashSet<PatToken>();
 
-            var client = new Mock<ITokensHttpClientProvider>(MockBehavior.Strict);
-            client.SetupSequence(thcp => thcp.ListPatsAsync(
+            var client = new Mock<ITokensHttpClientWrapper>(MockBehavior.Strict);
+            client.Setup(c => c.ListPatsAsync(
                 It.IsAny<DisplayFilterOptions?>(),
                 It.IsAny<SortByOptions?>(),
                 It.IsAny<bool?>(),
@@ -94,8 +110,8 @@ namespace Microsoft.Authentication.AdoPat.Test
             var page2 = new PagedPatTokens(continuationToken: string.Empty, patTokens: new List<PatToken> { pat2 });
             var expectedTokens = new HashSet<PatToken>() { pat1, pat2 };
 
-            var client = new Mock<ITokensHttpClientProvider>(MockBehavior.Strict);
-            client.SetupSequence(thcp => thcp.ListPatsAsync(
+            var client = new Mock<ITokensHttpClientWrapper>(MockBehavior.Strict);
+            client.SetupSequence(c => c.ListPatsAsync(
                 It.IsAny<DisplayFilterOptions?>(),
                 It.IsAny<SortByOptions?>(),
                 It.IsAny<bool?>(),
@@ -113,6 +129,94 @@ namespace Microsoft.Authentication.AdoPat.Test
 
             // Assert
             activePats.Should().BeEquivalentTo(expectedTokens);
+        }
+
+        [Test]
+        public async Task RegeneratePatAsync_RevokesOldPatAndReturnsNewPat()
+        {
+            // Arrange
+            var issued = this.unixEpoch.AddDays(-7);
+            var validTo = this.unixEpoch;
+            var regeneratedValidTo = this.unixEpoch.AddDays(7);
+            var patToken = new PatToken
+            {
+                DisplayName = DisplayName,
+                Scope = Scope,
+                TargetAccounts = this.targetAccounts,
+                ValidTo = validTo,
+                ValidFrom = issued,
+                AuthorizationId = this.authorizationId,
+                Token = Token,
+            };
+
+            var expectedToken = patToken;
+            expectedToken.ValidTo = regeneratedValidTo;
+            expectedToken.ValidFrom = validTo;
+
+            var patTokenResult = new PatTokenResult { PatToken = expectedToken };
+
+            var client = new Mock<ITokensHttpClientWrapper>(MockBehavior.Strict);
+            client.Setup(c => c.CreatePatAsync(
+                It.IsAny<PatTokenCreateRequest>(),
+                It.IsAny<object>(),
+                It.IsAny<CancellationToken>()))
+            .ReturnsAsync(patTokenResult);
+
+            // This returns a Task because it's normally a void method and
+            // moq has some trouble with that: https://stackoverflow.com/a/66799787/3288364
+            client.Setup(c => c.RevokeAsync(
+                It.IsAny<Guid>(),
+                It.IsAny<object>(),
+                It.IsAny<CancellationToken>()))
+            .Returns(Task.CompletedTask);
+
+            var patClient = new PatClient(client.Object);
+
+            // Act
+            var regeneratedPat = await patClient.RegeneratePatAsync(patToken, regeneratedValidTo);
+
+            // Assert
+            regeneratedPat.Should().BeEquivalentTo(expectedToken);
+        }
+
+        [Test]
+        public async Task RegeneratePatAsync_CreationFailureNullPatToken()
+        {
+            // Arrange
+            var issued = this.unixEpoch.AddDays(-7);
+            var validTo = this.unixEpoch;
+            var regeneratedValidTo = this.unixEpoch.AddDays(7);
+            var patToken = new PatToken
+            {
+                DisplayName = DisplayName,
+                Scope = Scope,
+                TargetAccounts = this.targetAccounts,
+                ValidTo = validTo,
+                ValidFrom = issued,
+                AuthorizationId = this.authorizationId,
+                Token = Token,
+            };
+            var failingResult = new PatTokenResult
+            {
+                PatToken = null,
+                PatTokenError = SessionTokenError.InvalidAuthorizationId,
+            };
+
+            var client = new Mock<ITokensHttpClientWrapper>(MockBehavior.Strict);
+            client.Setup(c => c.CreatePatAsync(
+                It.IsAny<PatTokenCreateRequest>(),
+                It.IsAny<object>(),
+                It.IsAny<CancellationToken>()))
+            .ReturnsAsync(failingResult);
+
+            var patClient = new PatClient(client.Object);
+
+            // Act
+            Func<Task> act = () => patClient.RegeneratePatAsync(patToken, regeneratedValidTo);
+
+            // Assert
+            await act.Should().ThrowAsync<PatClientException>()
+                .WithMessage("Failed to create PAT during regeneration: InvalidAuthorizationId");
         }
     }
 }
--- a/src/AdoPat.Test/PatClientTest.cs
+++ b/src/AdoPat.Test/PatClientTest.cs
@@ -176,6 +176,7 @@ namespace Microsoft.Authentication.AdoPat.Test
             var regeneratedPat = await patClient.RegeneratePatAsync(patToken, regeneratedValidTo);
 
             // Assert
+            client.VerifyAll();
             regeneratedPat.Should().BeEquivalentTo(expectedToken);
         }
 
--- a/src/AdoPat.Test/PatClientTest.cs
+++ b/src/AdoPat.Test/PatClientTest.cs
@@ -74,11 +74,11 @@ namespace Microsoft.Authentication.AdoPat.Test
         }
 
         [Test]
-        public async Task GetActivePatsAsync_ReturnsEmptySetWithNoResults()
+        public async Task GetActivePatsAsync_ReturnsEmptyDictionaryWithNoResults()
         {
             // Arrange
             var emptyPage = new PagedPatTokens(continuationToken: string.Empty, patTokens: new List<PatToken> { });
-            var expectedTokens = new HashSet<PatToken>();
+            var expectedTokens = new Dictionary<Guid, PatToken>();
 
             var client = new Mock<ITokensHttpClientWrapper>(MockBehavior.Strict);
             client.Setup(c => c.ListPatsAsync(
@@ -101,14 +101,14 @@ namespace Microsoft.Authentication.AdoPat.Test
         }
 
         [Test]
-        public async Task GetActivePatsAsync_ReturnsSetWithAllPages()
+        public async Task GetActivePatsAsync_ReturnsDictionaryWithAllPages()
         {
             // Arrange
-            var pat1 = new PatToken { DisplayName = "PAT 1" };
-            var pat2 = new PatToken { DisplayName = "PAT 2" };
+            var pat1 = new PatToken { AuthorizationId = new Guid("8e510ed2-f485-4dfb-963c-bbaa30d60ba0"), DisplayName = "PAT 1" };
+            var pat2 = new PatToken { AuthorizationId = new Guid("40ba0030-1dad-4bbe-82c5-8c49266d4e1d"), DisplayName = "PAT 2" };
             var page1 = new PagedPatTokens(continuationToken: "Page 1", patTokens: new List<PatToken> { pat1 });
             var page2 = new PagedPatTokens(continuationToken: string.Empty, patTokens: new List<PatToken> { pat2 });
-            var expectedTokens = new HashSet<PatToken>() { pat1, pat2 };
+            var expectedTokens = new Dictionary<Guid, PatToken>() { { pat1.AuthorizationId, pat1 }, { pat2.AuthorizationId, pat2 } };
 
             var client = new Mock<ITokensHttpClientWrapper>(MockBehavior.Strict);
             client.SetupSequence(c => c.ListPatsAsync(
--- a/src/AdoPat.Test/PatClientTest.cs
+++ b/src/AdoPat.Test/PatClientTest.cs
@@ -216,7 +216,7 @@ namespace Microsoft.Authentication.AdoPat.Test
 
             // Assert
             await act.Should().ThrowAsync<PatClientException>()
-                .WithMessage("Failed to create PAT during regeneration: InvalidAuthorizationId");
+                .WithMessage("Failed to create PAT: InvalidAuthorizationId");
         }
     }
 }
--- a/src/AdoPat.Test/PatClientTest.cs
+++ b/src/AdoPat.Test/PatClientTest.cs
@@ -1,0 +1,118 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.AdoPat.Test
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using FluentAssertions;
+    using Microsoft.VisualStudio.Services.DelegatedAuthorization;
+    using Moq;
+    using NUnit.Framework;
+
+    public class PatClientTest
+    {
+        [Test]
+        public async Task CreatePatAsync()
+        {
+            // Arrange
+            var displayName = "Test PAT";
+            var scope = "test.scope";
+            var validTo = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+            var allOrgs = false;
+            var patTokenCreateRequest = new PatTokenCreateRequest
+            {
+                DisplayName = displayName,
+                Scope = scope,
+                ValidTo = validTo,
+                AllOrgs = allOrgs,
+            };
+            var patToken = new PatToken
+            {
+                DisplayName = displayName,
+                Scope = scope,
+                TargetAccounts = new List<Guid> { new Guid("b7b59161-cd70-46e9-aca5-883f24060eb1") },
+                ValidTo = validTo,
+                ValidFrom = DateTime.UtcNow,
+                AuthorizationId = new Guid("ee0c5586-a96f-4a44-b1d9-8613028b1078"),
+                Token = "A real token would be much, much longer.",
+            };
+            var expectedResult = new PatTokenResult { PatToken = patToken, PatTokenError = SessionTokenError.None };
+
+            var client = new Mock<ITokensHttpClientProvider>(MockBehavior.Strict);
+            client.Setup(thcp => thcp.CreatePatAsync(
+                It.IsAny<PatTokenCreateRequest>(),
+                It.IsAny<object>(),
+                It.IsAny<CancellationToken>()))
+            .ReturnsAsync(expectedResult);
+
+            var patClient = new PatClient(client.Object);
+
+            // Act
+            var patTokenResult = await patClient.CreatePatAsync(patTokenCreateRequest);
+
+            // Assert
+            patTokenResult.Should().BeEquivalentTo(expectedResult);
+        }
+
+        [Test]
+        public async Task GetActivePatsAsync_ReturnsEmptySetWithNoResults()
+        {
+            // Arrange
+            var emptyPage = new PagedPatTokens(continuationToken: string.Empty, patTokens: new List<PatToken> { });
+            var expectedTokens = new HashSet<PatToken>();
+
+            var client = new Mock<ITokensHttpClientProvider>(MockBehavior.Strict);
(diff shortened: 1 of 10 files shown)


jcreamer898:
Super excited for this!

src/AdoPat/PatClient.cs
```diff
@@ -0,0 +1,139 @@
...
+    using System.Threading.Tasks;
+    using Microsoft.VisualStudio.Services.DelegatedAuthorization;
+
+    /// <summary>
+    /// An abstraction over common operations with the Azure DevOps PAT Lifecycle Management REST API.
+    /// </summary>
+    public class PatClient
+    {
+        private const int PageSize = 100; // Using the maximum allowable page size allows us to reduce HTTP calls.
+        private const bool AllOrgs = false; // Azure DevOps have recommended we always set this to false.
```
mvanchaa, line 18: [nit] I couldn't make out from here what `AllOrgs` is. And may be provide a reference to it about setting it to false.

src/AdoPat/TokensHttpClientWrapper.cs
```diff
@@ -0,0 +1,76 @@
...
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.VisualStudio.Services.DelegatedAuthorization;
+    using Microsoft.VisualStudio.Services.DelegatedAuthorization.Client;
+    using Microsoft.VisualStudio.Services.WebApi;
+
+    /// <summary>
+    /// A thin wrapper around a <see cref="TokensHttpClient"/> that implements the <see cref="ITokensHttpClientWrapper"/> interface.
```
mvanchaa, line 14: Is this class `TokensHttpClient` implemented in a different PR?

mvanchaa: oh! I see it's a VS library.

src/AdoPat/PatClient.cs
```diff
@@ -0,0 +1,139 @@
...
+        public async Task<PatToken> CreatePatAsync(
+            PatTokenCreateRequest patTokenCreateRequest,
+            CancellationToken cancellationToken = default)
+        {
+            var patTokenResult = await this.client.CreatePatAsync(
+                patTokenCreateRequest,
+                cancellationToken: cancellationToken)
+            .ConfigureAwait(false);
+
+            if (patTokenResult.PatToken == null)
```
mvanchaa, line 49: Will there be any possibility of creating a `PatToken` successfully but the `PatTokenError` is not empty?

src/AdoPat/PatClient.cs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        /// <summary>
+        /// Creates a new PAT with a new 'valid to' date by replicating an existing one.
+        /// </summary>
+        /// <param name="patToken">An existing <see cref="PatToken"/>.</param>
+        /// <param name="validTo">The new expiration date for the regenerated token.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
+        /// <returns>A <see cref="Task"/> returning a <see cref="PatToken"/>.</returns>
+        public async Task<PatToken> RegeneratePatAsync(
```
mvanchaa, line 96: How is a user going to call this? Will they provide the `DisplayName` of the PAT?

src/AdoPat/PatClient.cs
```diff
@@ -0,0 +1,139 @@
...
+
+            var renewedPatToken = await this.CreatePatAsync(
+                patTokenCreateRequest,
+                cancellationToken: cancellationToken)
+            .ConfigureAwait(false);
+
+            // XXX: If we create a new PAT, but fail to revoke the old PAT, should
+            // we continue or should we try to undo the creation and bail? It's
+            // possible this could fail because the token has already been revoked.
+            // Is that an error we should allow?
```
mvanchaa, line 130: Can two PATs have same `DisplayName`? 
If yes, then there will be a new PAT and an old PAT (which failed to get revoked) in system which I think should be okay, although I think we need to let the user know.
Since you are generating a new PAT before revoking it, I'm assuming you can have the same `DisplayName` for multiple tokens?

mvanchaa (commented):
Looks good to me. I liked the ER diagram and it made easy to understand the high-level design and workflow.

src/AdoPat/PatClient.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            this.client = client;
+        }
+
+        /// <summary>
+        /// Creates a new PAT.
+        /// </summary>
+        /// <param name="patTokenCreateRequest">
+        /// XXX: Not documented by Visual Studio client libraries. Should we depend on this or
+        /// should this method use the corresponding primitives as arguments? Not yet clear.
```
Haard30, line 36: Given that this is not documented by client libraries, how about we create a separate private method like `BuildPatTokenCreateRequest()` and use that method at all the places we need this object. And then the caller doesn't really have to worry about what's a `patTokenCreateRequest` object and can pass primitives as arguments.

src/AdoPat/PatClient.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// </param>
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
+        /// <returns>A <see cref="Task"/> returning a <see cref="PatTokenResult"/>.</returns>
+        public async Task<PatToken> CreatePatAsync(
+            PatTokenCreateRequest patTokenCreateRequest,
+            CancellationToken cancellationToken = default)
+        {
+            var patTokenResult = await this.client.CreatePatAsync(
+                patTokenCreateRequest,
+                cancellationToken: cancellationToken)
```
Haard30, line 46: Curious about what's the default timeout for all these requests how are we handling timeout/network exceptions

src/AdoPat/PatClient.cs
```diff
@@ -0,0 +1,139 @@
...
+            .ConfigureAwait(false);
+
+            if (patTokenResult.PatToken == null)
+            {
+                throw new PatClientException($"Failed to create PAT during regeneration: {patTokenResult.PatTokenError}");
```
Haard30, line 51: Shall we handle first time PAT creation message differently?

src/AdoPat/PatClient.cs
```diff
@@ -0,0 +1,139 @@
...
+                AllOrgs = AllOrgs,
+            };
+
+            var renewedPatToken = await this.CreatePatAsync(
+                patTokenCreateRequest,
+                cancellationToken: cancellationToken)
+            .ConfigureAwait(false);
+
+            // XXX: If we create a new PAT, but fail to revoke the old PAT, should
+            // we continue or should we try to undo the creation and bail? It's
```
Haard30, line 128: Had some questions to understand the scenario better
1. If the token has been revoked, we would still be able to create a new PAT using it's metadata - is that correct?
2. If the old PAT has been revoked and if we create a new PAT, there can be 2 scenarios if I understand correctly
a. User has already created a new PAT ()with same name) when they revoked previous which we won't know about?
b. There is no active PAT and we just create a new one.
While for scenario `b`, the user would be in good state if we continue, I am curious how we will handle (if at all it's possible) scenario `a`.?
3. How do we undo creation? Will we send another `revoke` request? While this might be addressed later so feel free to not answer - do we clean up the cache?

src/AdoPat/TokensHttpClientWrapper.cs
reillysiemens: Yup. This is a class in another library. 😄

src/AdoPat/PatClient.cs
reillysiemens: @mvanchaa: Does [this documentation](https://learn.microsoft.com/en-us/rest/api/azure/devops/tokens/pats/create?view=azure-devops-rest-7.1&tabs=HTTP#pattokencreaterequest) help make it more clear what `allOrgs` is?

mvanchaa: yes :)

src/AdoPat/PatClient.cs
reillysiemens: The [API documentation for `PatTokenResult`](https://learn.microsoft.com/en-us/rest/api/azure/devops/tokens/pats/create?view=azure-devops-rest-7.1&tabs=HTTP#pattokenresult) does not actually specify (as far as I can tell), but I don't believe that's possible. That is, I don't think it works like our errors where you might have eventual success, but we report every intermediate error that's happened.

src/AdoPat/PatClient.cs
reillysiemens: Uhh... probably we should just leave out the `during regeneration` part here. In regeneration or otherwise it's still just creating a PAT. I think `Failed to create PAT` will be enough.

src/AdoPat/PatClient.cs
reillysiemens: That is an _excellent_ question and the honest answer is that it's pretty obscured by the `Microsoft.VisualStudio.Services.Client` library, so I'll have to do some spelunking before I can give a proper answer... 😞

src/AdoPat/PatClient.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
+        /// <returns>The set of active PATs.</returns>
+        public async Task<ISet<PatToken>> GetActivePatsAsync(CancellationToken cancellationToken = default)
+        {
+            // Initialize a PagedPatTokens so that we can use the continuation token
+            // in the scope of the conditional of the do while loop below. Azure
+            // DevOps will return the empty string when there are no more pages.
+            var pagedPatTokens = new PagedPatTokens(continuationToken: string.Empty, patTokens: new List<PatToken>());
+            var pats = new HashSet<PatToken>();
```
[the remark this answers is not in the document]
reillysiemens: Great question. A `PatToken` has the following fields
```
string DisplayName
DateTime ValidTo
string Scope
IList<Guid> TargetAccounts
DateTime ValidFrom
Guid AuthorizationId
string Token
```
If we leave aside the `AuthorizationId`, because the `Token` is always `null` when returned from the API after the first time it would be possible to have all the other fields be the same and have duplicates. That's quite unlikely, I think, but not impossible.

However, the `AuthorizationId` does exist and it's unique to the PAT even if all the other fields are exactly the same. ❄️ 

So, no, I don't believe there's a possibility of getting duplicate PATs as long as we've got the `AuthorizationId`.

src/AdoPat/PatClient.cs
reillysiemens: The user of this class or of the whole tool? In the case of the former, we'll need to provide a `PatToken` (which we'll get from the cache). In the case of the latter, the end-user will supply an `organization`, `display name`, and `scope` combination that will be used as a key to retrieve an existing token from the cache. Or, if a token is not found, trigger creation of a new `PatToken`.

src/AdoPat/PatClient.cs
reillysiemens: Yes, it is possible that two PATs could have the same `DisplayName`, so this could cause confusion. We could try to alleviate that by adding our own unique prefix/suffix into the `DisplayName`, like when it was generated or something, but IDK how helpful that would actually be.

I posed this question because it's a _possible_ error case, but I don't think it's especially likely. Right now I'm trending towards not creating complicated retry logic or graceful recovery logic in favor of surfacing loud error messages to the user if something does go wrong, but I don't know if everyone will agree with me on that. 😅

src/AdoPat/PatClient.cs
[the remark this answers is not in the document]
reillysiemens: Hmm... thinking on this a bit more, the reason I wanted an `ISet` here was to easily compare a cached PAT with what Azure DevOps thinks the state of the world is, but because of the `null` `Token` field that comparison won't actually be perfect equality. It might be better to return an `IDictionary` keyed by the `AuthorizationId`. What do you think, @mvanchaa?

src/AdoPat/PatClient.cs
reillysiemens: 1. Yes, if a token has been revoked, but we still have its metadata it should be possible to create a new one with the same metadata.
2. For scenario `b`, yes, if there's no active PAT which corresponds to the one that was just requested, we can keep the new one. I'm not sure that I understand scenario `a` here, can you try to explain what you mean? 🤔
3. Yes, unfortunately there's no such thing as deleting a PAT. Creation can only be "undone" by sending another `revoke` request. If we did "undo" I think it would be wise to clean up the cache. In this case we won't necessarily have written the PAT we _just_ created to the cache, but we can remove the old entry (if one exists).
4. I'm not sure. I don't have certainty on the exact frequency, but I have only _observed_ emails being sent upon creation of a new PAT.

src/AdoPat/PatClient.cs
reillysiemens: A `PatTokenCreateRequest` is pretty simple to construct. We could add a private method, but it would be just as easy to create it in this method, I think. It has fields
```
string DisplayName
string Scope
DateTime? ValidTo
bool AllOrgs = false
```
The reason I didn't initially use the primitives is because this seems like a [data clump](https://refactoring.guru/smells/data-clumps) to me where all these parameters are bound up together and highly related.

I considered not leaking implementation details from a library we rely on, but I'm also returning `PatToken` elsewhere, which is another Visual Studio type. In that case and the case of `PatTokenCreateRequest`, however, those are both just data classes with no behavior on them, esentially just containers for data which belong together.

Since we're not publishing this as a C# library I've played a bit fast and loose by re-using external types instead of wrapping them in our own. Basically, what I'm asking here is whether that seems like an acceptable tradeoff.

src/AdoPat/PatClient.cs
reillysiemens: Awesome, I'll include a link or copy/paste that text here to better explain what `allOrgs` is about. 😁

src/AdoPat/PatClient.cs
Haard30: Thanks for answering all those question :)
For scenario `a`, I was wondering that if the user has used UI to regenerate the PAT - that would make old PAT invalid. Now, when we create new a PAT (because old became invalid when user used UI to regenerate), would it interfere with the previously (UI) regenerated PAT which is also active? Or would both of them work independently (meaning we will not bother about the PAT regenerated through UI and just create new one -> store it in cache and remove any invalid)?

I think we would not be bothered about (UI) regenerated PAT and just regenerate again, but just wanted to confirm.

src/AdoPat/PatClient.cs
Haard30: Got it. That makes things much clear. It seems reasonable tradeoff to me :)

src/AdoPat/AdoPat.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <PackageId>Microsoft.Authentication.AdoPat</PackageId>
+    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
```
shalinikhare27, line 5: Why are we not targeting net6.0?

src/AdoPat/ITokensHttpClientWrapper.cs
```diff
@@ -0,0 +1,60 @@
...
+            DisplayFilterOptions? displayFilterOption = null,
+            SortByOptions? sortByOption = null,
+            bool? isSortAscending = null,
+            string continuationToken = null,
+            int? top = null,
+            object userState = null,
+            CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// TODO.
```
shalinikhare27, line 49: Should we just remove this for now till we implement this?

src/AdoPat/PatClient.cs
shalinikhare27, line 18: Can we add a short description for "AllOrgs"?

src/AdoPat/PatClient.cs
shalinikhare27: Is it possible to validate it through tests?

src/AdoPat/PatClient.cs
shalinikhare27, line 51: Are we going to streamline these exceptions to count in the error list of our telemetry? And also log them
`this.logger.LogDebug($"Failed to create PAT during regeneration: {patTokenResult.PatTokenError}")`

src/AdoPat/PatClient.cs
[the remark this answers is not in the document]
shalinikhare27: Are you suggesting we have a key value pair? Key = `AuthorizationId` and Value = '{PatToken}`

src/AdoPat/PatClient.cs
```diff
@@ -0,0 +1,139 @@
...
+                    sortByOption: SortByOptions.DisplayDate,
+                    cancellationToken: cancellationToken)
+                .ConfigureAwait(false);
+
+                foreach (PatToken pat in pagedPatTokens.PatTokens)
+                {
+                    pats.Add(pat);
+                }
+            }
+            while (!string.IsNullOrEmpty(pagedPatTokens.ContinuationToken));
```
shalinikhare27, line 84: Is there a benefit of using `do-while` loop versus a simple `while` loop here?

src/AdoPat/PatClient.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                pagedPatTokens = await this.client.ListPatsAsync(
+                    displayFilterOption: DisplayFilterOptions.Active,
+                    top: PageSize,
+                    continuationToken: pagedPatTokens.ContinuationToken,
+                    sortByOption: SortByOptions.DisplayDate,
+                    cancellationToken: cancellationToken)
+                .ConfigureAwait(false);
+
+                foreach (PatToken pat in pagedPatTokens.PatTokens)
```
shalinikhare27, line 79: We should add a null check here if pagedPatTokens can be null.

shalinikhare27 (commented):
Took a first stab at it.

src/AdoPat/AdoPat.csproj
reillysiemens: That's a great question. I had actually just copied from other files we have in this repo here. @kyle-rader: Any reason we shouldn't just target `net6.0` here?

src/AdoPat/ITokensHttpClientWrapper.cs
reillysiemens: Lol. No, I just forgot to fill this out. I'll add a description here.

src/AdoPat/PatClient.cs
reillysiemens: The documentation link is pretty long, so I just described what the field does in a comment.

src/AdoPat/PatClient.cs
reillysiemens: I'm very confident this is not possible, so I'm going to resolve this, but we can keep an eye out for it as we kick the tires on this tool.

src/AdoPat/PatClient.cs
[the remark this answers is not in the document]
reillysiemens: Yes, @shalinikhare27, that's exactly what I'm suggesting. I think it will make it easier to validate whether a token that we have in the cache is in the active tokens collections. We can compare an existing `AuthorizationId` to the keys for this dictionary.

src/AdoPat/PatClient.cs
reillysiemens: I'm going to mark this as resolved now, but if this response wasn't sufficient, @mvanchaa, feel free to mark this comment as unresolved and we can talk about it.

src/AdoPat/PatClient.cs
reillysiemens: @kyle-rader: Do you have an opinion here?

src/AdoPat/PatClient.cs
reillysiemens: Not without actually making the API calls. And even then it would be difficult, because if there are retries in the client library they're likely to not be exposed to us and just have their own internal retry loops instead.

src/AdoPat/PatClient.cs
reillysiemens: The user using the UI to create a new PAT doesn't necessarily invalidate an older PAT. You can have more than one PAT with the same name and scopes. Unless the user themselves marks it as revoked or the PAT expires it's probably still valid. The UI created PAT would continue to work independently. At least as far as I understand Azure DevOps PATs.

My inclination here is to just ignore PATs that aren't in our cache even if we could discover through the API that they've independently created a PAT which matches the name/scopes we're looking for. I'm especially inclined to do so because we can never get the token value for that PAT, so it's useless to the machine running our CLI.

src/AdoPat/PatClient.cs
reillysiemens: I've done this. 😎

src/AdoPat/PatClient.cs
reillysiemens: Not sure what you mean by streamline, but yeah, as they bubble up, they should probably be counted in our telemetry. I had planned to log them, but at a higher level than this. Probably in the `PatManager` class mentioned in the entity relationship diagram in this PR (or maybe even in `Main`).

src/AdoPat/PatClient.cs
reillysiemens: Hmm... I think I tried to use a `do...while` here, but couldn't make it work the way I wanted. If you can demonstrate that working here I'd be happy to switch to more terse code, @shalinikhare27! ❤️

src/AdoPat/PatClient.cs
reillysiemens: I really hope that's not the case, but I'll check their docs again to be sure. I feel like it would be silly if it could be `null` and they could just provide an empty list instead...

src/AdoPat/AdoPat.csproj
kyle-rader: This is a fantastic question! Dotnet run-times are implementations of Dotnet standards (until you reach dotnet 5). A Library wanting to maximize compatibility, should always target a dotnet standard version (2.0 or 2.1), while an executable must choose a runtime version. [Dotnet Standard](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-1#net-5-and-net-standard) has a ton of great info here. From reading more on this just now since I hadn't read that page in a little while,

Dotnet standard 2.1 is supported by Dotnet 5+ and will continue to be. This means we can always use libraries still targeting that standard. But, As of dotnet 5 - the dotnet platform is consolidated (no core vs framework) and so within our own app - we can target net 5/6 directly in our libraries, and while that makes them only usable by net6 applications, this is fine, because it's only auzreauth using them. This opens up the door to use using C# 9!

So @reillysiemens I do think we should go ahead and target net6 directly here.

src/AdoPat/PatClient.cs
reillysiemens: Oh! It's not possible because **we** create the `pagedPatTokens` above with
```csharp
var pagedPatTokens = new PagedPatTokens(continuationToken: string.Empty, patTokens: new List<PatToken>());
```
I don't think there's a need for a `null` check here, @shalinikhare27.

src/AdoPat.Test/PatClientTest.cs
```diff
@@ -0,0 +1,222 @@
...
+            var patTokenResult = new PatTokenResult
+            {
+                PatToken = patToken,
+                PatTokenError = SessionTokenError.None,
+            };
+
+            var client = new Mock<ITokensHttpClientWrapper>(MockBehavior.Strict);
+            client.Setup(c => c.CreatePatAsync(
+                It.IsAny<PatTokenCreateRequest>(),
+                It.IsAny<object>(),
```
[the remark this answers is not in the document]
reillysiemens: Ask the Azure DevOps team. 😆 The signature of `CreatePatAsync` takes `object userState = null`. I'm just following what they specified. 😅

src/AdoPat/PatClient.cs
reillysiemens: Hmm... just kidding. It's still _technically_ possible because we reassign `pagedPatTokens`.

reillysiemens: OK. I went spelunking **waaay** deep into the implementation details of the Visual Studio client code we're using here and found that the underlying methods are generic and do
```csharp
return default(T);
```
I don't know enough about C# yet, but I don't _think_ that usually returns `null`. Do either of you know, @shalinikhare27 or @kyle-rader?

reillysiemens: Ugh. That's **also** not entirely correct. `default(T)` definitely could be `null` depending on the type, but there are also earlier returns in that same function that make use of `T`.

reillysiemens: @Haard30 and I dove pretty deep into the code for the library we're using here and it seems pretty unlikely that we'll get a `null` here without getting some other kind of exception first, so I'm inclined to not do the `null` checking here. Opinions, @shalinikhare27 &amp; @kyle-rader?

reillysiemens: @shalinikhare27 and I talked through this as well and I think we're in agreement that it's much more likely that some exception will be raised before we ever get a `null` for either `pagedPatTokens` or `pagedPatTokens.PatTokens`.

In any event, even if we did get a `null`, that's not exactly a recoverable situation and it means something else has gone wrong. My plan is to handle exceptions further up the stack in `PatManager`, so I think it's OK to let a `NullReferenceException` bubble up to there. Would still like to get your thoughts, @kyle-rader.

src/AdoPat/AdoPat.csproj
reillysiemens: Done. This now targets net6.

src/AdoPat/ITokensHttpClientWrapper.cs
reillysiemens: Done.

src/AdoPat/TokensHttpClientWrapper.cs
```diff
@@ -0,0 +1,76 @@
...
+            .ConfigureAwait(false);
+        }
+
+        /// <inheritdoc/>
+        public async Task RevokeAsync(
+            Guid authorizationId,
+            object userState = null,
+            CancellationToken cancellationToken = default)
+        {
+            await this.RevokeAsync(
```
shalinikhare27: The cause of all issues. Good catch, @reillysiemens ;)

src/AdoPat/PatClient.cs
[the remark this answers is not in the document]
reillysiemens: I made this method return an `IDictionary<Guid, PatToken>` instead of an `ISet<PatToken>`, which I think will work better in the long run.

src/AdoPat/PatClient.cs
```diff
@@ -0,0 +1,143 @@
...
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.VisualStudio.Services.DelegatedAuthorization;
+
+    /// <summary>
+    /// An abstraction over common operations with the Azure DevOps PAT Lifecycle Management REST API.
+    /// </summary>
+    public class PatClient
+    {
+        private const int PageSize = 100; // Using the maximum allowable page size allows us to reduce HTTP calls.
```
shalinikhare27, line 17: This might be part of another PR, but we should consider moving all constants to `Constants.cs` for consistentancy.

src/AdoPat/PatClient.cs
```diff
@@ -0,0 +1,143 @@
...
+        public async Task<IDictionary<Guid, PatToken>> GetActivePatsAsync(CancellationToken cancellationToken = default)
+        {
+            // Initialize a PagedPatTokens so that we can use the continuation token
+            // in the scope of the conditional of the do while loop below. Azure
+            // DevOps will return the empty string when there are no more pages.
+            var pagedPatTokens = new PagedPatTokens(continuationToken: string.Empty, patTokens: new List<PatToken>());
+            var pats = new Dictionary<Guid, PatToken>();
+            do
+            {
+                pagedPatTokens = await this.client.ListPatsAsync(
```
shalinikhare27, line 75: Did we discuss the null check here with Kyle? If not, we can move forward for now since we will not be shipping this until the PatManager is ready!

src/AdoPat/PatClient.cs
```diff
@@ -0,0 +1,143 @@
...
+                ValidTo = validTo,
+                AllOrgs = AllOrgs,
+            };
+
+            var renewedPatToken = await this.CreatePatAsync(
+                patTokenCreateRequest,
+                cancellationToken: cancellationToken)
+            .ConfigureAwait(false);
+
+            // XXX: If we create a new PAT, but fail to revoke the old PAT, should
```
shalinikhare27, line 131: This seems like a design discussion and should be removed from code. :)

shalinikhare27 (commented):
second look.

src/AdoPat/ITokensHttpClientWrapper.cs
```diff
@@ -0,0 +1,60 @@
...
+            CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// [Preview API] Revokes a personal access token (PAT) by authorizationId.
+        /// </summary>
+        /// <param name="authorizationId">The authorizationId identifying a single, unique personal access token (PAT).</param>
+        /// <param name="userState">(Undocumented).</param>
+        /// <param name="cancellationToken">The cancellation token to cancel operation.</param>
+        /// <returns>Void.</returns>
+        Task RevokeAsync(
```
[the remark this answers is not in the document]
reillysiemens: This is consistent, but with the underlying API that it wraps. The library method is called `RevokeAsync` even though I wish it was called `RevokePatAsync`. ☹️ To avoid confusion with the API that this wraps I'd prefer to leave this one as is.

src/AdoPat/PatClient.cs
reillysiemens: `Constant.cs` is part of the MSALWrapper, which is another namespace. I'd prefer not to mix namespaces like that. We could move the constants into a shared namespace, but I'd also avoid doing that because it reduces the locality of information. That is, you have to go further from the point of use to find out what a value is. This constant isn't used anywhere outside of this file (currently), so I'd prefer to leave it where it is for the foreseeable future.

src/AdoPat/PatClient.cs
reillysiemens: @kyle-rader never did weigh in on that, but you, me, and @Haard30 reached something of a consensus I think. Let's roll with that for now with the knowledge that we can question it again before we ship the `PatManager` class.

Haard30 (dismissed):
Looks very good to me.

mvanchaa (dismissed):

src/AdoPat/PatClientException.cs
```diff
@@ -0,0 +1,19 @@
+namespace Microsoft.Authentication.AdoPat
```
goagain, line 1: [nit]Microsoft Copyright is missing.😊

goagain (approved):
This PR is good!

src/AdoPat/PatClientException.cs
reillysiemens: Not a nit! That's important!

goagain (approved):

mvanchaa (approved):

Haard30 (approved):

PR #177, opened 2022-12-12
Pass template parameters when triggering a build on ADO

mvanchaa:
The ADO pipeline to generate Linux binaries requires inputs like `version`, `commitSHA`. So, updated the script to provide yaml template parameters when triggering a build.

In this PR, we use pipeline APIs instead of build APIs because only the run pipeline API allows us to pass template parameters and build API doesn't support it.
- run pipeline API: https://learn.microsoft.com/en-us/rest/api/azure/devops/pipelines/runs/run-pipeline?view=azure-devops-rest-6.0 
- queue build API: https://learn.microsoft.com/en-us/rest/api/azure/devops/build/builds/queue?view=azure-devops-rest-6.0.


bin/trigger_azure_pipelines.py
```diff
@@ -74,7 +78,8 @@ def main() -> None:
         organization = os.environ["ADO_ORGANIZATION"]
         project = os.environ["ADO_PROJECT"]
         pipeline_id = os.environ["ADO_AZUREAUTH_LINUX_PIPELINE_ID"]
-
+        version = os.environ["VERSION"]
+        commitSHA = os.environ["commitSHA"]
```
mvanchaa, line 82: `$GITHUB_SHA` has the same information. You can use it directly. https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

bin/trigger_azure_pipelines.py
```diff
@@ -26,43 +25,48 @@ def ado_connection(organization: str, ado_pat: str) -> Connection:
...
+    run = pipeline_client.get_run(project, pipeline_id, run_id)
 
     # polling interval is set in accordance with the rate limits specified here:
     # https://learn.microsoft.com/en-us/azure/devops/integrate/concepts/rate-limits?view=azure-devops
     polling_interval_seconds = 30
 
     # Wait until the build have a complete status.
-    while build.result not in COMPLETED_STATUSES:
+    # https://learn.microsoft.com/en-us/rest/api/azure/devops/pipelines/runs/get?view=azure-devops-rest-6.0#runstate
+    while run.state != "completed":
```
mvanchaa, line 38: put this in a variable.

bin/trigger_azure_pipelines.py
```diff
@@ -26,43 +25,48 @@ def ado_connection(organization: str, ado_pat: str) -> Connection:
...
-    return build
+        run = pipeline_client.get_run(project, pipeline_id, run_id)
+    return run
 
 
 def trigger_azure_pipeline_and_wait_until_its_completed(
     organization: str,
     project: str,
-    pipeline_id: str,
+    pipeline_id: int,
```
mvanchaa, line 47: Keep it as a `str`

bin/trigger_azure_pipelines.py
```diff
@@ -26,43 +25,48 @@ def ado_connection(organization: str, ado_pat: str) -> Connection:
...
     ado_pat: str,
+    version: str,
+    commitSHA: str
```
mvanchaa, line 50: `commit_sha`

mvanchaa: or `commit` or `commit_hash`

reillysiemens (approved):

reillysiemens (approved):

bin/trigger_azure_pipelines.py
```diff
@@ -74,7 +78,8 @@ def main() -> None:
         organization = os.environ["ADO_ORGANIZATION"]
         project = os.environ["ADO_PROJECT"]
         pipeline_id = os.environ["ADO_AZUREAUTH_LINUX_PIPELINE_ID"]
-
+        version = os.environ["VERSION"]
```
kyle-rader, line 81: Are the env var names `VERSION` and `GITHUB_SHA` ones we control through Github secrets? I'm wondering if we can make these names slightly more descriptive, because on their own, they could mean a lot of different things.

kyle-rader (approved):
Approved with a question :)

bin/trigger_azure_pipelines.py
mvanchaa: `GITHUB_SHA` environment variable is available by default for workflows. https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables 

`VERSION` is the release version. We need to set this as an environment variable in the release workflow. We get this value currently from `${{ github.event.inputs.version }}`

mvanchaa: How about updating `VERSION` to `RELEASE_VERSION`?

reillysiemens: I figured `VERSION` was implied because it's a GitHub release, but if it's confusing we can give it more context. Perhaps `AZUREAUTH_VERSION`? When I've created releases via the CLI (which uses the release workflow) I've run
```
gh workflow run Release --field version=<whatever> --field prerelease=true
```
, but I think it makes just as much sense to do
```
gh workflow run Release --field azureauth_version=<whatever> --field prerelease=true
```

mvanchaa: @reillysiemens when you try
```
gh workflow run Release --field version=<whatever> --field prerelease=true
```
does `${{ github.event.inputs.version }}` holds the <whatever> ?

bin/trigger_azure_pipelines.py
```diff
@@ -8,14 +8,13 @@
...
-from azure.devops.v5_1.build.models import Build
+from azure.devops.v6_0.pipelines.pipelines_client import PipelinesClient
+from azure.devops.v6_0.pipelines.models import Run
 from msrest.authentication import BasicAuthentication
 
-# https://learn.microsoft.com/en-us/rest/api/azure/devops/build/builds/get?view=azure-devops-rest-7.1#buildresult
-FAILED_STATUSES: set[str] = {"canceled", "partiallySucceeded", "failed"}
-COMPLETED_STATUSES: set[str] = FAILED_STATUSES | {"succeeded"}
-
+# https://learn.microsoft.com/en-us/rest/api/azure/devops/pipelines/runs/get?view=azure-devops-rest-6.0#runresult
```
goagain, line 15: downgraded version?

goagain (approved):

bin/trigger_azure_pipelines.py
```diff
@@ -8,14 +8,13 @@
 import time
 
 from azure.devops.connection import Connection
-from azure.devops.released.build.build_client import BuildClient
-from azure.devops.v5_1.build.models import Build
+from azure.devops.v6_0.pipelines.pipelines_client import PipelinesClient
+from azure.devops.v6_0.pipelines.models import Run
```
shalinikhare27, line 12: Any reason we run over build?

bin/trigger_azure_pipelines.py
```diff
@@ -26,43 +25,48 @@ def ado_connection(organization: str, ado_pat: str) -> Connection:
...
     """Triggers an azure pipeline and waits for it to be finished"""
-    build_client = ado_connection(organization, ado_pat).clients.get_build_client()
-    build_metadata = {"definition": {"id": pipeline_id}}
-
-    triggered_build = build_client.queue_build(build_metadata, project)
-    build_url = f"https://dev.azure.com/{organization}/{project}/_build/results?buildId={triggered_build.id}&view=results"
+    ado_client = ado_connection(organization, ado_pat).clients_v6_0
+    pipeline_client = ado_client.get_pipelines_client()
+    
+    # NOTE: We run a pipeline instead of queuing a build because only the run pipeline API allows us to pass template parameters and build API doesn't support it.
```
shalinikhare27, line 56: Oh never mind, you answered here.

shalinikhare27 (approved):

bin/trigger_azure_pipelines.py
kyle-rader: Ah that makes more sense, I am fine leaving it as `VERSION` - because we won't have an arbitrary `VERSION` set outside that context of the input field. Coming from the release workflow input field makes sense. Thank you for clarifying!

PR #179, opened 2022-12-19
Correct a mistake in app configuration docs

reillysiemens:
Correct a mistake in app configuration docs. Previously we said to add localhost under Web redirect URIs, but that was incorrect.


kyle-rader (approved):
Thanks!

Haard30 (approved):

PR #180, opened 2022-12-20
changes to readme telemetry collection

kajames-msft:
Made changes to readme (telemetry collection)


kyle-rader (approved):
I think this is a good first step, but Kayley I'd also like us to talk with an OSS champ to see how much we can customize this message. Because technically speaking, no even if you "turn on" data collection - it will only be sent to application Insights instance identified by the GUID you set. So a company can collect their own telemetry, we will get nothing. And I would really like to make that well known.

Haard30 (approved):

PR #182, opened 2023-01-07
Add sid in telemetry

goagain:
Sid in going to replace current alias in the future.

### DIFF
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -475,7 +475,7 @@ Allowed values: [all, web, devicecode]";
 
                 var tokenResult = successfulResult.TokenResult;
                 this.eventData.Add("silent", tokenResult.IsSilent);
-                this.eventData.Add("sid", tokenResult.Sid);
+                this.eventData.Add("sid", tokenResult.SID);
                 this.eventData.Add("succeeded_mode", successfulResult.AuthFlowName);
 
                 switch (this.Output)
--- a/src/AzureAuth/CommandMain.cs
+++ b/src/AzureAuth/CommandMain.cs
@@ -475,6 +475,7 @@ Allowed values: [all, web, devicecode]";
 
                 var tokenResult = successfulResult.TokenResult;
                 this.eventData.Add("silent", tokenResult.IsSilent);
+                this.eventData.Add("sid", tokenResult.Sid);
                 this.eventData.Add("succeeded_mode", successfulResult.AuthFlowName);
 
                 switch (this.Output)
--- a/src/MSALWrapper.Test/TokenResultTest.cs
+++ b/src/MSALWrapper.Test/TokenResultTest.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
 
     public class TokenResultTest
     {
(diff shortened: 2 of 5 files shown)


src/MSALWrapper.Test/TokenResultTest.cs
```diff
@@ -30,14 +30,14 @@ public void ConstructorTakesJWT()
         {
             this.subject.JWT.Should().Be(this.jwt);
             this.subject.Token.Should().Be(FakeToken);
-            this.subject.User.Should().Be("kyrader@microsoft.com");
+            this.subject.User.Should().Be("johndoe@consto.com");
```
shalinikhare27, line 33:
```suggestion
            this.subject.User.Should().Be("johndoe@contoso.com");

```

super nit.

src/MSALWrapper.Test/TokenResultTest.cs
```diff
@@ -30,14 +30,14 @@ public void ConstructorTakesJWT()
...
-            this.subject.User.Should().Be("kyrader@microsoft.com");
+            this.subject.User.Should().Be("johndoe@consto.com");
             this.subject.ValidFor.Should().NotBeCloseTo(new System.TimeSpan(0, 0, 0), new System.TimeSpan(0, 1, 0));
         }
 
         [Test]
         public void TokenResult_ToString()
         {
-            this.subject.ToString().Should().Be("Token cache warm for kyrader@microsoft.com (John Doe)");
+            this.subject.ToString().Should().Be("Token cache warm for johndoe@consto.com (John Doe)");
```
shalinikhare27, line 40:
```suggestion
            this.subject.ToString().Should().Be("Token cache warm for johndoe@contoso.com (John Doe)");

```

src/MSALWrapper.Test/TokenResultTest.cs
```diff
@@ -50,7 +50,7 @@ public void TokenResult_DisplayName()
         public void TokenResult_ToJson()
         {
             var expected = $@"{{
-    ""user"": ""kyrader@microsoft.com"",
+    ""user"": ""johndoe@consto.com"",
```
shalinikhare27, line 53:
```suggestion
    ""user"": ""johndoe@contoso.com"",

```

src/MSALWrapper/JsonWebTokenExtensions.cs
```diff
@@ -50,5 +50,16 @@ public static string GetDisplayName(this JsonWebToken jwt)
             jwt.TryGetClaim("name", out Claim name);
             return name?.Value;
         }
+
+        /// <summary>
+        /// Get a user's Sid.
+        /// </summary>
+        /// <param name="jwt">The jwt.</param>
```
shalinikhare27, line 57:
```suggestion
        /// <param name="jwt">The JSON Web Token.</param>

```

src/MSALWrapper/TokenResult.cs
```diff
@@ -70,6 +71,11 @@ public JsonWebToken JWT
         /// </summary>
         public bool IsSilent { get; internal set; }
 
+        /// <summary>
+        /// Gets the user's Sid.
```
shalinikhare27, line 75:
```suggestion
        /// Gets the user's security identifier.

```

src/AzureAuth/CommandMain.cs
```diff
@@ -475,6 +475,7 @@ private int GetToken()
 
                 var tokenResult = successfulResult.TokenResult;
                 this.eventData.Add("silent", tokenResult.IsSilent);
+                this.eventData.Add("sid", tokenResult.Sid);
```
shalinikhare27, line 478: How do we handle when sid is null?

shalinikhare27 (commented):
Some initial comments.

src/AzureAuth/CommandMain.cs
goagain: If it is null, it will be blank in telemetry, like `option_resource`. You can compare it with an empty string.

reillysiemens: Have we ever observed `Sid` to actually _be_ `null`? Is that a possible value? 🤔

reillysiemens: Lol. Nevermind, I think I got my answer below. 😅

goagain: SID was used in SourceDepot (according to Kyle). If it is null, SD won't work. 
So, I guess it cannot be null.

reillysiemens (dismissed):

shalinikhare27 (dismissed):

mvanchaa (approved):

src/MSALWrapper/JsonWebTokenExtensions.cs
```diff
@@ -50,5 +50,16 @@ public static string GetDisplayName(this JsonWebToken jwt)
             jwt.TryGetClaim("name", out Claim name);
             return name?.Value;
         }
+
+        /// <summary>
+        /// Get a user's Sid.
+        /// </summary>
+        /// <param name="jwt">The JSON web token.</param>
+        /// <returns>The <see cref="string"/>.</returns>
+        public static string GetUserSid(this JsonWebToken jwt)
```
Haard30, line 59: Is Sid something tied to a user. If so, can we name this method as just `GetSid` like we have for `GetDisplayName`? Feel free to ignore if you feel `GetUserSid` gives more context :)

kyle-rader: I think `User` in this context is okay to have - not all JWTs are going to be attached to a User, if User is defined as a human. I would support naming this `GetSID()` (only using Caps, if we also name the property `SID`).

src/MSALWrapper/TokenResult.cs
```diff
@@ -70,6 +71,11 @@ public JsonWebToken JWT
         /// </summary>
         public bool IsSilent { get; internal set; }
 
+        /// <summary>
+        /// Gets the user's security identifier.
+        /// </summary>
+        public string Sid { get; internal set; }
```
kyle-rader, line 77: I might consider keeping this name as `SID` (all caps) because it's an acronyum not a word. I do not feel strongly about this though.

kyle-rader (approved):
Approved with 2 comments to consider :)

Haard30 (approved):

mvanchaa (approved):

PR #183, opened 2023-01-10
Add unlit PAT caching classes

reillysiemens:
This PR adds a `PatCache` class to the `Microsoft.Authentication.AdoPat` namespace. It also adds supporting interfaces and classes, namely

- a `StorageWrapper` class, which wraps `Storage` from [`Microsoft.Identity.Client.Extensions.Msal` 2.25.3](https://www.nuget.org/packages/Microsoft.Identity.Client.Extensions.Msal/)
- a corresponding `IStorageWrapper` interface to make testing easier
- tests in the `Microsoft.Authentication.AdoPat.Test` namespace

There will need to be a handful of follow up PRs to fully implement this feature. I consider this phase 2 of 4. Phase 1 is covered by #176, so until that merges you'll see some overlap in commits. Until all of the phases are complete this code won't be "lit" or "live" in the application and should not factor into production builds.

This entity relationship diagram shows the rough layout of parts needed to make this work. This PR covers `PatCache` and `StorageWrapper`, so there's still work to be done for the other pieces of the puzzle.

```mermaid
erDiagram
    CLI ||..|| PatManager : uses
    PatCache }|..|{ StorageWrapper : uses
```

I've marked this PR as a draft for now because I'd like to get feedback while I work on the other phases. If it's well received, however, it should be safe to merge it, and I can lift the draft status. There might be some questions marked as `XXX` or `TODO` that I'm particularly interested in getting feedback on, so keep an eye out for those.


reillysiemens:
@jcreamer898, hopefully this is also an exciting PR. 😆

mvanchaa (commented):
Great work, Tucker!🎉

src/AdoPat/ITokensHttpClientWrapper.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    /// <summary>
+    /// An interface for a transparent wrapper of <see cref="TokensHttpClient"/>.
+    /// </summary>
+    public interface ITokensHttpClientWrapper
+    {
+        /// <summary>
+        /// [Preview API] Creates a new personal access token (PAT) for the requesting user.
+        /// </summary>
+        /// <param name="patTokenCreateRequest">(Undocumented).</param>
```
shalinikhare27, line 19: How are we going to handle the "undocumented" stuff. Make our best assumption or reach out to someone who knows about these APIs?

src/AdoPat/PatCache.cs
```diff
@@ -0,0 +1,81 @@
...
+    using System.Text.Json;
+    using Microsoft.VisualStudio.Services.DelegatedAuthorization;
+
+    /// <summary>
+    /// A simple cache mapping <see cref="string"/> to <see cref="PatToken"/> backed by secure storage.
+    /// </summary>
+    public class PatCache
+    {
+        private IStorageWrapper storage;
+        private Lazy<Dictionary<string, PatToken>> cache;
```
shalinikhare27, line 18: What's the purpose of using Lazy Dictionary here?

shalinikhare27 (commented):
First stab at this one as well.

src/AdoPat/PatCache.cs
kyle-rader: Using `Lazy<T>` here is a clean way to provide memoization of the cache itself, and has the benefit of moving the computation to read it to first use, instead of when creating the PatCache instance.

src/AdoPat/PatCache.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// <param name="storage">A class which implements <see cref="IStorageWrapper"/>.</param>
+        public PatCache(IStorageWrapper storage)
+        {
+            // Note: This assumes any required persistance validity has been
+            // checked beforehand.
+            this.storage = storage;
+
+            // The cache is lazy so that the initial cache read from the
+            // underlying storage only happens once and subsequent reads
+            // operate on the in-memory value.
```
kyle-rader, line 32: This comment may be more suited to the first use of lazy, so a reader finds it above first.

src/AdoPat/IStorageWrapper.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    /// <summary>
+    /// An interface for a thin, testable wrapper around MSAL's Storage class.
+    /// </summary>
+    public interface IStorageWrapper
+    {
+        /// <summary>
+        /// Read and unprotect cache data.
+        /// </summary>
+        /// <returns>Unprotected cache data.</returns>
```
Haard30, line 14: Curious why we call these data unprotected and protected

src/AdoPat.Test/PatCacheTest.cs
```diff
@@ -0,0 +1,251 @@
...
+            var expectedPat = pat;
+            expectedPat.Token = "New Token Value";
+
+            // First, write existing data as bytes.
+            var pats = new Dictionary<string, PatToken> { { key, pat } };
+            var existing_data = JsonSerializer.SerializeToUtf8Bytes(pats);
+
+            // Then, overwrite the existing data with the expected token.
+            pats[key] = expectedPat;
+            var expected_data = JsonSerializer.SerializeToUtf8Bytes(pats);
```
Haard30, line 232: have you named it `expected_data` instead of `expectedData` because of Serialization stuff?

src/AdoPat/IStorageWrapper.cs
reillysiemens: Just copying what the underlying library refers to them as. The data is usually encrypted by the MSAL cache helpers, so I think that's why this language is used.

src/AdoPat.Test/PatCacheTest.cs
reillysiemens: Nooope, just forgot which language I was writing. 😆

src/AdoPat.Test/PatCacheTest.cs
```diff
@@ -0,0 +1,251 @@
...
+                Token = Token,
+            };
+
+            // Make our expected PatToken differ only by the Token value.
+            var expectedPat = pat;
+            expectedPat.Token = "New Token Value";
+
+            // First, write existing data as bytes.
+            var pats = new Dictionary<string, PatToken> { { key, pat } };
+            var existing_data = JsonSerializer.SerializeToUtf8Bytes(pats);
```
Haard30, line 228:
```suggestion
            var existingData = JsonSerializer.SerializeToUtf8Bytes(pats);

```

src/AdoPat.Test/PatCacheTest.cs
```diff
@@ -0,0 +1,251 @@
...
+
+            // Then, overwrite the existing data with the expected token.
+            pats[key] = expectedPat;
+            var expectedData = JsonSerializer.SerializeToUtf8Bytes(pats);
+
+            var data = new byte[] { };
+            var storage = new Mock<IStorageWrapper>(MockBehavior.Strict);
+            storage.Setup(s => s.ReadData()).Returns(existing_data);
```
Haard30, line 236:
```suggestion
            storage.Setup(s => s.ReadData()).Returns(existingData);

```

src/AdoPat/ITokensHttpClientWrapper.cs
reillysiemens: We can make a best assumption here, I think. I put `(Undocumented).` because StyleCop wanted *something*, but the source code for the library just has
```csharp
/// <param name="patTokenCreateRequest"></param>
```
but I also think it's largely self-explanatory what these undocumented things are.

src/AdoPat/PatCache.cs
reillysiemens: @kyle-rader is exactly right. I just used `Lazy<T>` as a way of avoiding a read of the storage in the constructor for this class as I generally don't like to see I/O in constructors.

src/AdoPat/ITokensHttpClientWrapper.cs
reillysiemens: @shalinikhare27: I took a stab at documenting these items while still being honest about the fact that the underlying library says nothing about them.

Haard30 (approved):
Looks awesome to me! ⭐

shalinikhare27 (approved):

mvanchaa (approved):

PR #185, opened 2023-01-24
Collect env vars for Azure pipelines and Cloud Build Agent machines

shalinikhare27:
The telemetry events coming from Azure pipelines and Cloud Build Agent machines skew up AzureAuth telemetry. In this PR, I am adding environment variables to be set as a part of telemetry config. 

1. SYSTEM_DEFINITIONID will be set if a telemetry event is coming from Azure pipeline.
2. QBUILD_DISTRIBUTED is set if a telemetry event is coming from Cloud Build Agent machines.

Test: 
`.\bin\azureauth.cmd --verbosity trace`
The env vars get added to trace as:
`"env_system_definitionid": "",
 "env_qbuild_distributed": ""`

If these are set then we can exclude them from our aggregated data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 ### Added
-- Set the environment variable for Azure pipeline and Cloud Build Agent machine to avoid skewing aggregated telemetry.
+- Collect environment variables in telemetry distinguishing Azure Pipelines and Cloud Build.
 
 ### Added
 - Added On-premises Security Identifier in Telemetry as `sid`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 ### Added
+- Set the environment variable for Azure pipeline and Cloud Build Agent machine to avoid skewing aggregated telemetry.
+
+### Added
 - Added On-premises Security Identifier in Telemetry as `sid`.
 
 ### Changed
--- a/src/AzureAuth/Program.cs
+++ b/src/AzureAuth/Program.cs
@@ -44,12 +44,12 @@ namespace Microsoft.Authentication.AzureAuth
                 backend = TelemetryOutput.ApplicationInsights;
             }
 
-            TelemetryConfig telemetryConfig = new TelemetryConfig()
-            {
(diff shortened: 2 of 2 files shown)


goagain (approved):

kyle-rader (changes requested):
Code change is good, but also update the CHANGELOG :)

src/AzureAuth/Program.cs
```diff
@@ -44,12 +44,12 @@ private static void Main(string[] args)
...
-                Backend = backend,
-                IngestionToken = ingestionToken,
-                Async = true,
-            };
+            TelemetryConfig telemetryConfig = new TelemetryConfig(
+                eventNamespace: "azureauth",
+                backend: backend,
+                ingestionToken: ingestionToken,
+                useAsync: true,
+                envVarsToCollect: new[] { "SYSTEM_DEFINITIONID", "QBUILD_DISTRIBUTED" });
```
mvanchaa, line 52: Just curious if we can add these to Lasso? Because this seems like office specific terminology to me.

shalinikhare27: Lasso gives us the functionality to add env vars in CLI that uses Lasso. Also, did not understand the latter comment. :)

goagain (dismissed):

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 # Unreleased
+### Added
+- Set the environment variable for Azure pipeline and Cloud Build Agent machine to avoid skewing aggregated telemetry.
```
kyle-rader, line 9: Consider "- Collect environment variables in telemetry distinguishing Azure Pipelines and Cloud Build"

"Set the env..." could be read the wrong way that we are setting env vars.

The Changelog need not explain that deeply about why we are doing something, just the What, so that a developer can quickly see what's been done and decide for themselves if that change impacts them.

src/AzureAuth/Program.cs
kyle-rader: These are Microsoft specific, not Office specific. While it may be slightly less desirable to have them in someone else's telemetry (if they choose to set that up for themselves), we don't have a good alternative right now. Yes this could go into Lasso - but that would apply to every lasso app (which may or may not be desirable).

kyle-rader (approved):

goagain (approved):

PR #186, opened 2023-01-24
Upgrade to C# 10.0

kyle-rader:
There's nothing stopping us from using net6 across all our projects now which lets us update to C# 10.0. It's not always clear which version of the language you are using so I've explicitly set it using the `Directory.Build.props` file. See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version for reference.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,5 @@
+<Project>
+    <PropertyGroup>
+        <LangVersion>10.0</LangVersion>
+    </PropertyGroup>
+</Project>
--- a/src/MSALWrapper/MSALWrapper.csproj
+++ b/src/MSALWrapper/MSALWrapper.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <!-- Package Naming, Building, & Versioning -->
     <PackageId>Microsoft.Authentication.MSALWrapper</PackageId>
-    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
+    <TargetFramework>net6.0</TargetFramework>
     <RootNamespace>Microsoft.Authentication.MSALWrapper</RootNamespace>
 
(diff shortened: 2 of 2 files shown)


shalinikhare27 (dismissed):

goagain:
Do we need to add this improvement to changelog?

goagain (approved):

reillysiemens (approved):

kyle-rader:
> Do we need to add this improvement to changelog?

No, because the functionality of AzureAuth is not changing. The ChangeLog is for consumers to quickly see changes in behavior. Developers can use the git history to see what happened in-between :). Good question though!

reillysiemens:
I'm not sure that I agree the `CHANGELOG.md` is _only_ for consumers to see changes in _behavior_. I think it's there to document what has changed. This is a big change even though it changes nothing for the user. I'm not the type to go putting every commit into a `CHANGELOG.md`, but something as significant as a change in framework/language version seems relevant to me.

reillysiemens (approved):

goagain (approved):

kyle-rader:
> I'm not sure that I agree the `CHANGELOG.md` is _only_ for consumers to see changes in _behavior_. I think it's there to document what has changed. This is a big change even though it changes nothing for the user. I'm not the type to go putting every commit into a `CHANGELOG.md`, but something as significant as a change in framework/language version seems relevant to me.

Happy to discuss more if you want to. I have medium strength opinions on this :). I think the `CHANGELOG`'s value is in giving consumers and users a quick way to see what they should be aware of in each new version. Moving to C# 10 is equivalent to an implementation detail or refactor with no change they would care about. Therefore, it would only clutter the CHANGELOG and reduce it's value. If anyone _wants_ to know the implementation details, they can read the git log :).

PR #187, opened 2023-01-27
Upgrade lasso and use incognito mode.

goagain:
To achieve privacy compliance, AzureAuth need to stop collecting EUII data, such as alias, device name.
This PR will use the functionality that Lasso provides.

To make sure we are still able to support our customer, subcommand `info` is added. 
Run `azureauth info` will get the following output:
```
AzureAuth Version: 1.0.0.0
Device ID: 13e6409f-d18d-4e67-9644-b28efe9fc607
To reset your device identifier, Run `azureauth info reset-device-id`
```

Also, the helper text is added in help manual `azureauth --help`:
```
Commands:
  info           Show debug information of AzureAuth. Please provide when asking for help.
```


src/AzureAuth/CommandInfo.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            Assembly assembly = Assembly.GetExecutingAssembly();
+
+            string azureauthVersion = assembly.GetName().Version.ToString();
+            logger.LogInformation($"AzureAuth Version: {azureauthVersion}");
+
+            string deviceID = TelemetryMachineIDHelper.GetRandomDeviceIDAsync(fileSystem).Result;
+            logger.LogInformation($"Device ID: {deviceID}");
+
+            logger.LogInformation($"Device Identifier File Location: {TelemetryMachineIDHelper.GetIdentifierLocation(fileSystem)}");
```
reillysiemens, line 34: For the sake of consistency, I'd refer to this as `Device ID` as we did above. I would also replaced `File Location` with the shorter `Path`.

Compare
```
AzureAuth Version: 0.7.0.0
Device ID: fd5fad7a-1292-****-****-ba092aea1a4a
Device Identifier File Location: C:\Users\********\.lasso\azureauth.json
```
with
```
AzureAuth Version: 0.7.0.0
Device ID: fd5fad7a-1292-****-****-ba092aea1a4a
Device ID Path: C:\Users\********\.lasso\azureauth.json
```

src/AzureAuth/CommandInfo.cs
```diff
@@ -0,0 +1,42 @@
...
+            logger.LogInformation($"To reset your device identifier, delete the file at {TelemetryMachineIDHelper.GetIdentifierLocation(fileSystem)}.");
```
reillysiemens, line 35: Did we talk about making it easier for a user to do this without caring about the file location? Like maybe a
```
azureauth info --reset-device-id
```
or something?

src/AzureAuth/CommandInfo.cs
```diff
@@ -0,0 +1,42 @@
...
+
+            logger.LogInformation($"\nTo get the user's sid, use option --output=sid in normal authentication process.");
```
reillysiemens, line 37: This sentence could be a little more clear, I think. You might consider just telling them _exactly_ which command to run. In this case I think that's
```
azureauth --client <client> --resource <resource> --tenant <tenant> --output sid
```

src/AzureAuth/CommandInfo.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <returns>The error code: 0 is normal execution, and the rest means errors during execution.</returns>
+        public int OnExecute(ILogger<CommandInfo> logger, IFileSystem fileSystem)
+        {
+            Assembly assembly = Assembly.GetExecutingAssembly();
+
+            string azureauthVersion = assembly.GetName().Version.ToString();
+            logger.LogInformation($"AzureAuth Version: {azureauthVersion}");
+
+            string deviceID = TelemetryMachineIDHelper.GetRandomDeviceIDAsync(fileSystem).Result;
+            logger.LogInformation($"Device ID: {deviceID}");
```
reillysiemens, line 32: Not a big deal, but I think we could collapse all of these `logger.LogInformation` calls into a single one at the end after we've gathered up all the relevant information to be displayed.

src/AzureAuth/CommandInfo.cs
goagain: `resource` is deprecated and I use `--scope` instead.

src/AzureAuth/OutputMode.cs
```diff
@@ -29,5 +29,10 @@ public enum OutputMode
         /// The none.
         /// </summary>
         None,
+
+        /// <summary>
+        /// The SID.
```
shalinikhare27, line 34:
```suggestion
        /// The Security Identifier.

```

src/AzureAuth/ExceptionListToStringConverter.cs
```diff
@@ -100,7 +100,7 @@ public string Message
             /// <summary>
             /// Sets the error code.
             /// </summary>
-            /// <param name="exception">exception from which error code is extracted</param>
+            /// <param name="exception">exception from which error code is extracted.</param>
```
shalinikhare27, line 103:
```suggestion
            /// <param name="exception">Exception from which error code is extracted.</param>

```

src/AzureAuth/CommandMain.cs
```diff
@@ -23,6 +23,7 @@ namespace Microsoft.Authentication.AzureAuth
     /// The command main class parses commands and dispatches to the corresponding methods.
     /// </summary>
     [Command(Name = "azureauth", Description = "A CLI interface to MSAL authentication")]
+    [Subcommand(typeof(CommandInfo))]
```
shalinikhare27, line 26: <curious> - is this the first time we are ever introducing a subcommand?

src/AzureAuth/CommandInfo.cs
```diff
@@ -0,0 +1,64 @@
...
+    using System.IO.Abstractions;
+    using System.Reflection;
+    using McMaster.Extensions.CommandLineUtils;
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Office.Lasso.Telemetry;
+
+    /// <summary>
+    /// The command <see cref="CommandInfo"/> shows debug information and system runtime.
+    /// </summary>
+    [Command(Name = "info", Description = "Show debug information of AzureAuth. Please provide when asking for help.")]
```
shalinikhare27, line 15:
```suggestion
    [Command(Name = "info", Description = "Shows AzureAuth debug information. Please provide when asking for help.")]

```

src/AzureAuth/CommandInfo.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// <param name="fileSystem">The file system.</param>
+        /// <returns>The error code: 0 is normal execution, and the rest means errors during execution.</returns>
+        public int OnExecute(ILogger<CommandInfo> logger, IFileSystem fileSystem)
+        {
+            if (this.ResetDeviceID)
+            {
+                return this.ResetID(logger, fileSystem);
+            }
+
+            Assembly assembly = Assembly.GetExecutingAssembly();
```
shalinikhare27, line 39: Can this ever return a null?

shalinikhare27 (commented):
First look!

src/AzureAuth/CommandInfo.cs
goagain: Nope. Different with `GetEntryAssembly`, `GetExecutingAssembly` will return the assembly where it is invoked. Because AzureAuth is a .Net application, `GetExecutingAssembly` will always return a value.

src/AzureAuth/CommandMain.cs
goagain: Yes, it is. The usage of Subcommand is from Lasso documentation.

src/AzureAuth/OutputMode.cs
```diff
@@ -29,5 +29,10 @@ public enum OutputMode
         /// The none.
         /// </summary>
         None,
+
+        /// <summary>
+        /// The Security Identifier.
+        /// </summary>
+        SID,
```
kyle-rader, line 36: I recall the spec included having a way to get your SID, but I'm not sure we want to make it a formal output mode, I was thinking perhaps, the SID could be included in the "info" output. I realize that requires authenticating though, so I'm curious what others think about this as the medium for which to get a SID.

Here's what I don't like:
The goal of allowing the user to see their SID easily is for them to tell it to us, when they need us to review their telemetry. In order to use this easily, one would have to run the entire main (to be `aad`) command, with all the requires args, **or**, use an alias. As we branch out, making a consistent way to get this I think will be a much better user experience. 

I think an easier UX would be to include the SID in the output of the new `info` command. And the auth to do it, could be to AzureDevops - since we are going to be hardcoding the VS and ADO Guids into the code base for PATs anyway.

kyle-rader (changes requested):
Design question.

src/AzureAuth/OutputMode.cs
kyle-rader: We could also see if the SID is available on Windows directly from the OS, and only use auth to get it, on Mac and Linux.

goagain: On Windows I didn't get a reliable way to get the SID. Although some command like `wmic.exe useraccount where "localaccount=true" get name,sid,disabled` can show the SID on the computer, but it is different with what we get from token.

kyle-rader: That's interesting. Sorry I didn't see the internal design review first! Will continue discussion there.

reillysiemens (changes requested):
@goagain: Overall, I think this is looking great! Very clean code here. However, I think we should add some tests.

src/AzureAuth.Test/AzureAuth.Test.csproj
```diff
@@ -1,19 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup Condition="$([MSBuild]::IsOSPlatform('Windows'))">
-    <TargetFramework>net6.0-windows10.0.19041.0</TargetFramework>
```
kyle-rader, line 4: Ah this is some clean up from when we moved to purely net6. Thanks!

kyle-rader (changes requested):
@goagain I beleive the core business logic doesn't warrent unit tests, because it's mostly just a new CLI command, and printing some info. Managing the Machine ID itself, is worth unit testing, and that is owned by Lasso. 

After seeing that interface in use here, I think it would benfit us to make an update in Lasso, to handle fully unit testing the "MachineIDHelper" class. That class can have a more complete interface that handles all this.

src/AzureAuth/CommandInfo.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        /// <summary>
+        /// This method executes the info process.
+        /// </summary>
+        /// <returns>The error code: 0 is normal execution, and the rest means errors during execution.</returns>
+        public int OnExecute()
+        {
+            if (this.ResetDeviceID)
+            {
+                return this.ExecuteResetDeviceID();
```
kyle-rader, line 47: Last major design question about this - we are doing a different action given the presence of a flag. This is a design smell that I know I'm guilty of making in the past. 

If the user knows they want to function of the flag, and the flag is not modifying the actual command being run. Then you likely don't want to use a flag, what you have is another command. Perhaps a sub-command in this case. 

This is the same kind of code smell as passing a bool into a function, and then doing thing A if true, otherwise thing B if false. In that case - you can simplify by writing functionA and functionB, which take no parameters.

So, I dislike using a flag to change the entire logic of the command. I think it will be more simple and clear if we either:
1. Make the flag _modify_ the current command. meaning if set, do reset the device if, and then proceed with the rest of the command printing the info, or
2. Make the reset device id a separate, sub-command on the info command so that we have
   ```shell
   azureauth info -> prints info
   azureauth info reset-device-id -> only resets device id.
   ```

kyle-rader (changes requested):
I have 1 last design question :)

src/AzureAuth/CommandInfo.cs
kyle-rader: @reillysiemens @goagain what do you think about sub-command vs modifier for resetting device id?

goagain: I really thought about this question. Both options work for me. The answer why I took `--reset-device-id` was `--clear` did the same thing in CommandMain.

kyle-rader: That makes sense, and unfortunately in this case, `--clear` is also an example of this design smell which I regret. The reason `--clear` is attached to the command is that in order to clear, at present, we still need to make a Public client application instance with valid parameters, which we get on that command. I will open an issue to track simplifying the clear function.

kyle-rader: If you are fine with either option. I think having this flag as a modifer, so clear it, and then immeditaely regnerate and print the new one works, because ten someone can easily see that it's changing without needing more UI from another command.

goagain: I am glad to see things go to consistency.

src/AzureAuth/Commands/CommandInfo.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            Assembly assembly = Assembly.GetExecutingAssembly();
+            string azureauthVersion = assembly.GetName().Version.ToString();
+            string deviceID = TelemetryDeviceID.GetAsync(fileSystem).Result;
+
+            logger.LogInformation(
+                $"AzureAuth Version: {azureauthVersion} \n" +
+                $"Device ID: {deviceID} \n" +
+                $"To reset your device identifier, Run `azureauth info reset-device-id`");
+
```
kyle-rader, line 36:
```suggestion
            logger.LogInformation($"AzureAuth Version: {azureauthVersion}");
            logger.LogInformation($"Telemetry Device ID: {deviceID}");
            logger.LogInformation("");
            logger.LogInformation("To reset your device identifier run the following command:");
            logger.LogInformation("  azureauth info reset-device-id");


```

kyle-rader: This style of using the Logger will ensure the output all makes sense and is readable regardless of what verbosity level is being used. Otherwise you will have messages spanning lines that don't fit nicely with the debug or trace message styles. It also makes it easier to add and remove lines of output in the future.

Anytime you are writing a command that intend someone to copy and paste, put that command on it's own line with at least 2 extra leading spaces and no surrounding qutations.
* the quotes make it hard to select in your terminal
* the extra line and sapces makes it easy to see separate the command from the text, and make it easier to copy since it's not at the border of your CLI so clicking 1-2 characters off doesn't matter.

src/AzureAuth/Commands/Info/CommandInfoResetDeviceID.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    using System.IO.Abstractions;
+    using McMaster.Extensions.CommandLineUtils;
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Office.Lasso.Telemetry;
+
+    /// <summary>
+    /// The command <see cref="CommandResetDeviceID"/> reset the device ID in local storage.
+    /// </summary>
+    [Command(Name = "reset-device-id", Description = "Reset your device identifier.")]
```
kyle-rader, line 14:
```suggestion
    [Command(Name = "reset-device-id", Description = "Reset your AzureAuth telemetry device identifier.")]

```
Only saying deveice identitifer may scare someone as that could easily be a global thing. Let's make sure it's clear this is AzureAuth specific.

kyle-rader (changes requested):
A few final suggestions on formatting and wording.

kyle-rader (approved):

reillysiemens (approved):

PR #188, opened 2023-01-30
Update workflow to release Linux binaries

mvanchaa:


.github/workflows/release.yml
```diff
@@ -168,10 +164,23 @@ jobs:
       with:
         name: azureauth-${{ github.event.inputs.version }}-osx-arm64.tar.gz
         path: azureauth-${{ github.event.inputs.version }}-osx-arm64.tar.gz
+      
+  linux_release:
+    runs-on: ubuntu-latest
+    needs: [validate]
+    steps:
+    - name: Build, Sign and Publish Linux binaries
+      run: python .\bin\trigger_azure_pipelines.py
```
reillysiemens, line 173: These are backslashes and I think you're gonna find that gets weird (probably doesn't work as you expect) on Linux. Try `/` instead.

README.md
```diff
@@ -73,6 +73,27 @@ curl -sL https://raw.githubusercontent.com/AzureAD/microsoft-authentication-cli/
 of currently running processes, so you'll need to relaunch applications (or source your shell profile) before the `$PATH`
 changes take effect.
 
+## Linux
+
+We provide a `.deb` file that can be installed using the following script.
+
+```bash
+
```
reillysiemens, line 81: I think this extra newline is probably not necessary.

reillysiemens (changes requested):
I made some minuscule requests for changes here. Overall this looks good. However, I think we should wait to merge this until we know we can actually publish the `.deb` package.

mvanchaa:
> I made some minuscule requests for changes here. Overall this looks good. However, I think we should wait to merge this until we know we can actually publish the `.deb` package.

I agree with you. Thanks for reviewing @reillysiemens

mvanchaa:
Closed this by mistake.

PR #189, opened 2023-02-03
Migrate to Sub-Commands!

kyle-rader:
# Reviewing this PR
This PR has a big diff, but is actually doing very little in the way of behavioral changes. All the change is purely in CLI command structure.

## Move `CommandMain` to `CommandAad`
The main change here is a move, of the current (and only) `Command` from `CommandMain` to `CommandAad`.

## new Top level Command `CommandAzureAuth`.
This class is thew new top level command which only prints the help text.

Then we add shells for the incoming sub-commands of `ado`, `ado pat`, and `ado token`.

So all current funcionality is the same, but now refereced via
```
azureauth aad --resource ... --client ... --tenant
```

## Add skeleton for new commands
```
azureauth ado
azureauth ado pat
azureauth ado token
```


goagain (changes requested):
I feel very confident for this PR, but I also see there is a lot of warnings from stylecop.

reillysiemens:
I agree with @goagain here. Let's try to address those StyleCop warnings if we can.

kyle-rader:
> I feel very confident for this PR, but I also see there is a lot of warnings from stylecop.

Yes, great point - I'll update this PR with a couple things
1. Fix newly introduced stylecop issues
2. Revert the change in the CommandMainTest (Now CommandAadTest) from using `this.` - that was a VS refactor fail.

goagain (approved):
I am ok with the absence of comments in the skeleton. I believe we will add them in further.

src/AzureAuth/Commands/Ado/CommandToken.cs
```diff
@@ -0,0 +1,31 @@
...
+
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Office.Lasso.Telemetry;
+
+    /// <summary>
+    /// ADO Command for using either an ADO PAT or acquiring an AAD Access Token.
+    /// </summary>
+    [Command("token", Description = @"Return an AzureDevops PAT from an env var or authenticate the user via MSAL.
+This command should only be used by short running processes that can use either an Azure Devops PAT (long-lived)
+or an AAD Access Token (short-lived).")]
```
reillysiemens, line 16: I don't have an immediate suggestion, but is there any way for this help text to be... less long? Like maybe half the size? 😅

src/AzureAuth/Commands/CommandAzureAuth.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Microsoft.Authentication.AzureAuth.Commands
+{
+    using McMaster.Extensions.CommandLineUtils;
+
+    /// <summary>
+    /// The command main class parses commands and dispatches to the corresponding methods.
+    /// </summary>
+    [Command(Name = "azureauth", Description = "A CLI interface to MSAL (Microsoft Authentication Library)")]
+    [Subcommand(typeof(CommandAad), typeof(CommandAdo))]
```
reillysiemens, line 12: Just curious here. Why use
```csharp
[Subcommand(typeof(Ado.CommandPat))]
[Subcommand(typeof(Ado.CommandToken))]
```
in `CommandAdo`, but use
```csharp
[Subcommand(typeof(CommandAad), typeof(CommandAdo))]
```
here? Is one preferable over the other? Would it help us to be consistent?

src/AzureAuth/Commands/CommandAdo.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// </summary>
+    [Command("ado", Description = "A collection of Azure Devops (ADO) specific authentication commands.")]
+    [Subcommand(typeof(Ado.CommandPat))]
+    [Subcommand(typeof(Ado.CommandToken))]
+    public class CommandAdo
+    {
+        /// <summary>
+        /// Execute the command, showing the help text since this is only a parent command. returns 0.
+        /// </summary>
+        /// <param name="app">The command app insntace.</param>
```
reillysiemens, line 19:
```suggestion
        /// <param name="app">The command app instance.</param>

```

reillysiemens (changes requested):
Some minor suggestions here. Overall this is great and I'd like to approve as soon as they're addressed.

src/AzureAuth/Commands/CommandAzureAuth.cs
kyle-rader: Ah good catch, I will update the other to use individual rows. I prefer individual attributes, because they can be indiependetly added, removed, or re-ordered line by line, instead of having to edit the inline order, and manage commas.

src/AzureAuth/Commands/Ado/CommandToken.cs
kyle-rader: This is only rendered on the help specifically for `azureauth ado --help` at which point I think it's important the `pat` and `token` comands clearly distinguish themselves. We don't have, at present other plans to better document this elsewhere and I dislike the thought of the docs not living on the command itself. Happy to wordsmith but i don't think this is place to try to save a few characters.

reillysiemens: I think we can make a big improvement to this, but I don't think it needs to be done now.

reillysiemens (dismissed):

src/AzureAuth/Commands/Ado/CommandToken.cs
kyle-rader: I've tried to make this shorter, and dropped a link to our aka.ms/azureauth link. We can follow up this migration (or include) a PR revisiting the README and it's coverage of usage.

reillysiemens (approved):

goagain (approved):

goagain:
🚀🚀🚀

kyle-rader:
Ack I didn't update the CHANGELOG - will do in Follow up PR

PR #191, opened 2023-02-08
Update Changelog with recent features done and in progress. 

kyle-rader:
This adds details about the new command structure and new commands to the CHANGELOG, including calling out that it is a breaking change. 

This has made me wonder if we should avoid making this a breaking change. (It doesn't need to be, we could maintain current functionality and still add all the new commands.)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "cSpell.words": [
+        "MSAL"
+    ]
+}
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "cSpell.words": [
+        "MSAL"
```
goagain, line 3: What is this part used for?

goagain (dismissed):

.vscode/settings.json
kyle-rader: This prevents VSCode from teating "MSAL" like a misspelled word.

goagain (approved):

goagain (approved):

PR #194, opened 2023-02-09
Always add a newline before the path install on Mac.

kyle-rader:
If someone already has content in this file, and it doesn't end with a newline, then this will be appended incorrectly. I a simple way to avoid this is always include our own newline.


shalinikhare27 (dismissed):

goagain (dismissed):
Simple and straight!

goagain (approved):

goagain (approved):

shalinikhare27 (approved):

PR #195, opened 2023-02-09
Upgrade all dependencies for security policy

goagain:
Upgrade all dependencies for security policy.

`System.IO.Abstractions.TestingHelpers` is conflicted with Lasso, and because it's in a test project, we should be able to safely ignore it.


kyle-rader-msft:
@goagain before we update MSAL, I'd also like to make sure we review the Changelog to see what changes have been made there, and if/how they impact any of our usage.

shalinikhare27 (commented):
Only one outstanding question here is that are we relying on unit tests to make sure everything is functional?

PR #196, opened 2023-02-10
Add class for getting PAT value from env vars.

kyle-rader:
## This PR
This adds a new class (or sometimes called a service in OO code, where we are not using OO as much). 

`AdoPat` is a `static` class and this PR introduces the `PatFromEnv` method which takes an `IEnv` (defined by Lasso) and returns a `record`, `PatResult` which has the value, and source - which will be useful for debug logging to record where a PAT is coming from when it is used in the `ado token` command. 

## Next PR
The next PR will will focus on doing auth in the command and start finding the abstractions around ado auth we're going to want.


src/AzureAuth/Ado/Constants.cs
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.AzureAuth.Ado
```
kyle-rader, line 4: Note, this file is added in the prior PR, and will be merged indepednently first.

AzureAuth.lutconfig
```diff
@@ -0,0 +1,6 @@
+<LUTConfig Version="1.0">
+  <Repository />
+  <ParallelBuilds>true</ParallelBuilds>
+  <ParallelTestRuns>true</ParallelTestRuns>
+  <TestCaseTimeout>30000</TestCaseTimeout>
+</LUTConfig>
```
shalinikhare27, line 6: Is the LUT config used for Live Unit Testing configurations?

AzureAuth.lutconfig
```diff
@@ -0,0 +1,6 @@
+<LUTConfig Version="1.0">
+  <Repository />
+  <ParallelBuilds>true</ParallelBuilds>
+  <ParallelTestRuns>true</ParallelTestRuns>
+  <TestCaseTimeout>30000</TestCaseTimeout>
```
shalinikhare27, line 5: What does this timeout imply?

src/AzureAuth/Ado/AdoToken.cs
```diff
@@ -0,0 +1,74 @@
...
+namespace Microsoft.Authentication.AzureAuth.Ado
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    using Microsoft.Office.Lasso.Interfaces;
+
+    /// <summary>
+    /// Business logic for either getting an ADO PAT from <see cref="IEnv"/>
```
shalinikhare27, line 13: [nit] Since our code lives in github, we might want to update ADO usages to AzureDevOps?

src/AzureAuth/Ado/AdoToken.cs
```diff
@@ -0,0 +1,74 @@
...
+            /// <summary>
+            /// Gets the source env var that contained the PAT.
+            /// </summary>
+            public string EnvVarSource { get; init; }
+
+            /// <summary>
+            /// Gets the PAT value itself.
+            /// </summary>
+            public string Value { get; init; }
+        }
```
shalinikhare27, line 72: Consider moving this into a PatResult.cs file.

src/AzureAuth/Ado/AdoToken.cs
```diff
@@ -0,0 +1,74 @@
...
+            };
+        }
+
+        /// <summary>
+        /// The result struct for retrieving a PAT out of the environment.
+        /// </summary>
+        public record PatResult
+        {
+            /// <summary>
+            /// Gets a value indicating whether a PAT was found in the env.
```
shalinikhare27, line 59:
```suggestion
            /// Gets a value indicating whether a PAT was found in the environment.
```

src/AzureAuth/EnvVars.cs
```diff
@@ -23,6 +23,16 @@ internal static class EnvVars
         /// </summary>
         public static readonly string NoUser = $"{EnvVarPrefix}_NO_USER";
 
+        /// <summary>
+        /// Name of the env var for the Azure DevOps pipelines default personal access token.
+        /// </summary>
+        public static readonly string SystemAccessToken = "SYSTEM_ACCESSTOKEN";
```
shalinikhare27, line 29: We don't prefix this because this is the default AT from ADO?

src/AzureAuth/Ado/AdoToken.cs
```diff
@@ -0,0 +1,74 @@
...
+                        Value = pat,
+                        EnvVarSource = envVar,
+                    };
+                }
+            }
+
+            return new ()
+            {
+                Exists = false,
+            };
```
shalinikhare27, line 50: We only implement this if we came out of the foreach loop, which also means `Exists is false`. Any reason we set it to false again?

src/AzureAuth/Ado/AdoToken.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// <summary>
+    /// Business logic for either getting an ADO PAT from <see cref="IEnv"/>
+    /// or doing normal AAD Authentication.
+    /// </summary>
+    public static class AdoToken
+    {
+        private static readonly IEnumerable<string> PatEnvVars = new[]
+        {
+            EnvVars.AdoPat,
+            EnvVars.SystemAccessToken,
```
shalinikhare27, line 21: Will it be checking in this particular order? Meaning `AdoPat` and then for `SystemAccessToken`? If so, why is this order necessary or not?

src/AzureAuth/Ado/AdoToken.cs
```diff
@@ -0,0 +1,74 @@
...
+        };
+
+        /// <summary>
+        /// Get either an ADO PAT from the environment. If one is not set, then do normal ADO AAD Auth.
```
shalinikhare27, line 25: This PR has not implemented doing the normal ADO AAD Auth, right? This is just checking if env vars are set or not, then returns if any of them are set?

shalinikhare27 (commented):
First look with bunch of questions.

src/AzureAuth/Ado/AdoToken.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// </summary>
+        /// <param name="env">The <see cref="IEnv"/> to use for interacting with the environment.</param>
+        /// <returns>A PAT from the env if set, otherwise null.</returns>
```
reillysiemens, line 28: I think this is wrong. You never return `null`, it's always a `PatResult`. The `PatResult` just might have a `null` for `EnvVarSource` and `Value`.

src/AzureAuth/Ado/AdoToken.cs
```diff
@@ -0,0 +1,74 @@
...
+            }
+
+            return new ()
+            {
+                Exists = false,
+            };
+        }
+
+        /// <summary>
+        /// The result struct for retrieving a PAT out of the environment.
```
reillysiemens, line 54: It's a `record`, not a `struct`.

src/AzureAuth/Ado/AdoToken.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            /// <summary>
+            /// Gets a value indicating whether a PAT was found in the env.
+            /// </summary>
+            public bool Exists { get; init; }
+
+            /// <summary>
+            /// Gets the source env var that contained the PAT.
+            /// </summary>
+            public string EnvVarSource { get; init; }
```
reillysiemens, line 66: I think `EnvVar` is sufficient and you might not need the `Source` suffix. It's not doing any hard, but I don't know if it's necessary. The `<summary>` should help clarify in the event of confusion.

src/AzureAuth.Test/Ado/AdoTokenTest.cs
```diff
@@ -0,0 +1,95 @@
...
+                Exists = false,
+                EnvVarSource = null,
+                Value = null,
+            };
+
+            AdoToken.PatFromEnv(this.mockEnv.Object).Should().Be(expected);
+        }
+
+        [Test]
+        public void EmptyPAT()
```
reillysiemens, line 44:
```suggestion
        public void Empty_PAT()
```

src/AzureAuth.Test/Ado/AdoTokenTest.cs
```diff
@@ -0,0 +1,95 @@
...
+                Exists = false,
+                EnvVarSource = null,
+                Value = null,
+            };
+
+            AdoToken.PatFromEnv(this.mockEnv.Object).Should().Be(expected);
+        }
+
+        [Test]
+        public void Returns_First_PAT_From_EnvVar_List()
```
reillysiemens, line 60: I think it might be better to frame these tests according to the name of the environment variable rather than "first PAT" and "second PAT". The EnvVar List is private and merely an implementation detail.

It might also be better to frame these tests as `AZUREAUTH_ADO_PAT` taking precedence over `SYSTEM_ACCESSTOKEN`.

I would name this test `Returns_AZUREAUTH_ADO_TOKEN` and the next test as `Returns_SYSTEM_ACCESSTOKEN_When_AZUREAUTH_ADO_TOKEN_Not_Defined` (or similar). 🤷‍♂️

reillysiemens (changes requested):
Truly minor suggestions here. Aside from small typos or docs comments I don't think you _need_ to take anything I've said into account here, but you might consider these before merging.

AzureAuth.lutconfig
reillysiemens: I'm pretty sure that's the case, but IDK if it needs to be checked in. We haven't done that so far...

AzureAuth.lutconfig
reillysiemens: I think this is just the time in milliseconds (so, 30 seconds) before a test case is considered timed out. Of course I can't find **any** documentation for this feature, so... that's just a guess.

src/AzureAuth/Ado/AdoToken.cs
reillysiemens: `Azure DevOps` (with a space between words) would be better, but I also think ADO is a common enough abbreviation for us to get away with it here, especially for an internal comment rather than user-facing text. I could go either way. 🤷‍♂️

src/AzureAuth/EnvVars.cs
reillysiemens: Yeah, exactly. This is an env var that's defined outside of AzureAuth, but that we respect because of how common it is.

src/AzureAuth/Ado/AdoToken.cs
reillysiemens: Not having a strong C# background I'm genuinely curious about this. Why? (Or, conversely, why not?)

From my understanding of the structure here the `PatResult` is the main value that's meant to be passed around. The `AdoToken` class is effectively acting like a static method which constructs this value. I see `PatResult` and `AdoToken` as being intimately linked, but maybe they don't need to be?

Is there a structural reason to do this or is it just following the principal of "one class per file"?

src/AzureAuth/Ado/AdoToken.cs
reillysiemens: I don't think `Exists` exists if we come out of the loop. A brand new `PatResult` is being created here. Are you suggesting we let a default be used? I think that might be `null` in this case. 🤔

src/AzureAuth/Ado/AdoToken.cs
reillysiemens: Yeah. Like with other env vars we prefer the name we control over the other one we support. We respect env vars with our prefix above all else.

src/AzureAuth/Ado/AdoToken.cs
reillysiemens: I noticed the same thing. Maybe future code will change this, but for now I think this comment is incorrect. It just gets a PAT or `null`.

AzureAuth.lutconfig
kyle-rader: If you do not have this file, Visual Studio will ask you to complete the configuration every time you start Live UT. Having it checked in just makes it easier for new devs, or for yourself on a new machine.

AzureAuth.lutconfig
kyle-rader: This is my understanding as well, except I changed it to 30 sec. It defaulted to 180... which is ridiculous for a unit test.

src/AzureAuth/Ado/AdoToken.cs
kyle-rader: The fact they are intimately linked, is why I chose not to put this into another class. The Pat from env result is an implementation detail. I'd like to make it private. But we run into visibility issues in the Tests, since I can't just put the test for this directly into this file... like we could in Rust.

Alternatively to a Record - this could return a named tuple. But this is a little uglier IMO, though it would work and then no private type would be needed.

src/AzureAuth/Ado/AdoToken.cs
kyle-rader: The default value of `bool` is false, but this is not obvious. Strings default to `null` and this is the `default` but it's clearer to set it directly, epecially with primitive types like bool and int.

src/AzureAuth/Ado/AdoToken.cs
kyle-rader: This is specified in the PR description.

src/AzureAuth/Ado/AdoToken.cs
shalinikhare27: Yeah if we think ADO is known, then it's cool.

src/AzureAuth/Ado/AdoToken.cs
```diff
@@ -0,0 +1,74 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    using Microsoft.Office.Lasso.Interfaces;
+
+    /// <summary>
+    /// Business logic for either getting an ADO PAT from <see cref="IEnv"/>
+    /// or doing normal AAD Authentication.
```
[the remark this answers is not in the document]
shalinikhare27: Do we need to specify "normal"?

src/AzureAuth/Ado/AdoToken.cs
shalinikhare27: Yep, was a bit confusing from the summary comment.

src/AzureAuth/Ado/AdoToken.cs
kyle-rader: The public command is going to be `ado` with the name in the help, so I don't think we can make the argument that the full name is needed here, and not there. One could argu the command should be `azure-devops`, but I think the commands being short, easy to type, and are well enough understood we don't need that, + we do spell it out in the help text.

src/AzureAuth/Ado/AdoToken.cs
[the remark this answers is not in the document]
shalinikhare27: @reillysiemens see github automatically adds a new line while giving suggestions. I think when I had a nit comment yesterday, it added a new line which either the reviewer should manually remove or contributor can choose to not commit directly.

AzureAuth.lutconfig
```diff
@@ -0,0 +1,6 @@
+<LUTConfig Version="1.0">
```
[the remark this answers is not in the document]
kyle-rader: With out it checked in, every new machine or clone will have VS prompt to create a new one before you can launch Live Unit testing, so I'd love to have one checked in, and with shorter timeouts than the default of 180 seconds.

src/AzureAuth/Ado/AdoToken.cs
kyle-rader: @shalinikhare27 is this reasoning sound enough for you to leave this inside the AdoToken class?

src/AzureAuth/Ado/AdoToken.cs
kyle-rader: Goo catch, updated for the method this turned into :)

src/AzureAuth/Ado/AdoToken.cs
kyle-rader: Mmmm, I'm going to Won't Fix this one. I considerd just Source. But I prefer the self-documenting code to having the rely on XML docs because it's not clear enough on it's own.

AzureAuth.lutconfig
shalinikhare27, line 1: If we do want to keep it in source, we should add a comment on the purpose since it wasn't obvious to me immediately.

AzureAuth.lutconfig
shalinikhare27: What does test case timing out mean? This is just for my own understanding.

reillysiemens: It takes longer than 30 seconds to run.

src/AzureAuth/Ado/AdoToken.cs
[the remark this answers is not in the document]
reillysiemens: It was my mistake not to remove the automatically added newline.

reillysiemens (approved):

PR #199, opened 2023-02-14
Add CodeQL in release workflow

goagain:


kyle-rader (approved):

.github/workflows/release.yml
```diff
@@ -47,6 +47,13 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v2
+      
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
```
kyle-rader, line 52: If we can easily wrap our existing build process with this, can we also add it to the PR workflow. We only run this at release time, but it would be much better to get CodeQL warnings or errors at PR time.

kyle-rader (changes requested):

kyle-rader (dismissed):

kyle-rader:
@goagain Approved, but requires rebase or merge from main.

kyle-rader-msft (approved):

shalinikhare27 (approved):

PR #200, opened 2023-02-21
Dogfood azureauth for nuget auth

kyle-rader:
I dislike constantly battling the Azure Artifacts Credential provider and having to go through Device Code Flow auth all the time. Since version 1.0 of that tool web auth is still broken, so I propose a solution here to easily allow us to use AzureAuth to handle nuget authentication when building AzureAuth :)

## Opt-In to usage via new scripts
* `bin/dotnet` - a Mac dotnet wrapper that sets the `ADO_TOKEN` env var first.
* `bin/azureauth` - a Mac wrapper to build and run AzureAuth in debug mode.
* `bin\dotnet.cmd` - Runnable on Windows as `bin\dotnet` - sets `ADO_TOKEN` and calls `dotnet`
* `bin\azureauth.cmd` - Runnable on Windows as `bin\azureauth` - uses `bin\dotnet.cmd` to handle auth and run AzureAuth in debug mode.

### Locally
The new entry in nuget.config won't have any impact if the env var is not set locally. If the ADO cred provider is still installed and configured it will be used if the nuget.config gets an empty value that can't auth.

### In GitHub Actions
Having this entry will actually make the CI able to use the same auth config as we can use locally, and our CI can set the same ADO_TOKEN to a valid SYSTEM_ACCESSTOKEN (an ADO PAT).

### DIFF
--- a/bin/dotnet
+++ b/bin/dotnet
@@ -2,5 +2,5 @@
 
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
-export ADO_TOKEN=$(azureauth --client 872cd9fa-d31f-45e0-9eab-6e460a02d1f1 --scope 499b84ac-1321-427f-aa17-267ca6975798/.default --tenant 72f988bf-86f1-41af-91ab-2d7cd011db47 --output token)
+export ADO_TOKEN=$(azureauth --prompt-hint "azureauth dev nuget" --client 872cd9fa-d31f-45e0-9eab-6e460a02d1f1 --scope 499b84ac-1321-427f-aa17-267ca6975798/.default --tenant 72f988bf-86f1-41af-91ab-2d7cd011db47 --output token)
 dotnet $*
--- a/bin/dotnet
+++ b/bin/dotnet
@@ -1,0 +1,6 @@
+#!/usr/bin/env sh
+
+# Copyright (c) Microsoft Corporation.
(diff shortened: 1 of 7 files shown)


bin/dotnet
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+export ADO_TOKEN=$(azureauth --client 872cd9fa-d31f-45e0-9eab-6e460a02d1f1 --scope 499b84ac-1321-427f-aa17-267ca6975798/.default --tenant 72f988bf-86f1-41af-91ab-2d7cd011db47 --output token)
```
mvanchaa, line 5: should we consider having the prompt-hint option here?

mvanchaa: for telemetry reasons

kyle-rader: No telemetry is collected here ;)

kyle-rader: But if we get that sorted globally for ourselves, then it would be - so good call, I will add a prompt hint.

johnterickson:
No need for secrets (or AzureAuth) for this in the CI scenario https://github.com/johnterickson/actions-oidc-demo

reillysiemens (approved):

shalinikhare27 (approved):

PR #203, opened 2023-02-23
[Unlit] Add options to the `ado token` command

kyle-rader:
This PR adds the options declared in our design spec for the `ado token` command. Several of the options are intentionally the same as those on the `aad` command, so those have been made public, and re-used in this command.

The only options where that is not done, is where their behavior differs.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,10 @@
 {
     "cSpell.words": [
-        "MSAL"
+        "authflow",
+        "Corext",
+        "devicecode",
+        "MSAL",
+        "prompthint",
+        "validargs"
     ]
 }
--- a/src/AzureAuth/Commands/CommandAad.cs
+++ b/src/AzureAuth/Commands/CommandAad.cs
@@ -27,34 +27,68 @@ namespace Microsoft.Authentication.AzureAuth.Commands
     [Command("aad", Description = "todo")]
     public class CommandAad
     {
-        private const string ResourceOption = "--resource";
-        private const string ClientOption = "--client";
-        private const string TenantOption = "--tenant";
-        private const string PromptHintOption = "--prompt-hint";
-        private const string ScopeOption = "--scope";
-        private const string ClearOption = "--clear";
-        private const string DomainOption = "--domain";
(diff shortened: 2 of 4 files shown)


goagain (approved):

reillysiemens (approved):

goagain (dismissed):

reillysiemens (approved):

.vscode/settings.json
```diff
@@ -1,5 +1,10 @@
 {
     "cSpell.words": [
-        "MSAL"
+        "authflow",
+        "Corext",
```
shalinikhare27, line 4: Does C need to be upper case here since we use lower case unless it's an acronym like MSAL.

src/AzureAuth/Commands/CommandAad.cs
```diff
@@ -27,34 +27,77 @@ namespace Microsoft.Authentication.AzureAuth.Commands
...
+        /// </summary>
+        public const string AuthModeHelperText = @"Authentication mode. Default: web.
 You can use any combination with multiple instances of the --mode flag.
 Allowed values: [all, web, devicecode]";
 #endif
 
+        /// <summary>
+        /// The help text for the <see cref="DomainOption"/> option.
+        /// </summary>
+        public const string DomainHelpText = "Preferred domain to filter cached accounts by.If a single account matching the preferred domain is in the cache it is used, otherwise an account picker will be launched.\n";
```
shalinikhare27, line 79:
```suggestion
        public const string DomainHelpText = "Preferred domain to filter cached accounts by. If a single account matching the preferred domain is in the cache it is used, otherwise an account picker will be launched.\n";
```

shalinikhare27 (dismissed):

.vscode/settings.json
kyle-rader: Yes it does, because the spelling suggestion was for the actual Enviropnemnt Variable name which can be case sensitive depending on platform. The reason for adding all of these is to avoid accidental spelling or case changes to words in VS Code.

reillysiemens (approved):

shalinikhare27 (approved):

PR #207, opened 2023-02-23
Update signing script to use keycodes specific to the given platform

mvanchaa:
Changes:
- Currently, the `sign.py` requires keycodes of all platforms to be present in the env variables even if we are running it on a single platform. Updating the script to have keycodes specific to the given platform as mandatory.


reillysiemens (changes requested):
Copying my response from an earlier discussion:

My suggestion on your PR is to take the entire `try .. except block` and make that a function like `parse_env_vars()` to mirror `parse_args()`, returning `(aad_id, tenant_id, keycodes, customer_correlation_id)` (or a dictionary with those keys). That way environment variable collection is all still handled in one place instead of starting to be spread out across the code. Inside _that_ function do the platform specific env var collection.

kyle-rader-msft (dismissed):

shalinikhare27 (approved):

bin/sign.py
```diff
@@ -226,6 +226,35 @@ def json_tempfile(path: Path, data: JSON) -> Generator[None, None, None]:
...
+            case "win10-x64":
+                # This key code is used for signing .exes and .dlls on both Windows and Mac.
+                key_code_authenticode = os.environ["SIGNING_KEY_CODE_AUTHENTICODE"]
+                key_codes = {"authenticode": key_code_authenticode}
+            case "osx-x64" | "osx-arm64":
+                # This key code is used for signing .exes and .dlls on both Windows and Mac.
+                key_code_authenticode = os.environ["SIGNING_KEY_CODE_AUTHENTICODE"]
+                # This key code is used for signing .dylibs on Macs.
+                key_code_mac = os.environ["SIGNING_KEY_CODE_MAC"]
+                key_codes = {"authenticode": key_code_authenticode, "mac": key_code_mac}
```
reillysiemens, line 245:
```suggestion
                key_codes = {
                    "authenticode": os.environ["SIGNING_KEY_CODE_AUTHENTICODE"],
                    "mac": os.environ["SIGNING_KEY_CODE_MAC"],
                }
```

bin/sign.py
```diff
@@ -226,6 +226,35 @@ def json_tempfile(path: Path, data: JSON) -> Generator[None, None, None]:
...
+            case "linux-x64":
+                # This key code is used for signing .deb on Linux.
+                key_code_linux = os.environ["SIGNING_KEY_CODE_LINUX"]
+                key_codes = {"linux": key_code_linux}
```
reillysiemens, line 249:
```suggestion
                key_codes = {"linux": os.environ["SIGNING_KEY_CODE_LINUX"]}
```

reillysiemens (approved):

bin/sign.py
```diff
@@ -226,6 +229,36 @@ def json_tempfile(path: Path, data: JSON) -> Generator[None, None, None]:
...
+def parse_env_vars(runtime: str) -> tuple[str, str, str, JSON]:
+    """Parse and return environment variables"""
+    try:
+        aad_id = os.environ["SIGNING_AAD_ID"]
+        tenant_id = os.environ["SIGNING_TENANT_ID"]
+        customer_correlation_id = os.environ["SIGNING_CUSTOMER_CORRELATION_ID"]
+        match runtime:
+            case "win10-x64":
+                # This key code is used for signing .exes and .dlls on both Windows and Mac.
+                key_codes = {
```
kyle-rader-msft, line 241: In Python, does defining `key_codes` inside the match arms like this also define `key_codes` in the surround lexical scope? @mvanchaa @reillysiemens

mvanchaa:
> Copying my response from an earlier discussion:
> 
> My suggestion on your PR is to take the entire `try .. except block` and make that a function like `parse_env_vars()` to mirror `parse_args()`, returning `(aad_id, tenant_id, keycodes, customer_correlation_id)` (or a dictionary with those keys). That way environment variable collection is all still handled in one place instead of starting to be spread out across the code. Inside _that_ function do the platform specific env var collection.

Updated!

bin/sign.py
```diff
@@ -38,19 +38,22 @@ def sign_operation(key_code: str, operation: str) -> JSON:
...
-        "ToolVersion" : "1.0"
+        "KeyCode": key_code,
+        "OperationCode": operation,
+        "Parameters": {},
+        "ToolName": "sign",
+        "ToolVersion": "1.0",
     }
 
+
 def linux_sign(key_code: str) -> JSON:
```
shalinikhare27, line 52: [nit]: extra line added

bin/sign.py
```diff
@@ -38,19 +38,22 @@ def sign_operation(key_code: str, operation: str) -> JSON:
...
     """Return the JSON for a `LinuxSign` operation."""
     return sign_operation_linux(key_code, operation="LinuxSign")
 
+
```
shalinikhare27, line 56: [nit]: extra line added

bin/sign.py
```diff
@@ -198,7 +201,7 @@ def auth(tenant_id: str, client_id: str) -> JSON:
             "SubjectName": f"CN={client_id}.microsoft.com",
             "StoreLocation": "LocalMachine",
             "StoreName": "My",
-            "SendX5c" :  "true"
+            "SendX5c": "true",
```
shalinikhare27, line 204: No change except whitespaces. Is this formatting fix intentional?

bin/sign.py
```diff
@@ -226,6 +229,36 @@ def json_tempfile(path: Path, data: JSON) -> Generator[None, None, None]:
     path.unlink()
 
 
+def parse_env_vars(runtime: str) -> tuple[str, str, str, JSON]:
+    """Parse and return environment variables"""
```
shalinikhare27, line 233: Is `"""` required in Python to escape '`"`. I ask because it's `""` in C#. :)

bin/sign.py
```diff
@@ -226,6 +229,36 @@ def json_tempfile(path: Path, data: JSON) -> Generator[None, None, None]:
...
 
+def parse_env_vars(runtime: str) -> tuple[str, str, str, JSON]:
+    """Parse and return environment variables"""
+    try:
+        aad_id = os.environ["SIGNING_AAD_ID"]
+        tenant_id = os.environ["SIGNING_TENANT_ID"]
+        customer_correlation_id = os.environ["SIGNING_CUSTOMER_CORRELATION_ID"]
+        match runtime:
+            case "win10-x64":
+                # This key code is used for signing .exes and .dlls on both Windows and Mac.
```
shalinikhare27, line 240: Just curious, how does this cater to both platforms when you have a platform specific case for Mac right after.

shalinikhare27 (changes requested):
@mvanchaa I see a pattern of not including PR descriptions. This is a feedback and expedites PR reviews. Might not always be obvious from the title on 'why' :)

bin/sign.py
```diff
@@ -38,19 +38,22 @@ def sign_operation(key_code: str, operation: str) -> JSON:
...
 
+
 def sign_operation_linux(key_code: str, operation: str) -> JSON:
     return {
-        "KeyCode" : key_code,
-        "OperationCode" : operation,
-        "Parameters" : {},
-        "ToolName" : "sign",
-        "ToolVersion" : "1.0"
+        "KeyCode": key_code,
```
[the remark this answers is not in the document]
mvanchaa: This is added for styling. `Black` auto formats when this is added.

bin/sign.py
mvanchaa: I added this line for consistency. All the methods in this file are preceded by two lines.

bin/sign.py
mvanchaa: Also, formatting with `Black` introduces them.

bin/sign.py
mvanchaa: Yes. This change is introduced after formatting with Black

bin/sign.py
mvanchaa: This is a standard python docstring convention. https://peps.python.org/pep-0257/#one-line-docstrings

bin/sign.py
mvanchaa: It caters to only one platform. For Mac, there is a another case statement.

shalinikhare27 (approved):

bin/sign.py
mvanchaa: I'm not sure if I understood what you meant, I tried this now and the `key_codes` is not defined outside of match block if it's not executed in any of the case statements. Is that what you meant?

reillysiemens: @kyle-rader-msft: Yes.
```python
>>> x
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'x' is not defined
>>> match False:
...     case True:
...         x = 42
...     case False:
...         pass
...
>>> x
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'x' is not defined
>>> match True:
...     case True:
...         x = 42
...     case False:
...         pass
...
>>> x
42
```

bin/sign.py
reillysiemens: @shalinikhare27: That extra newline is the norm in Python. It's a community "standard".

bin/sign.py
reillysiemens: Technically, it could just be a `"`. You can see this if you do
```python
>>> def foobar():
...     "Amazingly, this is a valid docstring."
...     pass
...    
>>> foobar.__doc__
'Amazingly, this is a valid docstring.
```
, but it's very much against convention to do that and you almost always see docstrings with `"""`. The documentation that @mvanchaa linked points out the main reason
> Triple quotes are used even though the string fits on one line. This makes it easy to later expand it.

reillysiemens (approved):

PR #210, opened 2023-02-24
Upgrade Lasso and match the API changes.

goagain:
Lasso gets the machine id config from App name. We get the root app name from `CommandLineApplication`.
The extension method `GetRoot` came from Lasso since `2023.3.2.3`.
Concurrency bug also got fixed by Lasso.

### DIFF
--- a/src/AzureAuth/Commands/CommandInfo.cs
+++ b/src/AzureAuth/Commands/CommandInfo.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Authentication.AzureAuth.Commands
         /// <param name="fileSystem">The file system.</param>
         /// <param name="app">The command line application.</param>
         /// <returns>The error code: 0 is normal execution, and the rest means errors during execution.</returns>
-        public int OnExecute(ILogger<CommandInfo> logger, IFileSystem fileSystem, CommandLineApplication<CommandInfo> app)
+        public int OnExecute(ILogger<CommandInfo> logger, IFileSystem fileSystem, CommandLineApplication app)
         {
             Assembly assembly = Assembly.GetExecutingAssembly();
             string azureauthVersion = assembly.GetName().Version.ToString();
--- a/src/AzureAuth/Commands/CommandInfo.cs
+++ b/src/AzureAuth/Commands/CommandInfo.cs
@@ -8,6 +8,7 @@ namespace Microsoft.Authentication.AzureAuth.Commands
     using McMaster.Extensions.CommandLineUtils;
     using Microsoft.Authentication.AzureAuth.Commands.Info;
     using Microsoft.Extensions.Logging;
+    using Microsoft.Office.Lasso.Interfaces;
     using Microsoft.Office.Lasso.Telemetry;
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/AzureAuth/Commands/CommandInfo.cs
```diff
@@ -21,13 +22,13 @@ public class CommandInfo
...
-        /// <param name="fileSystem">The file system.</param>
+        /// <param name="telemetryService">The telemetry service.</param>
         /// <returns>The error code: 0 is normal execution, and the rest means errors during execution.</returns>
-        public int OnExecute(ILogger<CommandInfo> logger, IFileSystem fileSystem)
+        public int OnExecute(ILogger<CommandInfo> logger, ITelemetryService telemetryService)
         {
             Assembly assembly = Assembly.GetExecutingAssembly();
             string azureauthVersion = assembly.GetName().Version.ToString();
-            string deviceID = TelemetryDeviceID.GetAsync(fileSystem).Result;
+            string deviceID = (telemetryService as TelemetryServiceBase)?.MachineGUID;
```
kyle-rader, line 31: I'm confused why we need to do this. Casting the telemetry service here couples it to a particular instance, We should be able to continue using the TelemetryDeviceID interface as is, no?

src/AzureAuth/Commands/Info/CommandInfoResetDeviceID.cs
```diff
@@ -15,6 +15,8 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Info
     [Command(Name = "reset-device-id", Description = "Reset your AzureAuth telemetry device identifier.")]
     public class CommandResetDeviceID
     {
+        private const string ApplicationName = "azureauth";
```
kyle-rader, line 18: This creates duplication of the App name that we don't want, if the name ever needs to change we should be able to do that in 1 place. 

Instead, I think we can depend on (add the to argument list) the original `CommandLineApplication` that Lasso registers, and use the Name from that (this is how Lasso derives the app name internally).

kyle-rader (changes requested):

src/AzureAuth/Commands/Info/CommandInfoResetDeviceID.cs
goagain: If I inject `CommandLineApplication`, I will get `CommandLineApplication<CommandResetDeviceID>` and its name is `reset-device-id`. I cannot inject `CommandLineApplication<CommandLineAzureAuth>`

kyle-rader-msft: After looking at this in Lasso I was reminded that we don't need to do anything - we can always get the top level command by going through the current command's `Parent` property. So here, we can follow the `Parent` to get the top level command name. We could hardcode one reference to `.Parent` since we know there's only 1. It might be nice to write an integration test to pin this logic, so if the command ever moves, we know it still gets the right commend name.

kyle-rader-msft (changes requested):

src/AzureAuth/Commands/Info/CommandInfoResetDeviceID.cs
```diff
@@ -15,15 +15,18 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Info
...
```
shalinikhare27, line 18: This string was created in this PR, but never used? Did I miss?

goagain: Good catch!

reillysiemens (approved):

kyle-rader-msft (approved):

PR #211, opened 2023-02-27
Update release workflow to release Linux binaries

mvanchaa:
Changes:
- Added a job which runs parallelly with building and signing of windows and mac binaries. This job triggers ado pipeline which builds, signs and uploads binaries to packages.microsoft.com and to GitHub artifacts.
- The uploaded artifacts will be released as a part of the `release` job once reviewed and approved.

Testing
- Tested the workflow on a private repo.

Pending:
Once approved, I will add the secrets.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,13 +7,6 @@ on:
         description: 'Version'
         required: true
         type: string
-      # debian_revision indicates number of revisions made to the debian configuration without making changes to the source code.
-      # we would set the revision to 2, if we needed to release a new debian package, but the source code of AzureAuth itself had not changed.
-      debian_revision:
-        description: 'Debian revision'
-        required: true
-        default: '1'
-        type: string
       # We use SemVer, anything before 1.0.0 is a pre-release, but this could also include versions like 1.1.0-beta.
       prerelease:
         description: 'Prerelease'
@@ -152,7 +145,7 @@ jobs:
           path: azureauth-${{ github.event.inputs.version }}-${{ matrix.runtime }}
   
   # Build and sign Linux binaries on Azure DevOps and publish them to GitHub and packages.microsoft.com.
-  linux_build_and_sign:
+  linux_release:
     runs-on: ubuntu-latest
     needs: [validate]
     env:
@@ -177,9 +170,11 @@ jobs:
         VERSION: ${{ github.event.inputs.version }}
     - name: Upload ubuntu-amd64 artifact
       uses: actions/upload-artifact@v3
+      env:
+        DEBIAN_REVISION: 1
       with:
-        name: azureauth_${{ github.event.inputs.version }}-${{ github.event.inputs.debian_revision }}_amd64.deb
-        path: ${{ env.ADO_LINUX_ARTIFACT_DOWNLOAD_PATH }}/${{ env.ADO_LINUX_ARTIFACT_NAME }}/azureauth_${{ github.event.inputs.version }}-${{ github.event.inputs.debian_revision }}_amd64.deb
+        name: azureauth_${{ github.event.inputs.version }}-${{ env.DEBIAN_REVISION }}_amd64.deb
+        path: ${{ env.ADO_LINUX_ARTIFACT_DOWNLOAD_PATH }}/${{ env.ADO_LINUX_ARTIFACT_NAME }}/azureauth_${{ github.event.inputs.version }}-${{ env.DEBIAN_REVISION }}_amd64.deb
         
   # Currently we package artifacts into the most commonly accessible archive format for their respective platforms.
   package:
@@ -222,13 +217,15 @@ jobs:
 
   release:
     runs-on: ubuntu-latest
-    needs: [package, linux_build_and_sign]
+    needs: [package, linux_release]
     # The 'release' environment is what requires reviews before creating the release.
     environment:
       name: release
     # These permissions are required in order to use `softprops/action-gh-release` to upload.
     permissions:
       contents: write
+    env:
+        DEBIAN_REVISION: 1
     steps:
     - name: Download win10-x64 artifact
       uses: actions/download-artifact@v3
@@ -245,7 +242,7 @@ jobs:
     - name: Download ubuntu-amd64 artifact
       uses: actions/download-artifact@v3
       with:
-        name: azureauth_${{ github.event.inputs.version }}-${{ github.event.inputs.debian_revision }}_amd64.deb
+        name: azureauth_${{ github.event.inputs.version }}-${{ env.DEBIAN_REVISION }}_amd64.deb
     - name: Create Release
       uses: softprops/action-gh-release@v1
       with:
@@ -257,4 +254,4 @@ jobs:
           azureauth-${{ github.event.inputs.version }}-win10-x64.zip
           azureauth-${{ github.event.inputs.version }}-osx-x64.tar.gz
           azureauth-${{ github.event.inputs.version }}-osx-arm64.tar.gz
-          azureauth_${{ github.event.inputs.version }}-${{ github.event.inputs.debian_revision }}_amd64.deb
+          azureauth_${{ github.event.inputs.version }}-${{ env.DEBIAN_REVISION }}_amd64.deb
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,6 +7,11 @@ on:
         description: 'Version'
         required: true
         type: string
+      debian_revision:
+        description: 'Debian revision'
+        required: true
+        default: '1'
+        type: string
       # We use SemVer, anything before 1.0.0 is a pre-release, but this could also include versions like 1.1.0-beta.
       prerelease:
         description: 'Prerelease'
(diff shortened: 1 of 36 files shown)


.github/workflows/release.yml
```diff
@@ -129,7 +131,22 @@ jobs:
         with:
           name: azureauth-${{ github.event.inputs.version }}-${{ matrix.runtime }}
           path: azureauth-${{ github.event.inputs.version }}-${{ matrix.runtime }}
-
+  
+  # Build and sign Linux binaries on Azure DevOps and publish them to GitHub and packages.microsoft.com.
+  linux_release:
```
mvanchaa, line 136: I'm trying to come up with better name for this job. Looking for suggestions :)

shalinikhare27: Better name for `linus_release` ?

goagain: I disagree with `linus_release`

shalinikhare27 (dismissed):

.github/workflows/release.yml
```diff
@@ -7,6 +7,11 @@ on:
         description: 'Version'
         required: true
         type: string
+      debian_revision:
+        description: 'Debian revision'
+        required: true
+        default: '1'
+        type: string
```
kyle-rader-msft, line 14: Is the Debian revision something we expect to need to change as a workflow parameter? Or would this make more sense as a hard coded env var for the whole workflow?

kyle-rader-msft: Also, a comment about what this is would be nice `#` can be used for a comment line in YAML.

.github/workflows/release.yml
kyle-rader-msft: I'm good with `linux_release`.

kyle-rader-msft: You could use `linux_build_and_sign` to be more specific.

.github/workflows/release.yml
```diff
@@ -129,7 +131,23 @@ jobs:
...
-
+  
+  # Build and sign Linux binaries on Azure DevOps and publish them to GitHub and packages.microsoft.com.
+  linux_release:
+    runs-on: ubuntu-latest
+    needs: [validate]
+    steps:
+    - name: Build, Sign and Publish Linux binaries
+      run: |
+        pip install -r ./bin/requirements.txt
```
kyle-rader-msft, line 142: Do we not need to setup Python, or a particular version of Python?

.github/workflows/release.yml
```diff
@@ -191,6 +209,14 @@ jobs:
       uses: actions/download-artifact@v3
       with:
         name: azureauth-${{ github.event.inputs.version }}-osx-arm64.tar.gz
+    - name: Download osx-arm64 artifact
+      uses: actions/download-artifact@v3
+      with:
+        name: azureauth-${{ github.event.inputs.version }}-osx-arm64.tar.gz
```
kyle-rader-msft, line 215: This looks like a copy paste error, we already download the `osx-arm64` artifact ;)

kyle-rader-msft (changes requested):
:)

.github/workflows/release.yml
mvanchaa: The probability is pretty low.

.github/workflows/release.yml
mvanchaa: Yes we need to. I forgot to update this PR.

.github/workflows/release.yml
mvanchaa: haha yes. I have more update to this PR actually. I will mark it as a draft.

.github/workflows/release.yml
mvanchaa: @kyle-rader-msft added a comment. Does this looks okay?

kyle-rader-msft: AHA That's a very important comment and I think we could add more detail. It sounds like this is a revision number designed for the same reason we have `-rev-n` on the AzureAuth-Lib pacakge, so we can rev the package without reving the azureauth version it's associated with.

So, in theory - we would set the revision to 2, if we needed to release a new debian package, but the source code of AzureAuth itself had not changed, correct?

.github/workflows/release.yml
```diff
@@ -129,7 +132,36 @@ jobs:
...
           path: azureauth-${{ github.event.inputs.version }}-${{ matrix.runtime }}
-
+  
+  # Build and sign Linux binaries on Azure DevOps and publish them to GitHub and packages.microsoft.com.
+  linux_build_and_sign:
+    runs-on: ubuntu-latest
+    needs: [validate]
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
```
kyle-rader-msft, line 142:
```suggestion
      uses: actions/checkout@v3
```

Checkouts is now on `v3`

.github/workflows/release.yml
```diff
@@ -129,7 +132,36 @@ jobs:
...
+        pip install -r bin/requirements.txt
+        python ./bin/trigger_azure_pipelines.py
+      env:
+        AZURE_DEVOPS_BUILD_PAT: ${{ secrets.AZURE_DEVOPS_BUILD_PAT }}
+        ADO_ORGANIZATION: ${{ secrets.ADO_ORGANIZATION }}
+        ADO_PROJECT: ${{ secrets.ADO_PROJECT}}
+        ADO_AZUREAUTH_LINUX_PIPELINE_ID: ${{ secrets.ADO_AZUREAUTH_LINUX_PIPELINE_ID }}
+        VERSION: ${{ github.event.inputs.version }}
+        ADO_LINUX_ARTIFACT_NAME: ${{ secrets.ADO_LINUX_ARTIFACT_NAME }}
+        ADO_LINUX_ARTIFACT_DOWNLOAD_PATH: dist/linux-x64
```
kyle-rader-msft, line 158: Given this value (`ADO_LINUX_ARTIFACT_DOWNLOAD_PATH`) is hardcoded here, to change it would require a PR updating the pipeline. So, effectively it's not much different if the value were hardcoded in the script directly. It's no more difficult to change it - however, having the value in the script makes the script easier to understand beacuse you can see what that path is going to be.

kyle-rader-msft: Though I can also see you may have opted for putting it here, since it is used in the step below, and this let's you only define it in these 2 pipeline steps...

Pipeline are notoriosly easy to break, so to avoid that I would avoid hardcoding the value more than once, and declare this env var at the `Job` level and reference the env var still from the script, but also reference the env var in the upload step below.

kyle-rader-msft (changes requested):

.github/workflows/release.yml
mvanchaa: yes that's exactly right.

.github/workflows/release.yml
mvanchaa: Understood. I will use this as an env var at the job level.

.github/workflows/release.yml
mvanchaa: Added more details in the comment.

.github/workflows/release.yml
```diff
@@ -129,7 +133,37 @@ jobs:
...
+    - name: Build, Sign and Download Linux Binaries
+      run: |
+        pip install -r bin/requirements.txt
+        python ./bin/trigger_azure_pipelines.py
+      env:
+        AZURE_DEVOPS_BUILD_PAT: ${{ secrets.AZURE_DEVOPS_BUILD_PAT }}
+        ADO_ORGANIZATION: ${{ secrets.ADO_ORGANIZATION }}
+        ADO_PROJECT: ${{ secrets.ADO_PROJECT}}
+        ADO_AZUREAUTH_LINUX_PIPELINE_ID: ${{ secrets.ADO_AZUREAUTH_LINUX_PIPELINE_ID }}
+        VERSION: ${{ github.event.inputs.version }}
```
kyle-rader-msft, line 160: I could alsoo see moving all the non-secrets (i.e. actual passwords) to the job env, but I'm not going to block on this.

kyle-rader-msft (approved):

.github/workflows/release.yml
mvanchaa: I'm not sure if we can expose org, project and pipeline id.

mvanchaa: I didn't move them because they are not being used in the next steps :)

.github/workflows/release.yml
```diff
@@ -146,7 +150,37 @@ jobs:
         with:
           name: azureauth-${{ github.event.inputs.version }}-${{ matrix.runtime }}
           path: azureauth-${{ github.event.inputs.version }}-${{ matrix.runtime }}
-
+  
+  # Build and sign Linux binaries on Azure DevOps and publish them to GitHub and packages.microsoft.com.
+  linux_build_and_sign:
```
reillysiemens, line 155: This is really `linux_build_sign_and_publish`

.github/workflows/release.yml
```diff
@@ -7,6 +7,13 @@ on:
         description: 'Version'
         required: true
         type: string
+      # debian_revision indicates number of revisions made to the debian configuration without making changes to the source code.
+      # we would set the revision to 2, if we needed to release a new debian package, but the source code of AzureAuth itself had not changed.
+      debian_revision:
+        description: 'Debian revision'
+        required: true
+        default: '1'
+        type: string
```
reillysiemens, line 16: As discussed, unfortunately this can never change without also changing the entire GitHub release version, so I don't think it needs to be a parameter to the release job. We can hardcode this for now.

reillysiemens (changes requested):

.github/workflows/release.yml
kyle-rader-msft: This makes sense.

shalinikhare27 (approved):

reillysiemens (approved):

PR #212, opened 2023-02-28
Add unlit PatManager and supporting interfaces

reillysiemens:
This PR adds a `PatManager` class to the `Microsoft.Authentication.AdoPat` namespace. It also add supporting interfaces and classes, namely

- A `PatOption` record for collecting common options used when fetching or creating PATs.
- An `IPatCache` interface to make testing easier.
- An `IPatClient` interface to make testing easier.
- Corresponding tests in the `Microsoft.Authentication.AdoPat.Test` namespace.

As in #176 and #183, this entity relationship diagram shows the rough layout of parts needed to make this work.

```mermaid
erDiagram
    CLI ||..|| PatManager : uses
    PatCache }|..|{ StorageWrapper : uses
```

This is phase 3 of 4, where the cache and API client are tied together and orchestrated in the `PatManager`. A subsequent PR will utilize this class in `Program.Main` of the `azureauth ado pat` subcommand. Until that final phase this code is still "unlit" and should not factor into a public release.

### DIFF
--- a/src/AdoPat.Test/PatManagerTest.cs
+++ b/src/AdoPat.Test/PatManagerTest.cs
@@ -1,0 +1,215 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.AdoPat.Test
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using FluentAssertions;
+    using Microsoft.VisualStudio.Services.DelegatedAuthorization;
+    using Moq;
+    using NUnit.Framework;
+
+    public class PatManagerTest
+    {
+        private const string Organization = "contoso";
+        private const string DisplayName = "Test PAT";
+        private const string Scope = "test.scope";
+
+        // This is a test token. A real value would be a much longer string.
+        private const string Token = "Test Token";
+
+        // The Unix Epoch is used as an obviously fake test time which occurs in the past and cannot accidentally be valid.
+        private readonly DateTime unixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+
+        // This list of accounts uses dummy data, not real accounts.
+        private readonly List<Guid> targetAccounts = new List<Guid> { new Guid("b7b59161-cd70-46e9-aca5-883f24060eb1") };
+
+        // This is a dummy authorization ID, not valid in any real contexts.
+        private readonly Guid authorizationId = new Guid("ee0c5586-a96f-4a44-b1d9-8613028b1078");
+
+        // Common mocks which are configured via the Setup and Teardown methods.
+        private Mock<IPatCache> cache;
+        private Mock<IPatClient> client;
+
+        // PAT options which are common to all tests.
+        private static PatOptions PatOptions => new ()
+        {
+            DisplayName = DisplayName,
+            Organization = Organization,
+            Scopes = new string[] { Scope },
+        };
+
+        [SetUp]
+        public void Setup()
+        {
+            this.cache = new Mock<IPatCache>(MockBehavior.Strict);
+            this.client = new Mock<IPatClient>(MockBehavior.Strict);
+        }
+
+        [TearDown]
+        public void Teardown()
+        {
+            // To be certain we have called all the necessary methods we verify
+            // each mock after every test.
+            this.cache.VerifyAll();
+            this.client.VerifyAll();
+        }
+
+        public PatToken PatToken(
+            string displayName = null,
+            string scope = null,
+            List<Guid> targetAccounts = null,
+            DateTime? validTo = null,
+            DateTime? validFrom = null,
+            Guid? authorizationId = null,
+            string token = null)
+        {
+            return new PatToken
+            {
+                DisplayName = displayName ?? DisplayName,
+                Scope = scope ?? Scope,
+                TargetAccounts = targetAccounts ?? this.targetAccounts,
+                ValidTo = validTo ?? this.unixEpoch.AddDays(7),
+                ValidFrom = validFrom ?? this.unixEpoch,
+                AuthorizationId = authorizationId ?? this.authorizationId,
+                Token = token ?? Token,
+            };
+        }
+
+        [Test]
+        public async Task GetPatAsync_ReturnsCachedPatWhenCacheHasValidPat()
+        {
+            // Arrange
+            var expectedPat = this.PatToken();
+            var activePats = new Dictionary<Guid, PatToken>
+            {
+                { expectedPat.AuthorizationId, expectedPat },
+            };
+
+            // The cache should contain our target PAT.
+            this.cache.Setup(c => c.GetPat(It.IsAny<string>())).Returns(expectedPat);
+
+            // The client needs to be injected into the manager, but should be unused.
+            this.client.Setup(c => c.GetActivePatsAsync(It.IsAny<CancellationToken>())).ReturnsAsync(activePats);
+
+            var manager = new PatManager(
+                this.cache.Object,
+                this.client.Object,
+                now: () => this.unixEpoch);
+
+            // Act
+            var pat = await manager.GetPatAsync(PatOptions);
+
+            // Assert
+            pat.Should().BeEquivalentTo(expectedPat);
+        }
+
+        [Test]
+        public async Task GetPatAsync_CreatesNewPatWhenCacheIsMissingPat()
+        {
+            // Arrange
+            var expectedKey = $"{Organization} {DisplayName} {Scope}";
+            var expectedPat = this.PatToken();
+
+            // The cache is empty and therefore returns `null` for any key.
+            this.cache.Setup(c => c.GetPat(It.IsAny<string>())).Returns<PatToken>(null);
+            this.cache.Setup(c => c.PutPat(expectedKey, expectedPat));
+
+            // The client will return a new PatToken upon creation request.
+            this.client.Setup(c => c.CreatePatAsync(
+                It.IsAny<PatTokenCreateRequest>(),
+                It.IsAny<CancellationToken>()))
+            .ReturnsAsync(expectedPat);
+
+            var manager = new PatManager(
+                this.cache.Object,
+                this.client.Object,
+                now: () => this.unixEpoch);
+
+            // Act
+            var pat = await manager.GetPatAsync(PatOptions);
+
+            // Assert
+            pat.Should().BeEquivalentTo(expectedPat);
+        }
+
+        [Test]
+        public async Task GetPatAsync_RegeneratesPatWhenPatHasExpired()
+        {
+            // Arrange
+            var expectedKey = $"{Organization} {DisplayName} {Scope}";
+            var expiringPat = this.PatToken(
+                validTo: this.unixEpoch,
+                validFrom: this.unixEpoch.AddDays(-7));
+            var expectedPat = this.PatToken(
+                authorizationId: new Guid("6135c9a4-d08c-467f-9902-6ff439651657"),
+                token: "Fake Token");
+            var activePats = new Dictionary<Guid, PatToken>
+            {
+                { expiringPat.AuthorizationId, expiringPat },
+            };
+
+            // The PAT is found within the cache, but it has expired.
+            this.cache.Setup(c => c.GetPat(It.IsAny<string>())).Returns(expiringPat);
+            this.cache.Setup(c => c.PutPat(expectedKey, expectedPat));
+
+            // The client will return a new PatToken upon creation request.
+            this.client.Setup(c => c.GetActivePatsAsync(It.IsAny<CancellationToken>())).ReturnsAsync(activePats);
+            this.client.Setup(c => c.RegeneratePatAsync(
+                It.IsAny<PatToken>(),
+                It.IsAny<DateTime>(),
+                It.IsAny<CancellationToken>()))
+            .ReturnsAsync(expectedPat);
+
+            var manager = new PatManager(
+                this.cache.Object,
+                this.client.Object,
+                now: () => this.unixEpoch);
+
+            // Act
+            var pat = await manager.GetPatAsync(PatOptions);
+
+            // Assert
+            pat.Should().BeEquivalentTo(expectedPat);
+        }
+
+        [Test]
+        public async Task GetPatAsync_CreatesNewPatWhenPatIsInactive()
+        {
+            // Arrange
+            var expectedKey = $"{Organization} {DisplayName} {Scope}";
+            var inactivePat = this.PatToken();
+            var expectedPat = this.PatToken(
+                authorizationId: new Guid("5045619c-e1b1-46e3-a1e4-ba5f38ddf49b"),
+                token: "Fake Token");
+
+            // The inactivePat is not in the list of active PATs from Azure DevOps.
+            var activePats = new Dictionary<Guid, PatToken> { };
+
+            // The cache is empty and therefore returns `null` for any key.
+            this.cache.Setup(c => c.GetPat(It.IsAny<string>())).Returns(inactivePat);
+            this.cache.Setup(c => c.PutPat(expectedKey, expectedPat));
+
+            // The client will return a new PatToken upon creation request.
+            this.client.Setup(c => c.GetActivePatsAsync(It.IsAny<CancellationToken>())).ReturnsAsync(activePats);
+            this.client.Setup(c => c.CreatePatAsync(
+                It.IsAny<PatTokenCreateRequest>(),
+                It.IsAny<CancellationToken>()))
+            .ReturnsAsync(expectedPat);
+
+            var manager = new PatManager(
+                this.cache.Object,
+                this.client.Object,
+                now: () => this.unixEpoch);
+
+            // Act
+            var pat = await manager.GetPatAsync(PatOptions);
+
+            // Assert
+            pat.Should().BeEquivalentTo(expectedPat);
+        }
+    }
+}
--- a/src/AdoPat.Test/PatManagerTest.cs
+++ b/src/AdoPat.Test/PatManagerTest.cs
@@ -90,7 +90,7 @@ namespace Microsoft.Authentication.AdoPat.Test
             };
 
             // The cache should contain our target PAT.
-            this.cache.Setup(c => c.GetPat(It.IsAny<string>())).Returns(expectedPat);
+            this.cache.Setup(c => c.Get(It.IsAny<string>())).Returns(expectedPat);
 
             // The client needs to be injected into the manager, but should be unused.
             this.client.Setup(c => c.GetActivePatsAsync(It.IsAny<CancellationToken>())).ReturnsAsync(activePats);
@@ -115,8 +115,8 @@ namespace Microsoft.Authentication.AdoPat.Test
             var expectedPat = this.PatToken();
 
             // The cache is empty and therefore returns `null` for any key.
-            this.cache.Setup(c => c.GetPat(It.IsAny<string>())).Returns<PatToken>(null);
-            this.cache.Setup(c => c.PutPat(expectedKey, expectedPat));
+            this.cache.Setup(c => c.Get(It.IsAny<string>())).Returns<PatToken>(null);
+            this.cache.Setup(c => c.Put(expectedKey, expectedPat));
 
             // The client will return a new PatToken upon creation request.
             this.client.Setup(c => c.CreatePatAsync(
@@ -153,8 +153,8 @@ namespace Microsoft.Authentication.AdoPat.Test
             };
 
             // The PAT is found within the cache, but it has expired.
-            this.cache.Setup(c => c.GetPat(It.IsAny<string>())).Returns(expiringPat);
-            this.cache.Setup(c => c.PutPat(expectedKey, expectedPat));
+            this.cache.Setup(c => c.Get(It.IsAny<string>())).Returns(expiringPat);
+            this.cache.Setup(c => c.Put(expectedKey, expectedPat));
 
             // The client will return a new PatToken upon creation request.
             this.client.Setup(c => c.GetActivePatsAsync(It.IsAny<CancellationToken>())).ReturnsAsync(activePats);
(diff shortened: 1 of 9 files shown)


src/AdoPat/PatCache.cs
```diff
@@ -33,23 +33,15 @@ public PatCache(IStorageWrapper storage)
...
             return token;
         }
 
-        /// <summary>
-        /// Put a <see cref="PatToken"/> into the cache. May overwrite existing values.
-        /// </summary>
-        /// <param name="key">The key for this entry.</param>
-        /// <param name="patToken">The value for this entry.</param>
+        /// <inheritdoc/>
         public void PutPat(string key, PatToken patToken)
```
shalinikhare27, line 45: I was re-thinking how this implementation would work in a concurrent execution scenario. If PutPat and ReadStorage were called simultaneously, this approach of writing and reading will result in mismatched data. Can this code ever be run in parallel? (concurrent tasks, threads, etc). If not, should we document that the implementation is not thread safe

src/AdoPat/PatOptions.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// <summary>
+        /// Gets the string representation of these PAT options.
+        /// </summary>
+        /// <returns>The string representation of these PAT options.</returns>
+        public override string ToString()
+        {
+            string output = $"{this.Organization} {this.DisplayName}";
+
+            foreach (var scope in this.Scopes)
+            {
```
shalinikhare27, line 36: If there can be many scopes then this code can benefit from the use of StringBuilder instead of creating new string allocation per loop iteration

src/AdoPat/IPatCache.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// <param name="key">The key for the target entry.</param>
+        /// <returns>The target value.</returns>
+        PatToken GetPat(string key);
+
+        /// <summary>
+        /// Put a <see cref="PatToken"/> into the cache. May overwrite existing values.
+        /// </summary>
+        /// <param name="key">The key for this entry.</param>
+        /// <param name="patToken">The value for this entry.</param>
+        void PutPat(string key, PatToken patToken);
```
shalinikhare27, line 25: Nit - The name sounds a bit off. PutPat, GetPat. I also notice you're using the term ```PatToken``` and Pat to refer to the same PAT.  Since T in PAT stands for token, should be rename PatToken to be PersonalAccessToken? 

As for these methods, my personal choice would be to either name them as ```GetPAT```, ```PutPAT``` or  simply ```GetToken```, ```PutToken``` since we are already in the context of PATCache here

src/AdoPat/PatClient.cs
```diff
@@ -32,12 +32,7 @@ public PatClient(ITokensHttpClientWrapper client)
             this.client = client;
```
shalinikhare27, line 32: I was revisiting the TokensHttpClientWrapper implementation and I noticed that you create an http client in the constructor and store it for all methods to reuse. I have seen this been suggested as an anti-pattern because the same http client is reused for multiple requests. Instead, we could get a client per request and allow VSSConnection to cache the http clients depending how many are being created. This is a similar pattern where we should use HttpClientFactory to create http clients instead of storing http clients in our own class. (good read - https://learn.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests)

Azure Devops samples use this pattern -
https://github.com/microsoft/azure-devops-dotnet-samples/blob/main/ClientLibrary/Samples/WorkItemTracking/WorkItemsSample.cs

shalinikhare27 (changes requested):
first look.

src/AdoPat/PatCache.cs
reillysiemens: When all is said and done, no, this code will not ever run in parallel. Almost the entire application will be wrapped in a system-wide process lock. Doing any of these operations in parallel would certainly cause issues, but there should be little to no need to do that.

That said, if there's a standardized way to document that C# classes/methods are not threadsafe then I'm more than happy to do that.

src/AdoPat/PatCache.cs
```diff
@@ -33,23 +33,15 @@ public PatCache(IStorageWrapper storage)
...
 
-        /// <summary>
-        /// Get a <see cref="PatToken"/> from the cache.
-        /// </summary>
-        /// <param name="key">The key for the target entry.</param>
-        /// <returns>The target value.</returns>
+        /// <inheritdoc/>
         public PatToken GetPat(string key)
         {
             PatToken token = null;
```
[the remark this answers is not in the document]
reillysiemens: Oh, that's good to know. `TryGetValue` uses the default value for a type if it is not found, so as long as the default value for `PatToken` is `null` then I think that would work. I'll test that.

src/AdoPat/PatOptions.cs
reillysiemens: Good point. I'll look into that. 😄

src/AdoPat/IPatCache.cs
reillysiemens: I only used `patToken` because the class is called `PatToken` and I don't have control over that since it's a type that comes from the Visual Studio ADO API library. Personally, I think it's a silly name as well. Seems kinda like ATM Machine or PIN Number.

As for the method names, I was under the impression that in C# Pascal case names like `Pat` were preferred to uppercase names like `PAT` even for initialisms or acronyms. I don't like `GetToken` or `PutToken` because there's enough ambiguity in this code base with multiple types of tokens.

However, as you say, we're already in the context of a `PatCache` here, so I'd happily go for `PatCache.Get()` or `PatCache.Put()`, which I think would be simpler.

src/AdoPat/PatOptions.cs
kyle-rader: Since we are joining on spaces - we can even skip a Stringbuilder and use `string.join` on the List/Array/Enumerable of org, name, scopes...

src/AdoPat/PatManager.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// </summary>
+        /// <param name="options">Options used to match the PAT in the cache or when creating/regenerating a new one.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
+        /// <returns>An Azure DevOps Personal Access Token.</returns>
+        public async Task<PatToken> GetPatAsync(
+            PatOptions options,
+            CancellationToken cancellationToken = default)
+        {
+            bool writeBack = false;
+            var pat = this.cache.GetPat(options.ToString());
```
kyle-rader, line 48: I have a slightly odd feeling about using ToString as the role of the `key`. ToString is designed for the string representation, typically a readable one. Here, the string representation you've chosen is motivated specifically to be the Key. In the future I could see someone easily changing the ToString to be more human readable, and it's not encoded or obvious that that output is actually what is used for the cache keys. 

I think I'd prefer the ToString to focus on it's role, if we even want one - for displaying on stdout the Pat, and specically add a `CacheKey` method to return the string to be used for CacheKeying.

src/AdoPat/PatManager.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            if (pat == null || await this.Inactive(pat, cancellationToken).ConfigureAwait(false))
+            {
+                var patTokencreateRequest = new PatTokenCreateRequest
+                {
+                    DisplayName = options.DisplayName,
+                    Scope = string.Join(" ", options.Scopes),
+                    ValidTo = this.now().AddDays(ValidToExtensionDays),
+                    AllOrgs = default,
```
kyle-rader, line 57: I think explicitly saying false, even though yes that's the default, is more clear here. 
```suggestion
                    AllOrgs = false,

```

kyle-rader: It is unclear from reading this code that it's a bool, if you use default.

kyle-rader: (granted Intellisense would fix this in VS to some degree).

src/AdoPat/IPatClient.cs
```diff
@@ -0,0 +1,41 @@
...
+        Task<IDictionary<Guid, PatToken>> GetActivePatsAsync(CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Creates a new PAT with a new 'valid to' date by replicating an existing one.
+        /// </summary>
+        /// <param name="patToken">An existing <see cref="PatToken"/>.</param>
+        /// <param name="validTo">The new expiration date for the regenerated token.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
+        /// <returns>A <see cref="Task"/> returning a <see cref="PatToken"/>.</returns>
+        Task<PatToken> RegeneratePatAsync(PatToken patToken, DateTime validTo, CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
kyle-rader: Should we have a "Re-<create>" method at all? This actually seems confusing to me since we explicitly are NOT regenerating existing PATs, we are always making new PATs. I think this should be also named `CreateAsync` and it's an overload. We can make a new PAT by either taking a new request params, or an existing pat, to queue off of.

src/AdoPat/PatManager.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
+        /// <returns>An Azure DevOps Personal Access Token.</returns>
+        public async Task<PatToken> GetPatAsync(
+            PatOptions options,
+            CancellationToken cancellationToken = default)
+        {
+            bool writeBack = false;
+            var pat = this.cache.GetPat(options.ToString());
+
+            if (pat == null || await this.Inactive(pat, cancellationToken).ConfigureAwait(false))
```
kyle-rader, line 50: In general this method feels a little awkward because we're first seeing if we have a cached pat. Then we check if it's null, or inactive, and if so Create a new one. Then we immeditaely check if that potentially new PAT is close to expiring. I feel like the logic should split or could be more streramlined here, such that we don't check if a PAT we know we just made is about to expire.

kyle-rader (changes requested):
I've left a lot of comments on specific areas of code and I'll leave one general one here. One large absent item from these Pat maangement classes is logging.

Have we thought about at which points of the Pat management logic, it might be helpful to be emitting DebugLog events? Could the PatManager take an `ILogger` ? (Not an `ILogger<T>`)

src/AdoPat/PatCache.cs
kyle-rader: The XML docs support a Remarks section, which is where you will often see a note on thread [non]safety.

src/AdoPat/IPatCache.cs
kyle-rader: Well that is silly.

src/AdoPat/PatClient.cs
reillysiemens: I don't create an HTTP client in the constructor, it's injected. But I _do_ store it for all methods to use, yes. I did that on purpose in order to re-use the connection for the lifetime of this program, which should be **very** short.

The "`client`" here is not directly an [`System.Net.Http.HttpClient`](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-7.0), but an instance of [`TokensHttpClientWrapper`](https://github.com/AzureAD/microsoft-authentication-cli/blob/93e2778c0eefdec3d9c9f7665b422c73b591a930/src/AdoPat/TokensHttpClientWrapper.cs), which itself is a thin wrapper around [`TokensHttpClient`](https://github.com/AzureAD/microsoft-authentication-cli/blob/93e2778c0eefdec3d9c9f7665b422c73b591a930/src/AdoPat/TokensHttpClientWrapper.cs#L26) by using `VssConnection.GetClient<TokensHttpClient>()`.

That client is then injected into the `PatManager` and used where necessary, which I _think_ matches the ["Typed Client"](https://learn.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests#implement-your-typed-client-classes-that-use-the-injected-and-configured-httpclient) pattern demonstrated in the article you linked.


Also, from reading that link you provided I see
> Therefore, HttpClient is intended to be instantiated once and reused throughout the life of an application.
> ...
> This can be a good solution for short-lived console apps or similar, that run a few times a day.

That's exactly my intention here. This whole thing should conclude in a couple hundred milliseconds at most and I think it should be reasonable to use a single socket during that time.

src/AdoPat.Test/PatManagerTest.cs
```diff
@@ -0,0 +1,215 @@
...
+    {
+        private const string Organization = "contoso";
+        private const string DisplayName = "Test PAT";
+        private const string Scope = "test.scope";
+
+        // This is a test token. A real value would be a much longer string.
+        private const string Token = "Test Token";
+
+        // The Unix Epoch is used as an obviously fake test time which occurs in the past and cannot accidentally be valid.
+        private readonly DateTime unixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
```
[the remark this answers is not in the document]
reillysiemens: Oh, yeah, I think it's `DateTime.UnixEpoch`, which I only learned about recently. This was the pattern I'd used for previous check-ins, so I just kept doing the same for consistency, but I can go back and fix them all.

src/AdoPat/IPatClient.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// </summary>
+    public interface IPatClient
+    {
+        /// <summary>
+        /// Creates a new PAT.
+        /// </summary>
+        /// <param name="patTokenCreateRequest">The PAT creation request.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
+        /// <returns>A <see cref="Task"/> returning a <see cref="PatTokenResult"/>.</returns>
+        Task<PatToken> CreatePatAsync(PatTokenCreateRequest patTokenCreateRequest, CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
reillysiemens: I'm happy to see this named `CreateAsync` instead of `CreatePatAsync`, but I cannot rename `PatToken`, because that's from the Azure DevOps library.

src/AdoPat/PatManager.cs
reillysiemens: Fair enough. More than happy to just rename this method.

src/AdoPat/PatManager.cs
reillysiemens: I thought about this and used `default` here because we're getting guidance from Azure DevOps folks to use `false` right now. By using `default` we will always use whatever they say is the default.

However, I don't suspect they'll ever change their minds about this and I'm happy to just use `false`.

src/AdoPat/IPatClient.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>
+        /// Gets all active PATs.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
+        /// <returns>A mapping of authorization ID to active PATs.</returns>
+        Task<IDictionary<Guid, PatToken>> GetActivePatsAsync(CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
reillysiemens: It specifically filters for active PATs, but I could go for `ListActiveAsync`. How's that?

src/AdoPat/IPatClient.cs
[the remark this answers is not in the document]
reillysiemens: This is documented in the [`PatClient`](https://github.com/AzureAD/microsoft-authentication-cli/blob/93e2778c0eefdec3d9c9f7665b422c73b591a930/src/AdoPat/PatClient.cs#L102-L114) code. We **are** regenerating a PAT according to the Azure DevOps description of that logical operation. That operation is a combination of API calls and is consistent with the UI that a user would see when managing their own PATs.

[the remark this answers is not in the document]
reillysiemens: That said, I'm happy to drop `Pat` from this method name and just go with `RegenerateAsync`, but I think we need to keep the `Regenerate` to be consistent with Azure DevOps.

src/AdoPat/PatCache.cs
reillysiemens: I'll add a remark for the `PatClient`, the `PatCache`, and the `PatManager` saying that none of them is intended to be threadsafe.

reillysiemens: Oh, and no, @kyle-rader, the x-plat file lock is not automatically used in `StorageWrapper`, but my intention is to make the entire `PatManager.GetPatAsync` method wrapped in an x-plat lock from `Main`.

src/AdoPat/PatManager.cs
reillysiemens: I did it this way because I wanted to minimize the number of times we called `PutPat`, but I can search for a less "awkward" implementation and see what I can come up with.

reillysiemens: The simplest thing I can think of is to drop the concept of `writeBack` for an `else if` and write to the cache in both arms of the conditional.
```csharp
var pat = this.cache.GetPat(options.ToString());

if (pat == null || await this.Inactive(pat, cancellationToken).ConfigureAwait(false))
{
    var patTokencreateRequest = new PatTokenCreateRequest
    {
        DisplayName = options.DisplayName,
        Scope = string.Join(" ", options.Scopes),
        ValidTo = this.now().AddDays(ValidToExtensionDays),
        AllOrgs = default,
    };
    pat = await this.client.CreatePatAsync(patTokencreateRequest, cancellationToken).ConfigureAwait(false);
    this.cache.PutPat(options.ToString(), pat);
}
else if (this.ExpiringSoon(pat))
{
    pat = await this.client.RegeneratePatAsync(
        pat,
        this.now().AddDays(ValidToExtensionDays),
        cancellationToken)
    .ConfigureAwait(false);
    this.cache.PutPat(options.ToString(), pat);
}

return pat;
```
How about that?

kyle-rader: Hello?

src/AdoPat/PatManager.cs
reillysiemens: Instead, I think I'm going to make it so that `PatClient.CreateAsync` only takes primitives and internally generates a `PatTokenCreateRequest` with `AllOrgs` set to `false` using the existing `PatClient.AllOrgs` constant.

kyle-rader-msft: For posterity - `default` here actually does _not_ use the `[DefaultValue(false)]` attribute defined on that class for that field, it's taking the default value for the `bool` type which happen to be the same here. The attribute declares the default value to use when instantiating an instance of that struct/class and the value is not given. So - we are going to pass `false` ourselves, since we know we never want to request an all orgs pat.

src/AdoPat/PatManager.cs
kyle-rader-msft: As discussed offline - I'm good with this. I just didn't like always checking a _new_ Pat if it's close to expiring.

src/AdoPat/IPatCache.cs
shalinikhare27: `PatCache.GetToken()` should work and be consistent. Even if we use `GetToken` in multiple places for different types of token fetching, the class names are distinct.

reillysiemens:
@kyle-rader-msft: I need to give more consideration to logging and would like to handle that in a follow-up PR.

src/AdoPat/PatClient.cs
shalinikhare27: I used a similar pattern in a code where many concurrent http requests were made and I ran into a network exhaustion problem. However, this shouldn't be a problem here.

shalinikhare27: We may want to use `HttpClientFactory` if we have similar code in the library since consumer can have high scale scenarios.

src/AdoPat/PatManager.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// <param name="options">Options used to match the PAT in the cache or when creating/regenerating a new one.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
+        /// <returns>An Azure DevOps Personal Access Token.</returns>
+        public async Task<PatToken> GetPatAsync(
+            PatOptions options,
+            CancellationToken cancellationToken = default)
+        {
+            var pat = this.cache.Get(options.CacheKey());
+
+            if (pat == null || await this.Inactive(pat, cancellationToken).ConfigureAwait(false))
```
shalinikhare27, line 52: [nit] we could store this in a bool as `var inactive = await this.Inactive(pat, cancellationToken).ConfigureAwait(false)' and use it here.

It will look elegant `if (pat == null || inactive)`

src/AdoPat/PatManager.cs
```diff
@@ -0,0 +1,88 @@
...
+                this.cache.Put(options.CacheKey(), pat);
+            }
+            else if (this.ExpiringSoon(pat))
+            {
+                pat = await this.client.RegenerateAsync(
+                    pat,
+                    this.now().AddDays(ValidToExtensionDays),
+                    cancellationToken)
+                .ConfigureAwait(false);
+                this.cache.Put(options.CacheKey(), pat);
```
shalinikhare27, line 69: We can't move this out as we only want to put in cache when pat is null or inactive. Or if it's expiring soon? Other we just get the Pat and return to user?

src/AdoPat/PatManager.cs
reillysiemens: I purposefully avoided that because I'm using the `pat == null` to short-circuit the conditional and avoid an unnecessary (and costly) API call if we don't need to make one because we didn't have the PAT in the cache.

src/AdoPat/PatManager.cs
```diff
@@ -0,0 +1,88 @@
...
+            {
+                pat = await this.client.CreateAsync(
+                    displayName: options.DisplayName,
+                    scope: string.Join(" ", options.Scopes),
+                    validTo: this.now().AddDays(ValidToExtensionDays),
+                    cancellationToken: cancellationToken)
+                .ConfigureAwait(false);
+                this.cache.Put(options.CacheKey(), pat);
+            }
+            else if (this.ExpiringSoon(pat))
```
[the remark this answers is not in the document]
reillysiemens: Same here. The `this.ExpiringSoon(pat)` call is unnecessary on a `null` PAT or on a PAT we just created.

src/AdoPat/PatManager.cs
reillysiemens: If there is a PAT in the cache and it's active then we don't need to rewrite the entry in the cache as that's just extra I/O overhead. In all other scenarios we _do_ need to overwrite the cache with an updated entry.

shalinikhare27 (dismissed):

shalinikhare27 (approved):

kyle-rader-msft (dismissed):
There are some style cop warnings but I don't want to block on them.

src/AdoPat/PatManager.cs
varshini-elayappen, line 52: Should we add comment on the scenarios where we won't have a pat in the cache? Creating a new pat if it is inactive is obvious but the scenarios where pat can be null isn't.

varshini-elayappen (approved):

src/AdoPat/PatManager.cs
reillysiemens: Good point. I'll see what I can do there.

shalinikhare27 (approved):

shalinikhare27 (approved):

varshini-elayappen (approved):

kyle-rader-msft (approved):

PR #213, opened 2023-03-01
First structure for split Silent and Interactive.

goagain:


src/MSALWrapper/AuthFlow/Web.cs
```diff
@@ -139,6 +107,63 @@ public async Task<AuthFlowResult> GetTokenAsync()
...
+                    "Get Token Silent",
+                    (cancellationToken) => this.pcaWrapper.GetTokenSilentAsync(this.scopes, account, cancellationToken),
+                    this.errors)
+                    .ConfigureAwait(false);
+                tokenResult.SetSilent();
+                return new AuthFlowResult(tokenResult, this.errors, this.GetType().Name);
+            }
+            catch (MsalUiRequiredException)
+            {
+                return null;
```
goagain, line 163: Shouldn't return null. Return an AuthFlowResult.

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -81,60 +82,21 @@ private enum GetAncestorFlags
         /// <returns>A <see cref="Task"/> of <see cref="TokenResult"/>.</returns>
         public async Task<AuthFlowResult> GetTokenAsync()
         {
-            IAccount account = await this.pcaWrapper.TryToGetCachedAccountAsync(this.preferredDomain)
-                 ?? PublicClientApplication.OperatingSystemAccount;
+            IAccount account = await this.GetCachedAccountAsync().ConfigureAwait(false);
```
kyle-rader-msft, line 85: @goagain we do not want to cahnge this AuthFlow. Broker Always has an account beacuse of the `?? PublicClientApplication.OperatingSystemAccount;` - this is very important and is how on Corp machines, we default to silently authenticating.

goagain: I moved the logic to `GetCachedAccountAsync`. I want to match the interface's definition.

PR #214, opened 2023-03-01
Create functional `TokenFetcher` orchestrator for AAD and Azure Devops

kyle-rader-msft:
# What are we adding?
This PR reincarnates `TokenFetcher` but this time as a focussed, [functional](https://en.wikipedia.org/wiki/Functional_programming), abstraction around creating auth flows, an authflow executor, and applying our locking logic around doing the auth process with all those parameters.

You may notice some long argument lists. A Data Clump is emerging, but I don't think it's worth trying to address it at this time until we have more need to. Early abstractions of data clump easily lead to wrong abstractions.

## `TokenFetcher`

The function `AccessTokenAsync` on on `TokenFetcher` extracts the parts of [`GetToken`](https://github.com/AzureAD/microsoft-authentication-cli/blob/9488ba2f5c86a0ba41aa4202335c633e7aca7747/src/AzureAuth/Commands/CommandAad.cs#L458) out of `CommandAad` into a re-usable class. We've done this first, because we evaluated the needs of `CommandAad`, `Ado.CommandToken` and `Ado.CommandPat` and all 3 have the same need for orchestrating this auth process and then using the result.

# What are we not adding?
Notice, this PR is purely additive for `TokenFetcher`s and The only changes to existing code are to the unflighted/unsed PatFromEnv class, which is now more aptly named. 

# Inspiration from Elixir and Rust

The abstraction for TokenFetcher's here _could_ use objects. But we don't really _need_ to so we've taken a functional approach.

## Results
To return more meaningly data types I've renamed both sub records of `PatFromEnv` and `TokenFetcher` to have a `Result` type. Something we can always return and ensure is not null, and represent the outcome of token fetching, mostly without null. These are declared _inside_ their respective static classes because they are intimately coupled, intentionally.


src/AzureAuth/Ado/PatFromEnv.cs
```diff
@@ -3,15 +3,42 @@
...
     /// <summary>
     /// A class for getting an ADO PAT from an <see cref="IEnv"/> or an AAD access token through MSAL.
     /// </summary>
-    public static class AdoToken
+    public static class PatFromEnv
     {
+        /// <summary>
+        /// The result for retrieving a PAT out of the environment.
+        /// </summary>
+        public record Result
```
kyle-rader-msft, line 24: This is renamed and moved from below. Nothing actually new here.

src/MSALWrapper/TokenFetcher.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// Run the authenticatuion process.
+        /// </summary>
+        /// <param name="logger">A logger.</param>
+        /// <param name="client">Client.</param>
+        /// <param name="tenant">Tenant.</param>
+        /// <param name="scopes">Scopes.</param>
+        /// <param name="mode">Modes.</param>
+        /// <param name="domain">Domain.</param>
+        /// <param name="prompt">Prompt Hint.</param>
+        /// <param name="timeout">Timeout.</param>
```
reillysiemens, line 47: Same here. Real descriptions, please.

src/MSALWrapper/TokenFetcher.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
+        public static async Task<Result> AccessTokenAsync(ILogger logger, Guid client, Guid tenant, IEnumerable<string> scopes, AuthMode mode, string domain, string prompt, TimeSpan timeout)
```
reillysiemens, line 49:
```suggestion
        public static async Task<Result> AccessTokenAsync(
            ILogger logger,
            Guid client,
            Guid tenant,
            IEnumerable<string> scopes,
            AuthMode mode,
            string domain,
            string prompt,
            TimeSpan timeout)
```

src/MSALWrapper/TokenFetcher.cs
```diff
@@ -0,0 +1,109 @@
...
+                authMode: mode,
+                clientId: client,
+                tenantId: tenant,
+                scopes: scopes,
+                preferredDomain: domain,
+                promptHint: prompt);
+
+            var executor = new AuthFlowExecutor(logger, authFlows, new StopwatchTracker(timeout));
+
+            List<AuthFlowResult> results = new List<AuthFlowResult>();
```
reillysiemens, line 62:
```suggestion
            List<AuthFlowResult> results = new List<AuthFlowResult>();
            var executor = new AuthFlowExecutor(logger, authFlows, new StopwatchTracker(timeout));
```

src/MSALWrapper/TokenFetcher.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// <param name="tenant">Tenant.</param>
+        /// <param name="scopes">Scopes.</param>
+        /// <param name="mode">Modes.</param>
+        /// <param name="domain">Domain.</param>
+        /// <param name="prompt">Prompt Hint.</param>
+        /// <param name="timeout">Timeout.</param>
+        /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
+        public static async Task<Result> AccessTokenAsync(ILogger logger, Guid client, Guid tenant, IEnumerable<string> scopes, AuthMode mode, string domain, string prompt, TimeSpan timeout)
+        {
+            var authFlows = AuthFlowFactory.Create(
```
kyle-rader-msft, line 51: This method is almost copied directly from `CommandAad.GetToken` with 1 important change - there's not yet a way to inject an IAuthFlow. We are tightly coupled with the AuthFlowFactory.

There is only 1 test requireing that and it's sort of an integration level test as all the components used here, aside from the lock, are unit tested already. I'm still pondering if we will want to add this ability. I'm leaning yes - and will do that when required to do by updating the CommandAad usage.

src/MSALWrapper/TokenFetcher.cs
```diff
@@ -0,0 +1,109 @@
...
+            List<AuthFlowResult> results = new List<AuthFlowResult>();
+
+            // When running multiple AzureAuth processes with the same resource, client, and tenant IDs,
+            // They may prompt many times, which is annoying and unexpected.
+            // Use Mutex to ensure that only one process can access the corresponding resource at the same time.
+            string resource = string.Join(' ', scopes);
+            string lockName = $"Local\\{resource}_{client}_{tenant}";
+
+            // First parameter initiallyOwned indicates whether this lock is owned by current thread.
+            // It should be false otherwise a dead lock could occur.
```
reillysiemens, line 71:
```suggestion
            // It should be false otherwise a deadlock could occur.

```

reillysiemens (changes requested):

src/AzureAuth/Ado/TokenFetcher.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    using AADTokenFetcher = Microsoft.Authentication.MSALWrapper.TokenFetcher;
+
+    /// <summary>
+    /// An abstraction around <see cref="MSALWrapper.TokenFetcher"/> specifically for Azure Devops.
+    /// </summary>
+    internal static class TokenFetcher
+    {
+        /// <summary>
+        /// Get an Azure Devops AAD access token.
```
goagain, line 20: nit
```suggestion
        /// Get an Azure DevOps AAD access token.
```

src/AzureAuth/Ado/TokenFetcher.cs
```diff
@@ -0,0 +1,46 @@
...
+            AuthMode mode,
+            string domain,
+            string prompt,
+            TimeSpan timeout)
+        {
+            return await AADTokenFetcher.AccessTokenAsync(
+                logger: logger,
+                client: new Guid(Constants.Client.VisualStudio),
+                tenant: new Guid(Constants.Tenant.Microsoft),
+                scopes: new[] { Constants.Scope.AzureDevopsDefault },
```
goagain, line 39: Is DevOps better than Devops?
```suggestion
                scopes: new[] { Constants.Scope.AzureDevOpsDefault },
```

goagain (approved):

reillysiemens (dismissed):

reillysiemens (approved):

goagain (approved):

src/MSALWrapper/TokenFetcher.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            // The first parameter 'initiallyOwned' indicates whether this lock is owned by current thread.
+            // It should be false otherwise a deadlock could occur.
+            using (Mutex mutex = new Mutex(initiallyOwned: false, name: lockName))
+            {
+                bool lockAcquired = false;
+                try
+                {
+                    // Wait for the other session to exit.
+                    lockAcquired = mutex.WaitOne(TimeSpan.FromMinutes(15)); // TODO: Inject mutex timeout?
```
reillysiemens, line 85: @shalinikhare27 mentioned in #218 that we can remove the `TODO` comment here and the timespan can become `const`.

reillysiemens (approved):

reillysiemens (approved):

shalinikhare27 (approved):

PR #216, opened 2023-03-02
Extract Prompt Hint to pure function

kyle-rader-msft:
This logic was already static but doesn't need to live on the CommandAad so before we re-use it in other command moving to it's own static class to hold the static function.

### DIFF
--- a/src/AzureAuth/PromptHint.cs
+++ b/src/AzureAuth/PromptHint.cs
@@ -1,0 +1,26 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.AzureAuth
+{
+    /// <summary>
+    /// Functions for ensuring a prompt hint is displayed with an AzureAuth prefix.
+    /// </summary>
+    public static class PromptHint
+    {
+        /// <summary>
+        /// Azure Auth Display name for prompt hints.
+        /// </summary>
+        public const string AzureAuth = "Azure Auth";
+
+        public static string Prefixed(string prompt)
+        {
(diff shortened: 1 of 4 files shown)


goagain (dismissed):
Actually, we don't have chance to see the prompt hint since we moved Web mode to browser.

kyle-rader-msft:
> Actually, we don't have chance to see the prompt hint since we moved Web mode to browser.

This is true at the moment, but we want to continue using it and accepting it bceause
1. It's good for telemetry distinction and,
2. Improving the UX of prompts when they happen is still an important feature desired by developers so it will be on the AzureAuth roadmap :)

shalinikhare27 (approved):

src/AzureAuth/PromptHint.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    /// <summary>
+    /// Functions for ensuring a prompt hint is displayed with an AzureAuth prefix.
+    /// </summary>
+    public static class PromptHint
+    {
+        /// <summary>
+        /// Azure Auth Display name for prompt hints.
+        /// </summary>
+        public const string AzureAuth = "Azure Auth";
```
reillysiemens, line 14: Should this actually have been `AzureAuth` this whole time? 🤔

reillysiemens (approved):

goagain (approved):

shalinikhare27 (approved):

reillysiemens (approved):

mvanchaa (approved):

kyle-rader-msft:
Poor GitHub Actions 
image

PR #217, opened 2023-03-02
Extract AuthFlowResult conversion to Telemetry into Extension Methods for Reuse

kyle-rader-msft:
The act of running auth flows and reporting their telemetry will also be needed by `CommandAad` and the 2 new commands, and even the `info` command once it gets the SID by trying Ado auth.

So this logic also is moved from CommandAad (along with the relevant tests) to their own static functions. In this case, there is a good type(s) to host these methods on so that we don't have to invent a static class name just to hold the methods.


src/AzureAuth.Test/AzureAuth.Test.csproj
```diff
@@ -8,8 +8,7 @@
     <TargetFramework>net6.0</TargetFramework>
     <IsPackable>false</IsPackable>
 
-    <!-- Stylecop warnings as errors flag for build to fail -->
-    <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
+    <TreatWarningsAsErrors>False</TreatWarningsAsErrors>
```
kyle-rader-msft, line 11: This still shows the issues, and you can fix them - but doesn't constantly break Live unit testing. I think we should downgrade to warnings in all projects.

src/AzureAuth.Test/CommandAadTest.cs
```diff
@@ -467,72 +467,6 @@ public void TestPromptHintPrefixWithoutOption()
                             .Should().BeEquivalentTo(PromptHintPrefix);
         }
 
-        /// <summary>
```
kyle-rader-msft, line 470: These tests are moved to AuthFlowResultExtensionsTest

src/AzureAuth.Test/AuthFlowResultExtensionsTest.cs
```diff
@@ -0,0 +1,143 @@
...
+    using FluentAssertions;
+
+    using Microsoft.Authentication.AzureAuth;
+    using Microsoft.Authentication.MSALWrapper;
+    using Microsoft.Identity.Client;
+    using Microsoft.IdentityModel.JsonWebTokens;
+
+    using NUnit.Framework;
+
+    public class AuthFlowResultExtensionsTest
```
kyle-rader-msft, line 17: These are not new, just moved from `CommandAadTest`

src/AzureAuth/AuthFlowResultExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            return eventData;
+        }
+
+        /// <summary>
+        /// Send telemetry events for n <see cref="AuthFlowResult"/> instance.
+        /// </summary>
+        /// <param name="results"><see cref="AuthFlowResult"/>s to send telemetry for.</param>
+        /// <param name="telemetryService">The <see cref="ITelemetryService"/> to use to send the events.</param>
+        public static void SendTelemetry(this IEnumerable<AuthFlowResult> results, ITelemetryService telemetryService)
```
kyle-rader-msft, line 56: This method has integration style test coverage in CommandAadTest.

src/AzureAuth.Test/AzureAuth.Test.csproj
shalinikhare27, line 11: Do we want to downgrade to warnings in non-test projects too?

shalinikhare27 (dismissed):

src/AzureAuth.Test/AuthFlowResultExtensionsTest.cs
```diff
@@ -0,0 +1,143 @@
...
+    {
+        private const string FakeToken = "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsInJoIjoieHh4IiwieDV0IjoieHh4Iiwia2lkIjoieHh4In0.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYXVkIjoiMTExMTExMTEtMTExMS0xMTExLTExMTEtMTExMTExMTExMTExIiwiaWF0IjoxNjE3NjY0Mjc2LCJuYmYiOjE2MTc2NjQyNzYsImV4cCI6MTYxNzY2ODE3NiwiYWNyIjoiMSIsImFpbyI6IllTQjBiM1JoYkd4NUlHWmhhMlVnYTJWNUlDTWtKVjQ9Iiwic2NwIjoidXNlcl9pbXBlcnNvbmF0aW9uIiwidW5pcXVlX25hbWUiOiJreXJhZGVyQG1pY3Jvc29mdC5jb20iLCJ1cG4iOiJreXJhZGVyQG1pY3Jvc29mdC5jb20iLCJ2ZXIiOiIxLjAifQ.bNc3QlL4zIClzFqH68A4hxsR7K-jabQvzB2EodgujQqc0RND_VLVkk2h3iDy8so3azN-964c2z5AiBGY6PVtWKYB-h0Z_VnzbebhDjzPLspEsANyQxaDX_ugOrf7BerQOtILWT5Vqs-A3745Bh0eTDFZpobmeENpANNhRE-yKwScjU8BDY9RimdrA2Z00V0lSliUQwnovWmtfdlbEpWObSFQAK7wCcNnUesV-jNZAUMrDkmTItPA9Z1Ks3NUbqdqMP3D6n99sy8DxQeFmbNQGYocYqI7QH24oNXODq0XB-2zpvCqy4T2jiBLgN_XEaZ5zTzEOzztpgMIWH1AUvEIyw";
```
reillysiemens, line 19: Did this not need to be removed from the other tests?

src/AzureAuth.Test/AzureAuth.Test.csproj
reillysiemens: Yes. I would like to delete StyleCop unless we can treat everything as a warning instead of a hard error.

reillysiemens (approved):

src/AzureAuth/AuthFlowResultExtensions.cs
mvanchaa: wow this looks neat!

src/AzureAuth/AuthFlowResultExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// <summary>
+    /// Extension methods for <see cref="AuthFlowResult"/>.
+    /// </summary>
+    public static class AuthFlowResultExtensions
+    {
+        /// <summary>
+        /// Convert to an <see cref="Office.Lasso.Telemetry.EventData"/>.
+        /// </summary>
+        /// <param name="result">Current <see cref="AuthFlowResult"/>.</param>
+        /// <returns>An <see cref="Office.Lasso.Telemetry.EventData"/> representng this <see cref="AuthFlowResult"/>.</returns>
```
mvanchaa, line 22: nit 
```suggestion
        /// <returns>An <see cref="Office.Lasso.Telemetry.EventData"/> representing this <see cref="AuthFlowResult"/>.</returns>

```

mvanchaa (dismissed):

src/AzureAuth.Test/AzureAuth.Test.csproj
kyle-rader-msft: I agree. And I'd like to do a sweep all at once, I'll revert here.

src/AzureAuth.Test/AuthFlowResultExtensionsTest.cs
kyle-rader-msft: Ah, it can be actually. There's still a lot of other setup in the CommandAadTest.

reillysiemens (dismissed):

src/AzureAuth/AuthFlowResultExtensions.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// without bringing a reference to the CLI framework into the <see cref="MSALWrapper"/>. This is not a boundary we want to cross.
+        /// Instead, here in AzureAuth, we can define the conversion using an extension method to house it rather than an arbitrary
+        /// static class to hold the same function. Being an extension here is purely syntactic sugar. The main preference is
+        /// around defining a "pure" function for this.
+        /// </remarks>
+        public static void SendTelemetry(this IEnumerable<AuthFlowResult> results, ITelemetryService telemetryService)
+        {
+            foreach (var result in results)
+            {
+                telemetryService.SendEvent($"authflow_{result.AuthFlowName}", result.EventData());
```
shalinikhare27, line 85: Is there a reason we reverted it back to not using `Parallel.ForEach`?

kyle-rader-msft: Yes, good question: 2 reasons:
1. The telemetry service is not explicitly thread safe, so this was never truly safe but it seems it was working.
2. We're about to update Lasso with the change to buffer events and only start 1 process, and that is also not thread safe, but also does not benefit from concurrency, because we would have to use a Concurrent Buffer, and the adds would be synchronized. The intention here was that each call is spawning a process, and so there was a benefit to doing this in paralle.

shalinikhare27 (approved):

PR #218, opened 2023-03-02
Extract auth mode interaction filtering into static methods for reuse

kyle-rader-msft:
Each Command that auths will need to respect our current logic around disabling interactive auth, so we migrate that business logic to it's own extenion on `IEnv` and the lopgic for how that choice impacts the `AuthMode` we want to use, we extract to `AuthModeExtensions`

### DIFF
--- a/src/AzureAuth/Ado/PatFromEnv.cs
+++ b/src/AzureAuth/Ado/PatFromEnv.cs
@@ -73,6 +73,5 @@ namespace Microsoft.Authentication.AzureAuth.Ado
                 Exists = false,
             };
         }
-
     }
 }
--- a/src/AzureAuth/AuthModeExtensions.cs
+++ b/src/AzureAuth/AuthModeExtensions.cs
@@ -1,0 +1,48 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.AzureAuth
+{
+    using Microsoft.Authentication.MSALWrapper;
+    using Microsoft.Office.Lasso.Interfaces;
+
+    using System.Collections.Generic;
+    using System.Linq;
+
+    /// <summary>
+    /// Extensions to <see cref="AuthMode"/>s.
+    /// </summary>
+    public static class AuthModeExtensions
+    {
+        /// <summary>
+        /// Filters a list of <see cref="AuthMode"/> for interaction returning a single aggregate <see cref="AuthMode"/>.
+        /// </summary>
+        /// <param name="authModes"><see cref="AuthMode"/>s to aggregate and filter.</param>
+        /// <param name="env">An <see cref="IEnv"/> to use.</param>
+        /// <returns>A single aggregate <see cref="AuthMode"/> filtered for interactivity.</returns>
+        public static AuthMode FilterInteraction(this IEnumerable<AuthMode> authModes, IEnv env)
+        {
+            if (InteractiveAuthDisabled(env))
+            {
+#if PlatformWindows
+                return AuthMode.IWA;
+#else
+                return 0;
+#endif
+            }
+
+            return authModes.Aggregate((a1, a2) => a1 | a2);
+        }
+
+        /// <summary>
+        /// Determines whether interactive auth is allowed or not.
+        /// </summary>
+        /// <param name="env">An <see cref="IEnv"/> to use.</param>
+        /// <returns>A boolean to indicate PCA is disabled.</returns>
+        public static bool InteractiveAuthDisabled(IEnv env)
+        {
+            return !string.IsNullOrEmpty(env.Get(EnvVars.NoUser)) ||
+                string.Equals("1", env.Get(EnvVars.CorextNonInteractive));
+        }
+    }
+}
--- a/src/AzureAuth/AuthModeExtensions.cs
+++ b/src/AzureAuth/AuthModeExtensions.cs
@@ -3,11 +3,11 @@
 
 namespace Microsoft.Authentication.AzureAuth
 {
-    using Microsoft.Authentication.MSALWrapper;
-    using Microsoft.Office.Lasso.Interfaces;
-
     using System.Collections.Generic;
     using System.Linq;
+
+    using Microsoft.Authentication.MSALWrapper;
+    using Microsoft.Office.Lasso.Interfaces;
 
     /// <summary>
     /// Extensions to <see cref="AuthMode"/>s.
@@ -20,7 +20,7 @@ namespace Microsoft.Authentication.AzureAuth
         /// <param name="authModes"><see cref="AuthMode"/>s to aggregate and filter.</param>
         /// <param name="env">An <see cref="IEnv"/> to use.</param>
         /// <returns>A single aggregate <see cref="AuthMode"/> filtered for interactivity.</returns>
-        public static AuthMode FilterInteraction(this IEnumerable<AuthMode> authModes, IEnv env)
+        public static AuthMode CombinedAuthMode(this IEnumerable<AuthMode> authModes, IEnv env)
         {
             if (InteractiveAuthDisabled(env))
             {
--- a/src/AzureAuth/AuthModeExtensions.cs
+++ b/src/AzureAuth/AuthModeExtensions.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Authentication.AzureAuth
         /// <returns>A single aggregate <see cref="AuthMode"/> filtered for interactivity.</returns>
         public static AuthMode CombinedAuthMode(this IEnumerable<AuthMode> authModes, IEnv env)
         {
-            if (InteractiveAuthDisabled(env))
+            if (env.InteractiveAuthDisabled())
             {
 #if PlatformWindows
                 return AuthMode.IWA;
@@ -32,17 +32,6 @@ namespace Microsoft.Authentication.AzureAuth
             }
 
             return authModes.Aggregate((a1, a2) => a1 | a2);
-        }
-
-        /// <summary>
-        /// Determines whether interactive auth is allowed or not.
-        /// </summary>
-        /// <param name="env">An <see cref="IEnv"/> to use.</param>
-        /// <returns>A boolean to indicate PCA is disabled.</returns>
-        public static bool InteractiveAuthDisabled(IEnv env)
-        {
-            return !string.IsNullOrEmpty(env.Get(EnvVars.NoUser)) ||
-                string.Equals("1", env.Get(EnvVars.CorextNonInteractive));
         }
     }
 }
--- a/src/AzureAuth/Commands/CommandAad.cs
+++ b/src/AzureAuth/Commands/CommandAad.cs
@@ -21,6 +21,8 @@ namespace Microsoft.Authentication.AzureAuth.Commands
     using Microsoft.Office.Lasso.Interfaces;
     using Microsoft.Office.Lasso.Telemetry;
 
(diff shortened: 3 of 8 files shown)


src/AzureAuth/Ado/PatFromEnv.cs
```diff
@@ -3,15 +3,42 @@
...
     /// <summary>
     /// A class for getting an ADO PAT from an <see cref="IEnv"/> or an AAD access token through MSAL.
     /// </summary>
-    public static class AdoToken
+    public static class PatFromEnv
     {
+        /// <summary>
+        /// The result for retrieving a PAT out of the environment.
+        /// </summary>
+        public record Result
```
goagain, line 24: Welcome to modern C#.

src/AzureAuth/AuthModeExtensions.cs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.AzureAuth
+{
+    using Microsoft.Authentication.MSALWrapper;
+    using Microsoft.Office.Lasso.Interfaces;
+
+    using System.Collections.Generic;
```
goagain, line 9: Lots of warnings here. We need to re-order them.

src/AzureAuth/AuthModeExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+
+            return authModes.Aggregate((a1, a2) => a1 | a2);
+        }
+
+        /// <summary>
+        /// Determines whether interactive auth is allowed or not.
+        /// </summary>
+        /// <param name="env">An <see cref="IEnv"/> to use.</param>
+        /// <returns>A boolean to indicate PCA is disabled.</returns>
+        public static bool InteractiveAuthDisabled(IEnv env)
```
shalinikhare27, line 42: This method can be `private` because it is only used here?

src/AzureAuth/AuthModeExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            return !string.IsNullOrEmpty(env.Get(EnvVars.NoUser)) ||
+                string.Equals("1", env.Get(EnvVars.CorextNonInteractive));
```
shalinikhare27, line 45: Does `CorextNonInteractive` have different values that it can set? If not, we can be consistent and just check for `!string.IsNullOrEmpty(env.Get(EnvVars.CorextNonInteractive))`

src/AzureAuth/AuthModeExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <param name="env">An <see cref="IEnv"/> to use.</param>
+        /// <returns>A single aggregate <see cref="AuthMode"/> filtered for interactivity.</returns>
+        public static AuthMode FilterInteraction(this IEnumerable<AuthMode> authModes, IEnv env)
+        {
+            if (InteractiveAuthDisabled(env))
+            {
+#if PlatformWindows
+                return AuthMode.IWA;
+#else
+                return 0;
```
shalinikhare27, line 30: Basically here we are saying non-interactive auth in Mac or Linus is impossible?

src/MSALWrapper/TokenFetcher.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            /// <summary>
+            /// Gets all the attempts made to authenticate.
+            /// </summary>
+            public List<AuthFlowResult> Attempts { get; init; }
+        }
+
+        /// <summary>
+        /// Run the authenticatuion process.
+        /// </summary>
```
shalinikhare27, line 39: These should be more meaningful.

src/MSALWrapper/TokenFetcher.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            // First parameter initiallyOwned indicates whether this lock is owned by current thread.
+            // It should be false otherwise a dead lock could occur.
+            using (Mutex mutex = new Mutex(initiallyOwned: false, name: lockName))
+            {
+                bool lockAcquired = false;
+                try
+                {
+                    // Wait for the other session to exit.
+                    lockAcquired = mutex.WaitOne(TimeSpan.FromMinutes(15)); // TODO: Inject mutex timeout?
```
shalinikhare27, line 78: Move this to a `const`.

shalinikhare27, line 78: can be removed.

src/AzureAuth/AuthModeExtensions.cs
mvanchaa: `CorextNonInteractive` can only be set to `1` if they want to disable interactivity.

src/AzureAuth/AuthModeExtensions.cs
mvanchaa: I think for unit testing it's made public.

src/AzureAuth/AuthModeExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// </summary>
+    public static class AuthModeExtensions
+    {
+        /// <summary>
+        /// Filters a list of <see cref="AuthMode"/> for interaction returning a single aggregate <see cref="AuthMode"/>.
+        /// </summary>
+        /// <param name="authModes"><see cref="AuthMode"/>s to aggregate and filter.</param>
+        /// <param name="env">An <see cref="IEnv"/> to use.</param>
+        /// <returns>A single aggregate <see cref="AuthMode"/> filtered for interactivity.</returns>
+        public static AuthMode FilterInteraction(this IEnumerable<AuthMode> authModes, IEnv env)
```
reillysiemens, line 23: I think there's room to refactor/rename this method, but let's stick to mechanical transformations for now and keep this as `CombinedAuthMode`.

src/MSALWrapper/TokenFetcher.cs
```diff
@@ -0,0 +1,109 @@
...
+            public AuthFlowResult Success => this.Attempts.FirstOrDefault(result => result.Success);
+
+            /// <summary>
+            /// Gets all the attempts made to authenticate.
+            /// </summary>
+            public List<AuthFlowResult> Attempts { get; init; }
+        }
+
+        /// <summary>
+        /// Run the authenticatuion process.
```
mvanchaa, line 38: nit 
```suggestion
        /// Run the authentication process.

```

src/AzureAuth/AuthModeExtensions.cs
reillysiemens: With public client auth, yes. That's correct. This code is also being lifted as-is from `CombinedAuthMode`, so this hasn't changed at all.

I did suggest to @kyle-rader-msft that this `0` be given a name like `None` though (better done in a separate PR).

src/AzureAuth/AuthModeExtensions.cs
reillysiemens: Good question. I think it may have been `public` before because of tests?

src/AzureAuth/AuthModeExtensions.cs
reillysiemens: I think it operates on `0` or `1` and is not in our control. I think this is the only correct way to do this check.

src/AzureAuth/Commands/CommandAad.cs
```diff
@@ -382,29 +364,19 @@ public int OnExecute()
             // Small bug in Lasso - Add does not accept a null IEnumerable here.
             this.eventData.Add("settings_scopes", this.authSettings.Scopes ?? new List<string>());
 
-            if (this.InteractiveAuthDisabled())
+            if (ModeExtensions.InteractiveAuthDisabled(this.env))
```
reillysiemens, line 367: Could this be an `IEnv` extension method?
```csharp
this.env.InteractiveAuthDisabled()
```

mvanchaa (approved):

reillysiemens (changes requested):

src/MSALWrapper/TokenFetcher.cs
reillysiemens: The `TimeSpan.FromMinutes(15)`?

src/AzureAuth/AuthModeExtensions.cs
kyle-rader-msft: I'll add a comment about this.

src/MSALWrapper/TokenFetcher.cs
reillysiemens: This is part of #214.

reillysiemens: This is part of #214.

src/MSALWrapper/TokenFetcher.cs
reillysiemens: @mvanchaa: This is from #214.

src/AzureAuth/AuthModeExtensions.cs
kyle-rader-msft: Moved to a constant to give it an explicit name

src/MSALWrapper/TokenFetcher.cs
kyle-rader-msft: Updated in #214

reillysiemens (approved):

kyle-rader-msft:
Aha! Good job Linux CI. This is totally uncaught on Windows in Visual Studio because it doesn't event both to type-check the non windows compilation block :(

src/MSALWrapper/MSALWrapper.csproj
```diff
@@ -15,9 +15,6 @@
     <PackageTags>MSAL;authentication</PackageTags>
 
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-
-    <!-- Stylecop warnings as errors flag for build to fail -->
-    <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
```
shalinikhare27, line 20: My concern about removing them as errors is to deviate from stylecop completely and defies the purpose of adding them for code consistency. I get the annoyance though.

shalinikhare27 (approved):

src/MSALWrapper/MSALWrapper.csproj
kyle-rader-msft: In this case it's actually required. StyleCop doesn't know how to handle `record` declarations and it explodes. We could disable the single Rule, but I don't think we're at risk of deviating since GitHub does a good job of showing all the warnings inline in the diff. It's much more effective to be able to write code with the warnings, get it working, and then clean it up :)

shalinikhare27: Agreed

shalinikhare27 (dismissed):

src/AzureAuth/Commands/CommandAad.cs
```diff
@@ -21,6 +21,8 @@ namespace Microsoft.Authentication.AzureAuth.Commands
     using Microsoft.Office.Lasso.Interfaces;
     using Microsoft.Office.Lasso.Telemetry;
 
+    using ModeExtensions = Microsoft.Authentication.AzureAuth.AuthModeExtensions;
```
reillysiemens, line 24: Is this still needed?

kyle-rader-msft: Yes because we have a naming conflict between us and MSALWrapper.

goagain (approved):

reillysiemens (approved):

PR #219, opened 2023-03-02
[Urgent] Fix GitHub workflow.

goagain:
It is very likely the previous PR integrating CodeQL breaks our build and test workflow on main branch.
According to [Error: 403 "Resource not accessible by integration" when using Dependabot](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/troubleshooting-the-codeql-workflow?learn=code_security_actions#error-403-resource-not-accessible-by-integration-when-using-dependabot), we need to add `security_events: write` scope to the job.

### DIFF
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -12,7 +12,8 @@ jobs:
       actions: read
       contents: read
       security-events: write
-
+      statuses: write
+      
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


kyle-rader-msft (changes requested):
Actually, we will need the same update in the release yaml - I think we should change both at the same time here.

mvanchaa (dismissed):

goagain:
> Actually, we will need the same update in the release yaml - I think we should change both at the same time here.

I have made same changes in #199 because it hasn't been merged.

kyle-rader-msft:
Aha Ty!

kyle-rader-msft (approved):

shalinikhare27 (approved):

PR #223, opened 2023-03-07
Separate Silent and Interactive authflows.

goagain:
In the past, we always try silent auth even if there is no cached account. Separate the auth flow can reduce `MSALUiRequiredException`.


kyle-rader-msft:
I've still got a lot of thoughts on this, and I think it will be worth taking the time to do a team brainstorm about this issue.

goagain:
> I've still got a lot of thoughts on this, and I think it will be worth taking the time to do a team brainstorm about this issue.

I'd love to discuss and pair-code for these changes.

PR #224, opened 2023-03-08
Update Readme.md for release 0.7.1

goagain:


reillysiemens (approved):

kyle-rader-msft (changes requested):
Since we have found a remaining thread mutex bug in 0.7.1 I'd like to fix that and release 0.7.2 first before updating the readme to point at a 0.7.x release.

PR #227, opened 2023-03-10
Add a Token Formatter

kyle-rader-msft:
This creates a `TokenFormatter` to handle Authorization header formatting and base64 encoding.


src/AzureAuth.Test/Ado/TokenFormatterTest.cs
```diff
@@ -0,0 +1,55 @@
...
+    using Microsoft.Authentication.AzureAuth.Ado;
+
+    using NUnit.Framework;
+
+    internal class TokenFormatterTest
+    {
+        [Test]
+        public void Header_Bearer()
+        {
+            TokenFormatter.HeaderBearer("foobar").Should().Be("Authorization: Bearer foobar");
```
goagain, line 17: Somehow, I wonder if a string extension method `.WithHeaderBearer` is more fluent.

kyle-rader-msft: This is an option that might fit nicely with a recommendation Tucker had while pairing.

src/AzureAuth/Ado/TokenFormatting.cs
```diff
@@ -0,0 +1,64 @@
...
+
+namespace Microsoft.Authentication.AzureAuth.Ado
+{
+    using System;
+    using System.Text;
+
+    /// <summary>
+    /// <see cref="string"/> extension methods for formatting tokens.
+    /// </summary>
+    public static class TokenFormatting
```
goagain, line 12:
```suggestion
    public static class TokenFormatting
    public static class TokenFormattingExtensions
```

For extionsion class, I would like to add `extensions` as a suffix.

goagain (approved):

shalinikhare27 (approved):

src/AzureAuth.Test/Ado/TokenFormatterTest.cs
reillysiemens: Now if only I could remember what that recommendation was... 😆

reillysiemens (approved):

goagain (approved):

PR #228, opened 2023-03-10
Initial work to get `ado token` command working. 

kyle-rader-msft:
This adds a functioning version of the `ado token` command.

There is refactoring cleanup work that will come after this to extract the common logic into a more testable flow that's re-usable by the `ado pat` and `aad` commands.

### DIFF
--- a/src/AzureAuth/Commands/Ado/CommandToken.cs
+++ b/src/AzureAuth/Commands/Ado/CommandToken.cs
@@ -28,32 +28,6 @@ For use by short-lived processes. More info at https://aka.ms/AzureAuth")]
         private const string OutputOptionDescription = "How to print the token. One of [token, header, headervalue].\nDefault: token";
 
         /// <summary>
-        /// Format a PAT based on <paramref name="output"/>.
-        /// </summary>
-        /// <param name="value">The PAT value.</param>
-        /// <param name="output">The output mode.</param>
-        /// <returns>The formatted PAT value ready for printing.</returns>
-        public static string FormatPat(string value, OutputMode output) => output switch
-        {
-            OutputMode.Token => value,
-            OutputMode.Header => TokenFormatter.HeaderBasic(value),
-            OutputMode.HeaderValue => TokenFormatter.HeaderBasicValue(value),
-        };
-
-        /// <summary>
-        /// Format an Access Token based on <paramref name="output"/>.
-        /// </summary>
-        /// <param name="value">The access token.</param>
-        /// <param name="output">The output mode.</param>
-        /// <returns>The formatted Access Toekn ready for printing.</returns>
-        public static string FormatAccessToken(string value, OutputMode output) => output switch
-        {
-            OutputMode.Token => value,
-            OutputMode.Header => TokenFormatter.HeaderBearer(value),
-            OutputMode.HeaderValue => TokenFormatter.HeaderBearerValue(value),
-        };
-
-        /// <summary>
         /// The available Token Formats.
         /// </summary>
         public enum OutputMode
@@ -71,6 +45,19 @@ For use by short-lived processes. More info at https://aka.ms/AzureAuth")]
             /// <summary> TODO </summary>
             HeaderValue,
         }
+
+        /// <summary>
+        /// Format a PAT based on <paramref name="output"/>.
+        /// </summary>
+        /// <param name="value">The PAT value.</param>
+        /// <param name="output">The output mode.</param>
+        /// <returns>The formatted PAT value ready for printing.</returns>
+        public static string FormatToken(string value, OutputMode output, Authorization scheme) => output switch
+        {
+            OutputMode.Token => value,
+            OutputMode.Header => value.AsHeader(scheme),
+            OutputMode.HeaderValue => value.AsHeaderValue(scheme),
+        };
 
         [Option(OutputOption, OutputOptionDescription, CommandOptionType.SingleValue)]
         private OutputMode Output { get; set; } = OutputMode.Token;
@@ -104,7 +91,7 @@ For use by short-lived processes. More info at https://aka.ms/AzureAuth")]
             if (pat.Exists)
             {
                 logger.LogDebug($"Using PAT from env var {pat.EnvVarSource}");
-                logger.LogInformation(FormatPat(pat.Value, this.Output));
+                logger.LogInformation(FormatToken(pat.Value, this.Output, Authorization.Basic));
                 return 0;
             }
 
@@ -116,12 +103,11 @@ For use by short-lived processes. More info at https://aka.ms/AzureAuth")]
                 prompt: AzureAuth.PromptHint.Prefixed(this.PromptHint),
                 timeout: TimeSpan.FromMinutes(this.Timeout));
 
-
             var authflow = authResult.Success;
             if (authflow != null)
             {
                 logger.LogDebug($"Acquired AAD AT via {authflow.AuthFlowName} in {authflow.Duration.TotalSeconds:0.00} sec");
-                logger.LogInformation(FormatAccessToken(authflow.TokenResult.Token, this.Output));
+                logger.LogInformation(FormatToken(authflow.TokenResult.Token, this.Output, Authorization.Bearer));
                 return 0;
             }
 
--- a/src/AzureAuth/Commands/Ado/CommandToken.cs
+++ b/src/AzureAuth/Commands/Ado/CommandToken.cs
@@ -28,7 +28,7 @@ For use by short-lived processes. More info at https://aka.ms/AzureAuth")]
         private const string OutputOptionDescription = "How to print the token. One of [token, header, headervalue].\nDefault: token";
 
         /// <summary>
-        /// Format a PAT based on <see cref="OutputMode"/>.
+        /// Format a PAT based on <paramref name="output"/>.
         /// </summary>
         /// <param name="value">The PAT value.</param>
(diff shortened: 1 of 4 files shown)


src/AzureAuth/Commands/Ado/CommandToken.cs
```diff
@@ -3,59 +3,127 @@
...
         [Option(CommandAad.TimeoutOption, CommandAad.TimeoutHelpText, CommandOptionType.SingleValue)]
-        public double Timeout { get; set; } = CommandAad.GlobalTimeout.TotalMinutes;
+        private double Timeout { get; set; } = CommandAad.GlobalTimeout.TotalMinutes;
+
+        [Option(CommandAad.PromptHintOption, CommandAad.PromptHintHelpText, CommandOptionType.SingleValue)]
+        private string PromptHint { get; set; }
 
         /// <summary>
-        /// Gets or sets the prompt hint.
+        /// Format a PAT based on <paramref name="output"/>.
```
reillysiemens, line 68: I don't think this is strictly being used for formatting PATs anymore. Might want to check the docs for this method.

reillysiemens (approved):

src/AzureAuth/Commands/Ado/CommandToken.cs
kyle-rader-msft: Will fix this in a follow up PR, good call out.

kyle-rader-msft: #229

mvanchaa (dismissed):

src/AzureAuth/Commands/Ado/CommandToken.cs
```diff
@@ -3,59 +3,127 @@
...
+            /// Raw Token
+            /// </summary>
+            Token,
+
+            /// <summary>
+            /// Authorization http header value.
+            /// </summary>
+            Header,
+
+            /// <summary> TODO </summary>
```
shalinikhare27, line 45: Let's add the description here.

src/AzureAuth/Commands/Ado/CommandToken.cs
```diff
@@ -3,59 +3,127 @@
...
+
     /// <summary>
     /// ADO Command for using either an ADO PAT or acquiring an AAD Access Token.
     /// </summary>
     [Command("token", Description = @"Get a PAT from env vars or an AAD AT for Azure Devops.
 For use by short-lived processes. More info at https://aka.ms/AzureAuth")]
     public class CommandToken
     {
+        private const string OutputOption = "--output";
+        private const string OutputOptionDescription = "How to print the token. One of [token, header, headervalue].\nDefault: token";
```
shalinikhare27, line 28: [nit] - This description is confusing to me. It starts with "How to print the token" which seems like you will provide steps to print them. Can we simple say `Token options: [token , header, headervalue].` something?

src/AzureAuth/Commands/Ado/CommandToken.cs
```diff
@@ -3,59 +3,127 @@
...
 
         /// <summary>
-        /// Gets or sets the prompt hint.
+        /// Format a PAT based on <paramref name="output"/>.
         /// </summary>
-        [Option(CommandAad.PromptHintOption, CommandAad.PromptHintHelpText, CommandOptionType.SingleValue)]
-        public string PromptHint { get; set; }
+        /// <param name="value">The PAT value.</param>
+        /// <param name="output">The output mode.</param>
+        /// <param name="scheme">The <see cref="Authorization"/> scheme to use.</param>
```
shalinikhare27, line 72: Scheme means basic or bearer?

src/AzureAuth/Commands/Ado/CommandToken.cs
```diff
@@ -3,59 +3,127 @@
...
+                mode: this.AuthModes.Combine().PreventInteractionIfNeeded(env),
+                domain: this.Domain,
+                prompt: AzureAuth.PromptHint.Prefixed(this.PromptHint),
+                timeout: TimeSpan.FromMinutes(this.Timeout));
+
+            var authflow = authResult.Success;
+            if (authflow != null)
+            {
+                logger.LogDebug($"Acquired AAD AT via {authflow.AuthFlowName} in {authflow.Duration.TotalSeconds:0.00} sec");
+                logger.LogInformation(FormatToken(authflow.TokenResult.Token, this.Output, Authorization.Bearer));
```
shalinikhare27, line 112: If `authflow` is not null, we can completely ensure we have received an access token?

src/AzureAuth/Commands/Ado/CommandToken.cs
```diff
@@ -3,59 +3,127 @@
...
+                return 0;
+            }
+
+            logger.LogError($"Failed to find a PAT and authenticate to ADO.");
+            foreach (var attempt in authResult.Attempts)
+            {
+                logger.LogError($"{attempt.AuthFlowName} failed after {attempt.Duration.TotalSeconds:0.00} sec. Error count: {attempt.Errors.Count}");
```
shalinikhare27, line 119: Do we need to capture this? We know that if land here all authflows have failed. The timeout for each is already hard coded in code.

src/AzureAuth/Commands/Ado/CommandToken.cs
```diff
@@ -3,59 +3,127 @@
...
+                foreach (var e in attempt.Errors)
+                {
+                    logger.LogError($"  {e.Message}");
```
shalinikhare27, line 122:
```suggestion
                    logger.LogError($"{e.Message}");
```

shalinikhare27 (changes requested):
There are bunch of stylecop warnings that I see in every PR from `src/AdoPat/PatOptions.cs`. Can we fix them in a follow up PR as it's distracting.

src/AzureAuth/Commands/Ado/CommandToken.cs
kyle-rader-msft: (Ah SonarLint would catch this!)

src/AzureAuth/Commands/Ado/CommandToken.cs
kyle-rader-msft: Is this more clear?
```
Options:

  --output        Controls token output format. Options: [token, header, headervalue].
                  Default: token
```

src/AzureAuth/Commands/Ado/CommandToken.cs
kyle-rader-msft: Correct, and doc comment has a link to the Authorization scheme Enum.

src/AzureAuth/Commands/Ado/CommandToken.cs
kyle-rader-msft: Yes, because `authflow` is declared just above as the `authResult.Success` - which is either a succesful AuthFlowResult or null.

src/AzureAuth/Commands/Ado/CommandToken.cs
kyle-rader-msft: This spacing is intentional. It uses the indent to "hang" the actual errors off the header printed first.

src/AzureAuth/Commands/Ado/CommandToken.cs
kyle-rader-msft: It's just just printing more information about each flow. This exact logic I plan to move into a common orchestrator. This way more of the logging can be consolidated into 1 place, and the indivudal auth flows can be more concerned with executing and returning results. The caller (this level of code) can then take more ownership over how to print those results.

kyle-rader-msft:
> There are bunch of stylecop warnings that I see in every PR from `src/AdoPat/PatOptions.cs`. Can we fix them in a follow up PR as it's distracting.

I've opened a PR to migrate to Sonar Lint, and can then work on cleaning up it's warnings. StyleCop inherently isn't going to work because it's not updating for modern versions of C#.

src/AzureAuth/Commands/Ado/CommandToken.cs
kyle-rader-msft: @shalinikhare27 - does that make sense and are you sated?

src/AzureAuth/Commands/Ado/CommandToken.cs
shalinikhare27: yeah works!

src/AzureAuth/Commands/Ado/CommandToken.cs
```diff
@@ -3,59 +3,127 @@
...
+        {
+            /// <summary>
+            /// Raw Token
+            /// </summary>
+            Token,
+
+            /// <summary>
+            /// Authorization http header.
+            /// </summary>
+            Header,
```
reillysiemens, line 43: Maybe make _all_ of these summaries one line?

reillysiemens (approved):

shalinikhare27 (approved):

src/AzureAuth/Commands/Ado/CommandToken.cs
kyle-rader-msft: Good suggestion - will do in a follow up PR.

src/AzureAuth/Commands/Ado/CommandToken.cs
kyle-rader-msft: Great, I will update that in a follow up PR.

Issue #233, opened 2023-03-11
AzureAuth 0.7.2 doesn't work on Mac


goagain:
AzureAuth 0.7.2 can't create the mutex on Mac because of the mutex path contains a `/` which is not allowed on Unix.

PR #234, opened 2023-03-11
[urgent]Remove scopes in the lock name because scopes may contains `/`

goagain:
# Summary
AzureAuth 0.7.2 can't create the mutex on Mac because of the mutex path contains a `/` which is not allowed on Unix.

# Details
The string concatenation operation causes the concatenated string to contain `/`. This causes it to be recognized as an unavailable folder on the Unix platform, resulting in unavailable services. We need to fix this bug immediately and release version 0.7.3.


goagain:
Resolve #233

kyle-rader-msft (approved):
Generally, I would prefer to see a unit test that proves the bug exists. Given that we have a few pending refactors, 1 of which will actually make all the mutex code covered by unit tests, I am okay with manually testing this fix. I have just confirmed the issue on Mac in 0.7.2 and that this is the cause.

I'm also okay expanding the scope of the lock given most usage does not use a different set of scopes.

Haard30 (approved):
+1 to unit tests for this. Also, it would be great if description can briefly talk about solution and what all testing we have done to ensure this works.

PR #235, opened 2023-03-11
Upload CodeQL result to artifact.

goagain:
Microsoft security policy requires us to keep the <abbr title="Static Code Analysis">SCA</abbr> result for at least 12 months. Uploading the artifact can meet the requirement.

Based on the analysis log, we can find out the file location is 
`/home/runner/work/microsoft-authentication-cli/results/csharp.sarif`

```
  Processing sarif files: ["/home/runner/work/microsoft-authentication-cli/results/csharp.sarif"]
  Uploading results
  Successfully uploaded results
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
     - name: Setup Python
       uses: actions/setup-python@v3
       with:
@@ -52,7 +52,7 @@ jobs:
             os: windows-latest
     steps:
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -74,10 +74,17 @@ jobs:
     - name: Perform CodeQL Analysis
       uses: github/codeql-action/analyze@v2
 
+    - name: Upload CodeQL Analysis Results
+      uses: actions/upload-artifact@v3
+      with:
+        name: azureauth-static-code-analysis-${{ matrix.runtime }}.sarif
+        path: results/csharp.sarif
```
kyle-rader-msft, line 81: This is going to override the results from each job, with the last one winning won't it? Could we instead upload all 3 results from each platform job - similar to how you've used the runtime in the name for this step?

kyle-rader-msft (changes requested):

.github/workflows/release.yml
goagain: Because we specified the argument `name`, there will be separate files like
- azureauth-static-code-analysis-windows-x64.sarif
- azureauth-static-code-analysis-osx-x64.sarif
- azureauth-static-code-analysis-osx-arm.sarif
And ideally it won't override our current artifacts.

kyle-rader-msft: Oh I see the artifact is uploaded as the name attribute, not the path. Got it!

kyle-rader-msft (approved):
Generally, with our finicky pipelines are, I recommend not updating unrelated task versions. In the event that breaks something, we want to be able to revert only that change, easily with Git. But merging changes makes that harder. This is a fairly low risk one so I'll approve, but in general - separate PRs for unrelated changes.

shalinikhare27 (approved):

PR #238, opened 2023-03-16
Move locking logic into AuthFlowExecutor

kyle-rader-msft:
This PR deletes `TokenFetcher` in favor of housing the locking code directly in `AuthFlowExecutor` since that class is what holds the behavior we always want to have a lock on.

This also brings all the locking code under unit test coverage, which we previously did not have. 

I apologize this PR has a larger diff than I would like, most of the diff is in moving or removing code. The majority of meaningful updates are in the AuthFlowExeccutorTest which now tests the locking.

As we had to update every one of those tests, I also noticed we had some redundant assertions that I felt we really did not need, so some tests are removed instead of updated.


src/AzureAuth/Ado/TokenFetcher.cs
reillysiemens, line 1: I don't have any particular feedback here. Just wanted to test the new feature that allows you to comment on whole files instead of just lines within a file.

PR #239, opened 2023-03-16
Update Readme to 0.7.3

### LINKED ISSUE #174: The option --cache=".IdentityService/msal_.cache ..." is not a valid absolute file path.
seanadams@Seans-MacBook-Pro ~ % azureauth
The --resource field or the --scope field is required.
The --client field is required.
The --tenant field is required.
The option --cache=`.IdentityService/msal_.cache` or environment varable AZUREAUTH_CACHE=`` is not a valid absolute file path.

Fresh install of 0.6.0, opened new terminal and ran azureauth with no params and got the error.

goagain:
Although AzureAuth 0.7.3 has been fully deployed, I am going to update Readme in advance. Leaving the readme at a deprecated version is dangerous, and we can see 0.7.2 has been downloaded more than 3k times.

resolve #174

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # Microsoft Authentication CLI
 
 [![Tests](https://img.shields.io/github/actions/workflow/status/AzureAd/microsoft-authentication-cli/.github/workflows/dotnet-test.yml?branch=main&style=for-the-badge&logo=github)](https://github.com/AzureAD/microsoft-authentication-cli/actions/workflows/dotnet-test.yml)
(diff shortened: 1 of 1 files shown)


kyle-rader-msft (changes requested):
We also need to update the badges at the top of the file :)

goagain:
> We also need to update the badges at the top of the file :)

Somehow, I missed them.

kyle-rader-msft (approved):
🚀

reillysiemens (approved):

PR #241, opened 2023-03-17
[1/3] 🏝️ Isolate and test Mutex in standalone class `Locked` 

kyle-rader-msft:
Originally, we had this locking code in `CommandAad`, then we moved it into the reincarnated `TokenFetcher`. But even there, it's a static method with no direct ties to token fetching.

Furthermore, it was not unit tested. By pulling that code into it's own public static method, we can more clearly unit test it. 

This PR adds a new `Locked` static class housing the one `Execute` method that takes an action and executes it, safely, while holding a Mutex.

The unit test coverage has gone from 0% to 100% :) Including the "Abandoned Mutex" scenario which ended up being pretty straightforward.

Follow up PRs to come to replace the existing locking code with this.

### DIFF
--- a/src/MSALWrapper.Test/LockedTest.cs
+++ b/src/MSALWrapper.Test/LockedTest.cs
@@ -48,41 +48,30 @@ namespace MSALWrapper.Test
         public void Execute_TimeOut()
         {
             var lockName = "short task times out while long task is running";
+            var shortTime = TimeSpan.FromMilliseconds(1);
 
-            // this int will be used to signal across threads
-            int lockAcquired = 0;
-            var oneMs = TimeSpan.FromMilliseconds(0.25);
+            // AutoResetEvent gives you a nice signaling primitive.
+            AutoResetEvent hasLock = new AutoResetEvent(false);
+            AutoResetEvent assertionMade = new AutoResetEvent(false);
 
-            Func<int> longFunc = () => Locked.Execute(this.mockLogger.Object, lockName, TenSec, async () =>
+            Func<int> longFunc = () => Locked.Execute(this.mockLogger.Object, lockName, TenSec, () =>
             {
-                // signal that we have acquired the lock
-                Interlocked.Increment(ref lockAcquired);
-
-                // wait for the signal that our test has made it's assertion
-                while (lockAcquired == 1)
-                {
-                    await Task.Delay(oneMs);
-                }
-
-                return 42;
+                hasLock.Set();
+                assertionMade.WaitOne();
+                return Task.FromResult(42);
             });
 
-            Action subject = () => Locked.Execute(this.mockLogger.Object, lockName, oneMs, () => Task.FromResult(0));
+            Action subject = () => Locked.Execute(this.mockLogger.Object, lockName, shortTime, () => Task.FromResult(0));
 
             Task<int> longTask = Task.Run(longFunc);
 
-            // wait for our signal
-            while (lockAcquired == 0)
-            {
-                Thread.Sleep(oneMs);
-            }
-
+            hasLock.WaitOne();
             subject.Should().Throw<TimeoutException>();
 
             // Release the long Task by signaling we've made our assertion.
             // This prevents us from abandoning the longTask which has the lock and would not actually release
             // the Mutex correctly. This could be a problem if another test accidentally re-used a lock name.
-            Interlocked.Increment(ref lockAcquired);
+            assertionMade.Set();
             longTask.Result.Should().Be(42);
         }
     }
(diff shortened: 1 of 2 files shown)


goagain (dismissed):

src/MSALWrapper/Locked.cs
```diff
@@ -0,0 +1,68 @@
...
+            // It should be false otherwise a deadlock could occur.
+            using (Mutex mutex = new Mutex(initiallyOwned: false, name: lockName))
+            {
+                bool lockAcquired = false;
+                try
+                {
+                    // Wait for other sessions to exit.
+                    lockAcquired = mutex.WaitOne(maxLockWaitTime);
+                }
+                catch (AbandonedMutexException)
```
shalinikhare27, line 40: Does `Abandon_The_Mutex` test cover when this exception might be thrown?

src/MSALWrapper.Test/LockedTest.cs
```diff
@@ -0,0 +1,100 @@
...
+                assertionMade.WaitOne();
+                return Task.FromResult(42);
+            });
+
+            Action subject = () => Locked.Execute(this.mockLogger.Object, lockName, shortTimeOut, () => Task.FromResult(0));
+
+            // Start longFunc, and wait for the lock to be acquired
+            Task<int> longTask = Task.Run(longFunc);
+            hasLock.WaitOne();
+            subject.Should().Throw<TimeoutException>();
```
shalinikhare27, line 69: We can also assert the timeout message here?

src/MSALWrapper/Locked.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Extensions.Logging;
+
+    /// <summary>
+    /// A method for executing a <see cref="Func{T}"/> under an inter-process lock.
+    /// </summary>
+    public static class Locked
```
shalinikhare27, line 14: Just curious why did we name it `Locked` instead of `Lock`?

src/MSALWrapper/Locked.cs
kyle-rader-msft: Yes it does 🎊🎊🎊🎊, which is why I am so excited. It took 2 hours and I would not have gotten here without @reillysiemens who actually paired with me the entire time on all 3 of these PRs :)

src/MSALWrapper/Locked.cs
kyle-rader-msft: Partially English preference. I usually find myself saying "we want to do this in a locked fashion", and partially to be different from `Lock` because `lock` (lowercase) is a keyword.

src/MSALWrapper.Test/LockedTest.cs
kyle-rader-msft: Good call out, we can make an assertion about the Exception Message here.

src/MSALWrapper.Test/LockedTest.cs
```diff
@@ -0,0 +1,100 @@
...
+            {
+                m.WaitOne();
+                hasLock.Set();
+            }).Start();
+
+            // Once lock is acquired, we can start our second task which waits for the lock.
+            hasLock.WaitOne();
+            int subject = Locked.Execute(this.mockLogger.Object, lockName, tenSeconds, () => Task.FromResult(13));
+            subject.Should().Be(13);
+        }
```
shalinikhare27, line 98: We can assert the `logger` warning here then:
```
logMock.Verify(log => log.Log(
                LogLevel.Warning,
                It.IsAny<EventId>(),
                It.IsAny<It.IsAnyType>(),
                It.IsAny<Exception>(),
                (Func<It.IsAnyType, Exception, string>)It.IsAny<object>()),
                Times.Exactly(1));

            logMock.VerifyNoOtherCalls();
```

This will ensure it was logged only once and there were no other logger calls.

kyle-rader-msft: This is a great example! I'm glad to see that this does work, but I'm hesitant to use it here, because the 3rd PR in this series adds log assertions that are stronger than what I think we can do with the mock, at least in a readable fashion. @shalinikhare27 do you mind waiting for log asserions until PR #243 ?

shalinikhare27 (approved):

src/MSALWrapper.Test/LockedTest.cs
kyle-rader-msft: Assertion added.

reillysiemens (approved):

PR #242, opened 2023-03-17
[2/3] 🦺 Assert lock names are safe

kyle-rader-msft:
Add tests to `Locked` to cover the scenario that caused one of the extra releases we had to ship (AzureAuth failed on mac using an invalid character in the lock name).

### DIFF
--- a/src/MSALWrapper/Locked.cs
+++ b/src/MSALWrapper/Locked.cs
@@ -4,11 +4,9 @@
 namespace Microsoft.Authentication.MSALWrapper
 {
     using System;
-    using System.IO;
     using System.Linq;
     using System.Security.Cryptography;
     using System.Text;
-    using System.Text.Unicode;
     using System.Threading;
     using System.Threading.Tasks;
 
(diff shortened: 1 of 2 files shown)


src/MSALWrapper/Locked.cs
```diff
@@ -64,5 +80,20 @@ public static T Execute<T>(ILogger logger, string lockName, TimeSpan maxLockWait
...
             return result;
         }
+
+        /// <summary>
+        /// Create a safe lockname.
+        /// </summary>
+        /// <param name="lockName">The lock name.</param>
+        /// <param name="visibility">The Mutex <see cref="Visibility"/>.</param>
+        /// <returns>A safe lock name derived from <paramref name="visibility"/> and <paramref name="lockName"/>.</returns>
+        public static string LockName(string lockName, Visibility visibility)
```
shalinikhare27, line 90: @reillysiemens I see this method 'public' too instead of 'private'. I bring up because you had a comment on my PR for making methods private wherever possible. It's public here for unit tests I believe and works for me. Same reason I had public in my PR. Do let me know if I missed something in my PR. :)

shalinikhare27 (approved):

reillysiemens (approved):

PR #244, opened 2023-03-18
🪛 fix test assertion and namespaces

kyle-rader-msft:
The namespaces of a few tests got out of sync beceause the default was actually still based on the project name. 

Fixed a test assertion that added in my prior 2/3 PR, but an update to the 1/3 changes the expected value.

### DIFF
--- a/src/MSALWrapper.Test/LockedTest.cs
+++ b/src/MSALWrapper.Test/LockedTest.cs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-namespace MSALWrapper.Test
+namespace Microsoft.Authentication.MSALWrapper.Test
 {
     using System;
     using System.Threading;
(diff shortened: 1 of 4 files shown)


shalinikhare27 (approved):

Haard30 (approved):

src/MSALWrapper.Test/LockedTest.cs
```diff
@@ -81,7 +79,7 @@ public void Execute_TimeOut()
             // Start longFunc, and wait for the lock to be acquired
             Task<int> longTask = Task.Run(longFunc);
             hasLock.WaitOne();
-            subject.Should().Throw<TimeoutException>().WithMessage("The application did not gain access to the lock named 'short task times out while long task is running' in the expected time.");
+            subject.Should().Throw<TimeoutException>().WithMessage("The application did not gain access to the lock named 'Local\\203b4357cac4279e7ed3492be002ca1bfbf3dbd87f87a88bdc7e1690a41266ee' in the expected time.");
```
goagain, line 82: What does the hash value mean?

PR #245, opened 2023-03-18
[1/x] 🐸 Create `AuthFlows` class.

kyle-rader-msft:
This class aims to be ever so slightly more knowledgeable than an `IEnumerable<IAuthFlow>`. We pass `IEnumerable<IAuthFlow>` to an `AuthFlowExecutor`, after getting one from an `AuthFlowFactory`, but then inside the executor, we aren't going to know what this set of authflows should use for a lock name. 

If they just had an Id, to name them, we could use to lock with ;)

`AuthFlows` is also an `IEnumerable<IAuthFlow>`, which allows this diff to be fairly small. Everywhere we previously iterated through a list of authflows, we still can. The only difference is in how we create them.

Next PR: `AuthFlowFactory` can be effectively be turned into the constructor for `AuthFlows`.


src/MSALWrapper/AuthFlow/AuthFlows.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    /// <summary>
+    /// A list of <see cref="IAuthFlow"/> to be executed by an <see cref="AuthFlowExecutor"/>.
+    /// </summary>
+    public class AuthFlows : IEnumerable<IAuthFlow>
+    {
+        /// <summary>
+        /// The Id of this group of auth flows.
+        /// </summary>
+        public readonly string Id;
```
shalinikhare27, line 18:
```suggestion
        public readonly string id;
```

readonly is generally lowercase

src/MSALWrapper/AuthFlow/AuthFlows.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AuthFlows"/> class.
+        /// </summary>
+        /// <param name="client">The client for these <paramref name="authFlows"/>.</param>
+        /// <param name="tenant">The tenant for these <paramref name="authFlows"/>.</param>
+        /// <param name="authFlows">An <see cref="IEnumerable{IAuthFlow}"/> of auth flows.</param>
+        public AuthFlows(Guid client, Guid tenant, IEnumerable<IAuthFlow> authFlows)
+        {
+            this.Id = $"{client}_{tenant}";
```
shalinikhare27, line 30: I am a bit unclear on why are we creating ids this way?

src/MSALWrapper.Test/AuthFlow/AuthFlowsTest.cs
```diff
@@ -0,0 +1,43 @@
...
+        public void AuthFlows_Has_Id()
+        {
+            var client = Guid.NewGuid();
+            var tenant = Guid.NewGuid();
+            var subject = new AuthFlows(client, tenant, new List<IAuthFlow>());
+            subject.Id.Should().Be($"{client}_{tenant}");
+        }
+
+        [Test]
+        public void Is_Iterable()
```
shalinikhare27, line 35: We can probably write a test on `Is_Iterable` when `IEnumerable<IAuthFlow> flows` is not an empty list.

src/MSALWrapper/AuthFlow/AuthFlows.cs
reillysiemens: @kyle-rader-msft mentioned in the description
> If they just had an Id, to name them, we could use to lock with

This matches how we're currently creating lock names, so I assume the intention is to use an ID as the lock name.

src/MSALWrapper/TokenFetcher.cs
```diff
@@ -59,14 +59,14 @@ public record Result
             string prompt,
             TimeSpan timeout)
         {
-            var authFlows = AuthFlowFactory.Create(
+            var authFlows = new AuthFlows(client, tenant, AuthFlowFactory.Create(
```
reillysiemens, line 62: We're now passing `client` and `tenant` in two places here... Why doesn't the `AuthFlowFactory` return an `AuthFlows`?

src/MSALWrapper/AuthFlow/AuthFlows.cs
kyle-rader-msft: That will trigger a StyleCop/Sonar warning - `readonly` does not impact casing preferences (in StyleCop, or generally). The rule here is that `public` properties are `camelCase`, and `protected`, `internal`, and `private` properties are `PascalCase`.

src/MSALWrapper/AuthFlow/AuthFlows.cs
kyle-rader-msft: Here, I'm specifically taking what we use today for the lock name around a set of auth flows. I could rename `Id` to LockName to make this more explicit. (In other words, I did not invent this here, it's taken from TokenFetcher)

src/MSALWrapper.Test/AuthFlow/AuthFlowsTest.cs
kyle-rader-msft: The only reason I wrote this test was to really TDD it into existence. The fact that a LINQ query on an IEnumerable<T> works, tells us it's interable. This is a case of where adding more tests on it being iterable are testing DotNet more than our own code, and we don't need to test DotNet here :).

src/MSALWrapper/AuthFlow/AuthFlows.cs
shalinikhare27: Aha! I see. Perhaps, we could rename it to be `LockName` and be more explicit.

src/MSALWrapper/AuthFlow/AuthFlows.cs
shalinikhare27: Since this is public then should be camelCase?

src/MSALWrapper/AuthFlow/AuthFlows.cs
reillysiemens: From the perspective of calling code that might make sense, but this class is unaware of any locking. It just so happens that the unique identifier for this set of auth flows can be used as the name of the lock.

src/MSALWrapper/AuthFlow/AuthFlows.cs
kyle-rader-msft: Oh I'm sorry I'm wrote that backwards, it's the reverse, all our public properties are PascalCase, internal+ are camelCase.

src/MSALWrapper/TokenFetcher.cs
reillysiemens: As long as that's something we're consciously doing now and it gets adjusted later, I'm OK with it in this PR.

src/MSALWrapper/AuthFlow/AuthFlows.cs
reillysiemens: I'm perfectly happy with `id` or `Id`, but I think `iD` is a bit strange. And at any rate, I think the camel casing applies to words and ID is one "word", so `id` seems reasonable by that measure. Does StyleCop really give a warning if we do `id`?

reillysiemens (approved):

shalinikhare27 (dismissed):

reillysiemens (dismissed):

src/MSALWrapper/AuthFlow/AuthFlows.cs
kyle-rader-msft: Yes it does. I tried.

src/MSALWrapper/AuthFlow/AuthFlows.cs
kyle-rader-msft: Yeah, and the fact we're exposing it indicates to me we're finding the right abstraction within AuthFlows, and that it should house AuthFlowFactory's method as a ctor, and AuthFlowExecutor as a method. Then the LockName, becomes private since it's an implementation detail. Being public here is a signal that as is, AuthFlows is a leaky abstraction. So, follow UP coming.

shalinikhare27 (approved):

kyle-rader-msft:
After further exploration, I'm not convinced this was the right abstraction to attempt to make next. The goal was to enable the AuthFlowExecutor to directly handle locking it's operations. The AuthFlowExecutor and AuthFlowFactory both hold static methods that are well tested, and I don't think makes sense to try to merge them overall. There's maybe a cleaner way we could split apart the logic from the Executor which is doing 2 big things: Execute each AuthFlow, AND, keep track of the time, and manage bailing from an overall timeout and log about it. That logic on it's own represents roughly half of the Executor and that logic, similar to the `Locked` interface could be separated into it's own generic class for doing an operation overall that may timeout, while each individual authFlow also can timeout. But that's more than we can take on right now - so instead, I'm thinking we just move forward with utilizing `Locked` in TokenFetcher, which can easily own the lock name and locking, but uses a well tested method to do so.

I'll leave it there, and move on with the last abstraction we need in AzureAuth for "get a token, log results, send telemetry, and give back token result."

PR #247, opened 2023-03-21
[1/n] 📜 Move common in memory logger setup to shared method.

kyle-rader-msft:
A lot of our tests in MSALWrapper use an `ILogger`. Today, they go through a lot of boilerplate in order to create a simple in-memory logger through a `Microsoft.DependencyInjection.ServiceCollection`. This adds a level of overhead we don't really need or take advantage of in most of our tests.

We can make a fully named `ILogger` by reflecting on the calling class's full name.
*This prevent copy and paste errors* because the same code pasted into another test automatically creates a logger named after that test. We don't currently use this name in our tests anywhere. But now it's always correct 😄 accurate.

Special thanks to @reillysiemens and @shalinikhare27 for having a deep discussion on `ILogger`s and testing with them ❤️ !

This will be a familiar diff in this PR:
image

### DIFF
--- a/TestHelper/MemoryLogger.cs
+++ b/TestHelper/MemoryLogger.cs
@@ -1,0 +1,31 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.TestHelper
+{
+    using Microsoft.Extensions.Logging;
+
+    using NLog;
+    using NLog.Config;
+    using NLog.Targets;
+
+    public static class MemoryLogger
+    {
+        public static (ILogger<T> logger, MemoryTarget logTarget) Create<T>()
+        {
+            var loggerFactory = new NLog.Extensions.Logging.NLogLoggerFactory();
+            var logger = loggerFactory.CreateLogger<T>();
+
+            // Setup in memory logging target with NLog - allows making assertions against what has been logged.
+            var loggingConfig = new LoggingConfiguration();
+            var logTarget = new MemoryTarget("memory_target");
+            logTarget.Layout = "${message}";
+            loggingConfig.AddTarget(logTarget);
+            loggingConfig.AddRuleForAllLevels(logTarget);
+
+            // Set the Config
+            LogManager.Configuration = loggingConfig;
+            return (logger, logTarget);
+        }
+    }
+}
--- a/TestHelper/MemoryLogger.cs
+++ b/TestHelper/MemoryLogger.cs
@@ -3,20 +3,26 @@
 
 namespace Microsoft.Authentication.TestHelper
 {
-    using Microsoft.Extensions.Logging;
-
     using NLog;
     using NLog.Config;
     using NLog.Targets;
 
+    using System.Diagnostics;
+
     public static class MemoryLogger
     {
-        public static (ILogger<T> logger, MemoryTarget logTarget) Create<T>()
+        public static (Extensions.Logging.ILogger logger, MemoryTarget logTarget) Create()
         {
-            var loggerFactory = new NLog.Extensions.Logging.NLogLoggerFactory();
-            var logger = loggerFactory.CreateLogger<T>();
+            // Use reflection to get the name of the class which declares our caller.
+            // Expected to be the name of a Test class.
+            StackFrame frame = new StackFrame(1);
+            var method = frame.GetMethod();
+            var type = method.DeclaringType;
 
-            // Setup in memory logging target with NLog - allows making assertions against what has been logged.
+            var loggerFactory = new NLog.Extensions.Logging.NLogLoggerFactory();
+            var logger = loggerFactory.CreateLogger(type.FullName);
+
+            // Setup in memory logging target with NLog
             var loggingConfig = new LoggingConfiguration();
             var logTarget = new MemoryTarget("memory_target");
             logTarget.Layout = "${message}";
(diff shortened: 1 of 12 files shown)


goagain (approved):
Don’t perform it in front of the Muggles!

Just kidding. I also saw some changes not related to ILogger. Does it come from another Pull Request?

kyle-rader-msft:
> Don’t perform it in front of the Muggles!
> 
> Just kidding. I also saw some changes not related to ILogger. Does it come from another Pull Request?

@goagain , which changes did you see that were unrelated? Every change here is intended to be directly scoped to using the `MemoryLogger.Create` method to setup the loggers for unit tests, or a required update from that change.

goagain:
> > Don’t perform it in front of the Muggles!
> > Just kidding. I also saw some changes not related to ILogger. Does it come from another Pull Request?
> 
> @goagain , which changes did you see that were unrelated? Every change here is intended to be directly scoped to using the `MemoryLogger.Create` method to setup the loggers for unit tests, or a required update from that change.

Sorry never mind. I saw some changes were related to AuthFlow, but it was actually ILogger<AuthFlow>.

TestHelper/MemoryLogger.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    using System.Diagnostics;
+
+    public static class MemoryLogger
+    {
+        public static (Extensions.Logging.ILogger logger, MemoryTarget logTarget) Create()
+        {
+            // Use reflection to get the name of the class which declares our caller.
+            // Expected to be the name of a Test class.
+            StackFrame frame = new StackFrame(1);
```
shalinikhare27, line 18: I was looking into the `StackFrame` constructor and it was `Initializes a new instance of the [StackFrame](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.stackframe?view=net-6.0) class that corresponds to a frame above the current stack frame.` We choose this value to be 1. Basically this is the reflection we use to get the name of our class?

TestHelper/TestHelper.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.1</TargetFramework>
```
shalinikhare27, line 4: Can it not target `net6.0`?

TestHelper/MemoryLogger.cs
```diff
@@ -0,0 +1,38 @@
...
+    using NLog.Config;
+    using NLog.Targets;
+
+    using System.Diagnostics;
+
+    public static class MemoryLogger
+    {
+        public static (Extensions.Logging.ILogger logger, MemoryTarget logTarget) Create()
+        {
+            // Use reflection to get the name of the class which declares our caller.
```
shalinikhare27, line 16: Is using reflection a good idea here? Any design implications?

reillysiemens (approved):

kyle-rader-msft:
Oh, @goagain I did fix the bin/win/azureauth script - that is unrelated, it was just broken.

src/MSALWrapper.Test/AuthFlow/BrokerTest.cs
```diff
@@ -44,39 +46,19 @@ public class BrokerTest
         [SetUp]
         public void Setup()
         {
-            // Setup in memory logging target with NLog - allows making assertions against what has been logged.
-            var loggingConfig = new NLog.Config.LoggingConfiguration();
-            this.logTarget = new MemoryTarget("memory_target");
-            loggingConfig.AddTarget(this.logTarget);
-            loggingConfig.AddRuleForAllLevels(this.logTarget);
+            (this.logger, this.logTarget) = MemoryLogger.Create();
```
shalinikhare27, line 49: we are updating each test with this, which is great! We simplified it by moving memory log setup into it's own thing. However, none of these tests are asserting the log messages though. A lot of tests do have logs and we should probably assert those.

TestHelper/MemoryLogger.cs
kyle-rader-msft: Yep! The (1) goes up 1 stack frame, which is generally going to be a Unit test class.

TestHelper/TestHelper.csproj
kyle-rader-msft: Discussed offline, I chose netstardard2.1 out of old habits for creating a Library project. While this has greater backwards compatibility - we do not need backwards compatibility, and this limits the C# language version, so I will update this to `net6.0` so we have one language version across all projects.

goagain: TestHelper is a library, so it is not necessary to target net6.0

TestHelper/MemoryLogger.cs
kyle-rader-msft: Somewhat covered in the description - I chose reflection here after initially taking a generic paramter `T` to create `logger` as an `ILogger<T>` to set the name. But then I copy and pasted the wrong class arg into all the tests. Since no assertions use it yet, no tests failed, but in the future the name would not have been expected or made sense. So with reflection, you can copy and paste the same line and it does the right thing in each test automatically :)

src/MSALWrapper.Test/AuthFlow/BrokerTest.cs
kyle-rader-msft: Correct, I've intentionally avoided changing the _behavior_ of any tests to keep this PR scoped to a `refactor`, and not an `enhancement` of the tests themselves. We could have made the same enhancements already, this is more about the deduplication of the logging setup.

PR #249, opened 2023-03-21
Move 📦 TestHelper into src and target net6.0

kyle-rader-msft:
I accidentally had TestHelper in the root instead of `src`. Also followup from #248 to target net6.0 since we do not need the backwards compat of netstandard 2.1.

We also hit a version downgrade error, which I think we can avoid by enabling the use of dotbnet lockfiles for all our projects.


shalinikhare27 (approved):

goagain (approved):

src/TestHelper/TestHelper.csproj
```diff
@@ -1,13 +1,13 @@
...
   <PropertyGroup>
-    <TargetFramework>netstandard2.1</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.1.4" />
-    <PackageReference Include="NLog.Extensions.Logging" Version="1.7.3" />
+    <PackageReference Include="NLog.Extensions.Logging" Version="1.6.3" />
```
goagain, line 10: Is there any specific reason to downgrade the library?

PR #250, opened 2023-03-23
[1/n] 📦 Move TokenFetcher method behind interface.

kyle-rader-msft:
When re-creating this `TokenFetcher` as a static class, there wasn't a clear use case for mocking it's usage. We now have a clear use case for mocking it's usage in wanting to TDD the common orchestrator in the AzureAuth project into existence, asserting that the right logging, and telemetry calls are made.

This change keeps the logic exactly the same, but puts it behind an instance method so that a mock object can be created for it. Note that we specifically avoid state though! We can get the benefits of a static function by not maintaining any state.

## How is this going to get used?
in AzureAuth[.Test], we can now create the token orchestrator (call it a `melon` for now). The `melon` will take an `ITokenFetcher`, an `ILogger`, and an `ITelemetryService`. We will be able to register `TokenFetcher` in the main DI service collection as well as the `melon`, and allow top-level commands to have these components auto-injected into them, also making them more testable.

### DIFF
--- a/src/AzureAuth/Ado/TokenFetcher.cs
+++ b/src/AzureAuth/Ado/TokenFetcher.cs
@@ -32,7 +32,7 @@ namespace Microsoft.Authentication.AzureAuth.Ado
             string prompt,
             TimeSpan timeout)
         {
-            return AADTokenFetcher.AccessToken(
+            return new AADTokenFetcher().AccessToken(
                 logger: logger,
                 client: new Guid(Constants.Client.VisualStudio),
                 tenant: new Guid(Constants.Tenant.Microsoft),
--- a/src/MSALWrapper/ITokenFetcher.cs
+++ b/src/MSALWrapper/ITokenFetcher.cs
@@ -1,0 +1,39 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.MSALWrapper
+{
+    using System;
+    using System.Collections.Generic;
+    using Microsoft.Authentication.MSALWrapper.AuthFlow;
+    using Microsoft.Extensions.Logging;
+
+    /// <summary>
+    /// An interface for acquiring an AAT Access Token.
+    /// </summary>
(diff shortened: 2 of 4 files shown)


src/MSALWrapper/ITokenFetcher.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Microsoft.Authentication.MSALWrapper
+{
+    using System;
+    using System.Collections.Generic;
+    using Microsoft.Authentication.MSALWrapper.AuthFlow;
+    using Microsoft.Extensions.Logging;
+
+    /// <summary>
+    /// An interface for acquiring an AAT Access Token.
```
shalinikhare27, line 12:
```suggestion
    /// An interface for acquiring an AAD Access Token.
```

src/AzureAuth/Ado/TokenFetcher.cs
```diff
@@ -32,7 +32,7 @@ internal static class TokenFetcher
             string prompt,
             TimeSpan timeout)
         {
-            return AADTokenFetcher.AccessToken(
+            return new AADTokenFetcher().AccessToken(
```
reillysiemens, line 35: I'm having a hard time locating the corresponding change to the `AADTokenFetcher`. Am I missing something?

kyle-rader-msft: `AADTokenFetcher` is a type alias define above for `MSALWrapper.TokenFetcher` :)

shalinikhare27: ` using AADTokenFetcher = Microsoft.Authentication.MSALWrapper.TokenFetcher;`

shalinikhare27: Okay that makes sense now. I am excited to us able to TDD the common orchestrator in the AzureAuth project.

shalinikhare27 (approved):

reillysiemens (approved):

PR #251, opened 2023-03-23
[2/n] 🍉 Create AuthOrchestrator (renamed in next PR)

kyle-rader-msft:
## `AuthOrchestrator` - feel free to ponder a better name. 

This PR's goal is to handle the boilerplate of the new class and it's UnitTests. The following PR will add to those tests defining it's behavior. 

This PR creates `AuthDriver` and the setup for `AuthDriverTest`. The idea behind `AuthDriver` is that this is the "Orchestrator" for doing
* auth to AAD
* logging about it
* sending telemtry about it
* Codify our MSALWrapper Usage in 1 place. 

# How is this different from `TokenFetcher`?

`TokenFetcher` only knows about MSAL, while this class _does know about Lasso, IEnv, And ITelemetryService_. The behavior is not about doing auth, but enforcing how we call TokenFetcher to do auth, and interpreting the results. 

The next PR will focus on TDD-ing it's behavior into existence.

### DIFF
--- a/src/AzureAuth.Test/AuthDriverTest.cs
+++ b/src/AzureAuth.Test/AuthDriverTest.cs
@@ -16,7 +16,7 @@ namespace AzureAuth.Test
     using NLog.Targets;
     using NUnit.Framework;
 
-    internal class MelonTest
+    internal class AuthDriverTest
     {
         private ILogger logger;
         private MemoryTarget logTarget;
@@ -33,7 +33,7 @@ namespace AzureAuth.Test
             this.tokenFetcher = new Mock<ITokenFetcher>(MockBehavior.Strict);
         }
 
-        public Melon Subject() => new Melon(this.logger, this.env.Object, this.telemetryService.Object, this.tokenFetcher.Object);
+        public AuthDriver Subject() => new AuthDriver(this.logger, this.env.Object, this.telemetryService.Object, this.tokenFetcher.Object);
 
         [Test]
         public void Contructor_Works()
--- a/src/AzureAuth.Test/AuthDriverTest.cs
+++ b/src/AzureAuth.Test/AuthDriverTest.cs
@@ -16,6 +16,8 @@ namespace AzureAuth.Test
     using NLog.Targets;
     using NUnit.Framework;
 
+    using System;
+
     internal class AuthDriverTest
     {
         private ILogger logger;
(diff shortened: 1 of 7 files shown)


shalinikhare27:
The idea behind AuthDriver is that this is the "Orchestrator" - I would vote for 'AuthOrchestrator'.

src/AzureAuth.Test/AuthDriverTest.cs
```diff
@@ -0,0 +1,44 @@
...
+        [SetUp]
+        public void SetUp()
+        {
+            (this.logger, this.logTarget) = MemoryLogger.Create();
+            this.env = new Mock<IEnv>(MockBehavior.Strict);
+            this.telemetryService = new Mock<ITelemetryService>(MockBehavior.Strict);
+            this.tokenFetcher = new Mock<ITokenFetcher>(MockBehavior.Strict);
+        }
+
+        public AuthDriver Subject() => new AuthDriver(this.logger, this.env.Object, this.telemetryService.Object, this.tokenFetcher.Object);
```
shalinikhare27, line 38: Will we add the log assertions in the next PR.

shalinikhare27 (commented):
I would prefer `AuthOrchestrator` as the name of this class.

src/AzureAuth.Test/AuthDriverTest.cs
kyle-rader-msft: Yes, many assertions in the next PR.

reillysiemens (approved):

shalinikhare27 (approved):

src/AzureAuth/AuthOrchestrator.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace Microsoft.Authentication.AzureAuth
+{
+    using System;
+
+    using Microsoft.Authentication.MSALWrapper;
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Office.Lasso.Interfaces;
+
+    /// <summary>
+    /// A class for handling AAD Token acuisition, results logging, and telemetry collection.
```
reillysiemens, line 13: Haha. Just kidding.
```suggestion
    /// A class for handling AAD Token acquisition, results logging, and telemetry collection.

```

reillysiemens (changes requested):

reillysiemens (approved):

shalinikhare27 (approved):

PR #252, opened 2023-03-24
[3/n] 🍓 Arrive at `PublicClientAuth` as AzureAuth's abstraction over `IMsalWrapper` (rename of `TokenFetcher`)

kyle-rader-msft:
# Renames `PublicClientAuth`

After implementing this the naming came to me. At a high level, what we are doing is public client auth.
What we are using to do public client auth is our own `MsalWrapper` (renamed from `TokenFetcher`) as this is the main interface of that library.

## Ready for growth

In upcoming features we will be adding more types of auth that are different from public client. Those modes may make sense to add to IMsalWrapper, or make new interfaces with method for them within MSALWrapper. Either way, we will likely have abstractions over MSAL (`IMsalWrapper` et.al) and in AzureAuth, we will have likely have an abstraction over their usage (`IPublicClientAuth, et.al).

# Add 1 test.
I've only added one test, but it asserts quite a lot. The test itself has comments on each assertion.


src/AzureAuth.Test/PublicClientAuthTest.cs
```diff
@@ -0,0 +1,115 @@
...
+            this.telemetryService.SetupSequence(t => t.SendEvent("authflow_TestAuthFlow", It.IsAny<EventData>()));
+
+            // Act
+            TokenResult subject = this.Subject().Token(this.client, this.tenant, this.scopes, new[] { AuthMode.Web, AuthMode.DeviceCode }, this.domain, this.prompt, this.timeout);
+
+            // Assert
+            subject.Should().Be(expected);
+            // These logs are only those generated by this class/test.
+            // What we do not see here are the debug and trace logs generated by the services that are mocked for the test.
+            this.logTarget.Logs.Should().BeEmpty();
```
shalinikhare27, line 112: Is this essentially the same thing as `logMock.VerifyNoOtherCalls();`

src/AzureAuth/PublicClientAuth.cs
```diff
@@ -12,27 +13,50 @@ namespace Microsoft.Authentication.AzureAuth
...
+                domain,
+                PromptHint.Prefixed(prompt),
+                timeout);
+
+            result.Attempts.SendTelemetry(this.telemetryService);
+
+            if (result.Success != null)
+            {
+                return result.Success.TokenResult;
+            }
```
shalinikhare27, line 57: [nit]: Can we reverse this? Kind of like returning `result.Success.TokenResult` in the end. So

`if (result.Success == null)
            {
                return null;
            }

            return result.Success.TokenResult;`

src/AzureAuth.Test/PublicClientAuthTest.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            Action nullTelemetryService = () => new PublicClientAuth(this.logger, this.env.Object, null, null);
+            nullTelemetryService.Should().Throw<ArgumentNullException>().WithParameterName("telemetryService");
+
+            Action nullTokenFetcher = () => new PublicClientAuth(this.logger, this.env.Object, this.telemetryService.Object, null);
+            nullTokenFetcher.Should().Throw<ArgumentNullException>().WithParameterName("tokenFetcher");
+        }
+
+        [Test]
+        public void Good_Auth_Returns_TokenResult()
```
reillysiemens, line 74: I would like to see some acknowledgement that this is not a typical unit test, but really more like an integration test.

src/AzureAuth.Test/PublicClientAuthTest.cs
kyle-rader-msft: Partially. It's like doing that + no mocking any calls to the logger at all.

src/AzureAuth/PublicClientAuth.cs
```diff
@@ -12,27 +13,50 @@ namespace Microsoft.Authentication.AzureAuth
     /// <summary>
     /// A class for handling AAD Token acuisition, results logging, and telemetry collection.
     /// </summary>
-    public class AuthOrchestrator
+    public class PublicClientAuth : IPublicClientAuth
```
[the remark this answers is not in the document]
kyle-rader-msft: If people prefer Authentication I could be swayed, but Auth is generally accepted as a shorthand for it, and it's much harder to mis-spell :)

src/AzureAuth/PublicClientAuth.cs
kyle-rader-msft: Yep, we can do that!

reillysiemens (approved):

shalinikhare27 (approved):

PR #253, opened 2023-03-24
[4/n] 🎉 Use PublicClientAuth and Enable `azureauth ado token` command 🎊

kyle-rader-msft:
# Use PublicClientAuth

This PR focuses on using the new abstractions we've made in #252 #251 and #250 .

We remove the `Ado.TokenFetcher` and directly use `PublicClientAuth` from CommandToken.

In order to make that work, we introduce our own `ServiceCollection` in `Program.cs` for the first time, registering our own services and letting the service collection handle creation of both the `IMsalWrapper` and `IPublicClientAuth`.

This same exact pattern can now be used CommandAad, CommandToken, CommandPat, and CommandInfo. In all cases, they will log and importantly, handle telemetry, interaction via AuthMode, and prompt hints the same way.

## Reviewing this PR
There's a lot going here. As I updated the `GetToken` method on `CommandAad` I went line by line, ensuring any logging, and telemetry collection was moved to a common location.
### Logic that is now centralized
* Logging: Warning about interaction being disabled: `CommandAad` to `IEnvExtensions`
* Logging: A debug summary of auth flows attempted: `CommandToken` -> `PublicClientAuth`
* Telem: attempts and error counts: `CommandAad` -> `PublicCLientAuth`
* Telem: succeeded auth mode, silent, sid: `CommandAad` -> `PublicClientAuth`
* Telem: env var collection, previously we conditionally collected env vars if interaction was disabled, these are moved into the list for Lasso to always collect on all events.


src/AzureAuth/Ado/Constants.cs
```diff
@@ -8,6 +8,11 @@ namespace Microsoft.Authentication.AzureAuth.Ado
     /// </summary>
     internal static class Constants
     {
+        /// <summary>
+        /// ADO pipeline id.
+        /// </summary>
+        public const string SystemDefinitionId = "SYSTEM_DEFINITIONID";
```
shalinikhare27, line 14: Let's add the doc - https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#system-variables-devops-services

src/AzureAuth/Program.cs
```diff
@@ -45,12 +49,20 @@ private static void Main(string[] args)
                 backend = TelemetryOutput.ApplicationInsights;
             }
 
+            var envVarsToCollect = new[]
+            {
+                Ado.Constants.SystemDefinitionId,
+                "QBUILD_DISTRIBUTED",
```
shalinikhare27, line 55: We can move this to constants as they are `Cloud Build Agent` env var.

src/AzureAuth/Program.cs
```diff
@@ -64,7 +76,15 @@ private static void Main(string[] args)
                 sendCommandEvents: true,
                 minStderrLoglevel: stdErrLogLevel);
 
-            new Lasso(app, options).Execute(args);
+            var loggerFactory = new NLog.Extensions.Logging.NLogLoggerFactory();
+            var logger = loggerFactory.CreateLogger("AzureAuth");
+
+            IServiceCollection services = new ServiceCollection();
+            services.AddSingleton<IMsalWrapper, MsalWrapper>();
+            services.AddSingleton<IPublicClientAuth, PublicClientAuth>();
```
shalinikhare27, line 84: I like this. :)

src/AzureAuth.Test/CommandAadTest.cs
```diff
@@ -444,40 +444,6 @@ public void TestEvaluateOptionsWithoutAliasValidCommandLineOptions()
             subject.TokenFetcherOptions.Should().BeEquivalentTo(expected);
         }
 
-        [Ignore(reason: "No way to inject authflow yet")]
-        public void TestSendEvent_From_AuthFlowResult_With_Errors_And_Null_TokenResult()
```
shalinikhare27, line 448: We want to get rid of this test all together?

src/AzureAuth.Test/PublicClientAuthTest.cs
```diff
@@ -107,14 +107,38 @@ public void Good_Auth_Returns_TokenResult()
...
+                this.timeout,
+                eventData);
+
+            // Asserts
             subject.Should().Be(expected);
+
             // These logs are only those generated by this class/test.
             // What we do not see here are the debug and trace logs generated by the services that are mocked for the test.
-            this.logTarget.Logs.Should().BeEmpty();
+            this.logTarget.Logs.Should().ContainInOrder("Acquired AAD AT via TestAuthFlow in 0.00 sec");
```
shalinikhare27, line 128: I see multiple types of log statements in `PublicClientAuth` . We should perhaps have tests that asserts logging in case of `authflow == null`

src/AzureAuth/Commands/Ado/CommandToken.cs
```diff
@@ -98,31 +97,23 @@ public int OnExecute(ILogger<CommandToken> logger, IEnv env, CommandExecuteEvent
...
-            if (authflow != null)
+            if (token != null)
             {
-                logger.LogDebug($"Acquired AAD AT via {authflow.AuthFlowName} in {authflow.Duration.TotalSeconds:0.00} sec");
-                logger.LogInformation(FormatToken(authflow.TokenResult.Token, this.Output, Authorization.Bearer));
+                Console.Write(FormatToken(token.Token, this.Output, Authorization.Bearer));
                 return 0;
             }
 
             logger.LogError($"Failed to find a PAT and authenticate to ADO.");
```
shalinikhare27, line 116: Again, reversing the order to return null early is easier to read.

src/AzureAuth/PublicClientAuth.cs
```diff
@@ -37,26 +41,46 @@ public PublicClientAuth(ILogger logger, IEnv env, ITelemetryService telemetrySer
...
+
+            var authflow = result.Success;
+            if (authflow == null)
             {
+                foreach (var attempt in result.Attempts)
+                {
+                    this.logger.LogDebug($"{attempt.AuthFlowName} failed after {attempt.Duration.TotalSeconds:0.00} sec. Error count: {attempt.Errors.Count}");
+                    foreach (var e in attempt.Errors)
+                    {
+                        this.logger.LogDebug($"  {e.Message}");
```
shalinikhare27, line 71: We should add tests to asserts these logs. :)

src/AzureAuth/PublicClientAuth.cs
```diff
@@ -37,26 +41,46 @@ public PublicClientAuth(ILogger logger, IEnv env, ITelemetryService telemetrySer
...
 
+            // Report individual auth flow telemetry
             result.Attempts.SendTelemetry(this.telemetryService);
 
-            if (result.Success == null)
+            var totalErrorCount = result.Attempts.SelectMany(attempt => attempt.Errors).Count();
+            eventData.Add("error_count", totalErrorCount);
+            eventData.Add("authflow_count", result.Attempts.Count);
+
+            var authflow = result.Success;
```
shalinikhare27, line 63: This is not obvious how we set. May be add a comment why `result.Success` should be set if we had a successful authflow outcome?

src/AzureAuth/Ado/Constants.cs
kyle-rader-msft: Yeah we can do that.

src/AzureAuth/Program.cs
kyle-rader-msft: Yeah I was still debating _which_ constants to put it in. It's not ADO, and it's not AzureAuth. This is a special snowflake of an env var and I think a constant on main might be the least bad thing.

kyle-rader-msft: Or another constants file?

src/AzureAuth/Commands/Ado/CommandToken.cs
```diff
@@ -84,9 +81,11 @@ public enum OutputMode
         /// </summary>
         /// <param name="logger">The <see cref="ILogger{T}"/> instance that is used for logging.</param>
         /// <param name="env">An <see cref="IEnv"/> to use.</param>
+        /// <param name="telemetryService">An <see cref="ITelemetryService"/>.</param>
+        /// <param name="pca">An <see cref="IPublicClientAuth"/>.</param>
```
reillysiemens, line 85: Can we call `pca` a `publicClientAuth` instead?

src/AzureAuth/PublicClientAuth.cs
```diff
@@ -37,26 +41,46 @@ public PublicClientAuth(ILogger logger, IEnv env, ITelemetryService telemetrySer
...
+                    {
+                        this.logger.LogDebug($"  {e.Message}");
+                    }
+                }
+
                 return null;
             }
 
-            return result.Success.TokenResult;
+            this.logger.LogDebug($"Acquired AAD AT via {authflow.AuthFlowName} in {authflow.Duration.TotalSeconds:0.00} sec");
```
reillysiemens, line 78:
```suggestion
            this.logger.LogDebug($"Acquired AAD Access Token via {authflow.AuthFlowName} in {authflow.Duration.TotalSeconds:0.00} sec");
```

src/AzureAuth/PublicClientAuth.cs
reillysiemens: Can you elaborate on what you mean here, @shalinikhare27? I think the documentation for `TokenResult` should indicate when/why this field is set on success.

src/AzureAuth.Test/CommandAadTest.cs
kyle-rader-msft: Yes, so the way this test was written is no longer relevant, and we've increased the test coverage of all the components that CommandAad is using. We are setup to write integration tests for this now if we want to, and we can come back and add more of this as needed.

src/AzureAuth.Test/PublicClientAuthTest.cs
kyle-rader-msft: Yes! These logs specifically have no test coverage right now and are moved in this PR from CommandToken to PublicClientAuth. A follow up PR (today hopefully) will add another integration test for the auth failed, and auth has errors test cases, where these will be exercised.

src/AzureAuth/PublicClientAuth.cs
kyle-rader-msft: Same response from the first comment, a follow up PR will contain another integration test, or 2 about error cases and what get's logged.

src/AzureAuth/PublicClientAuth.cs
kyle-rader-msft: Hmm, so though docs on `MsalWrapper.Success` describe how this is the first successful auth flow result or null. It is intentional that at this level we don't know how it's set specifically, just that we either have a success or we do not.

This is a LINQ computation, so we don't want to call it multiple times. Yes, it could be cached - but that brings with it mutability concerns and is more complicated that I think is worth implementing now.

src/AzureAuth/Program.cs
shalinikhare27: Another constant file is fine too

shalinikhare27 (approved):

shalinikhare27 (approved):

reillysiemens (approved):

PR #255, opened 2023-03-27
Enforce AuthFlow Name usage and improve names for telemetry

kyle-rader-msft:
This change does 2 primary things
1. Sets friendlier names for each auth flow to use in telemetry.
2. We create a small base class for auth flows, that enforces an AuthFlow implement a name and use it in the method which implements the `IAuthFlow` interface, helping ensure all auth flows are name themselves intentionally and that those names be used in a consistent way.

### DIFF
--- a/src/MSALWrapper.Test/AuthFlow/AuthFlowResultTest.cs
+++ b/src/MSALWrapper.Test/AuthFlow/AuthFlowResultTest.cs
@@ -18,23 +18,23 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         public const string FakeToken = "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsInJoIjoieHh4IiwieDV0IjoieHh4Iiwia2lkIjoieHh4In0.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYXVkIjoiMTExMTExMTEtMTExMS0xMTExLTExMTEtMTExMTExMTExMTExIiwiaWF0IjoxNjE3NjY0Mjc2LCJuYmYiOjE2MTc2NjQyNzYsImV4cCI6MTYxNzY2ODE3NiwiYWNyIjoiMSIsImFpbyI6IllTQjBiM1JoYkd4NUlHWmhhMlVnYTJWNUlDTWtKVjQ9Iiwic2NwIjoidXNlcl9pbXBlcnNvbmF0aW9uIiwidW5pcXVlX25hbWUiOiJreXJhZGVyQG1pY3Jvc29mdC5jb20iLCJ1cG4iOiJreXJhZGVyQG1pY3Jvc29mdC5jb20iLCJ2ZXIiOiIxLjAifQ.bNc3QlL4zIClzFqH68A4hxsR7K-jabQvzB2EodgujQqc0RND_VLVkk2h3iDy8so3azN-964c2z5AiBGY6PVtWKYB-h0Z_VnzbebhDjzPLspEsANyQxaDX_ugOrf7BerQOtILWT5Vqs-A3745Bh0eTDFZpobmeENpANNhRE-yKwScjU8BDY9RimdrA2Z00V0lSliUQwnovWmtfdlbEpWObSFQAK7wCcNnUesV-jNZAUMrDkmTItPA9Z1Ks3NUbqdqMP3D6n99sy8DxQeFmbNQGYocYqI7QH24oNXODq0XB-2zpvCqy4T2jiBLgN_XEaZ5zTzEOzztpgMIWH1AUvEIyw";
 
         [Test]
-        public void ConstructorWithnullArgs()
-        {
-            var subject = new AuthFlowResult();
-            subject.Success.Should().BeFalse();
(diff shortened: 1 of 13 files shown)


src/MSALWrapper.Test/AuthFlow/AuthFlowResultTest.cs
```diff
@@ -18,23 +18,23 @@ internal class AuthFlowResultTest
...
-            var subject = new AuthFlowResult();
-            subject.Success.Should().BeFalse();
-            subject.TokenResult.Should().BeNull();
-            subject.Errors.Should().BeEmpty();
-            subject.AuthFlowName.Should().BeEmpty();
+            var tokenResult = new TokenResult(new JsonWebToken(FakeToken), Guid.NewGuid());
+            var errors = new List<Exception>();
+            Action authFlowResult = () => new AuthFlowResult(tokenResult, errors, null);
+
+            authFlowResult.Should().Throw<ArgumentNullException>();
```
kyle-rader-msft, line 27: This assertion could be made stronger using `.WithMessage`

src/MSALWrapper.Test/AuthFlow/AuthFlowResultTest.cs
```diff
@@ -18,23 +18,23 @@ internal class AuthFlowResultTest
         public const string FakeToken = "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsInJoIjoieHh4IiwieDV0IjoieHh4Iiwia2lkIjoieHh4In0.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYXVkIjoiMTExMTExMTEtMTExMS0xMTExLTExMTEtMTExMTExMTExMTExIiwiaWF0IjoxNjE3NjY0Mjc2LCJuYmYiOjE2MTc2NjQyNzYsImV4cCI6MTYxNzY2ODE3NiwiYWNyIjoiMSIsImFpbyI6IllTQjBiM1JoYkd4NUlHWmhhMlVnYTJWNUlDTWtKVjQ9Iiwic2NwIjoidXNlcl9pbXBlcnNvbmF0aW9uIiwidW5pcXVlX25hbWUiOiJreXJhZGVyQG1pY3Jvc29mdC5jb20iLCJ1cG4iOiJreXJhZGVyQG1pY3Jvc29mdC5jb20iLCJ2ZXIiOiIxLjAifQ.bNc3QlL4zIClzFqH68A4hxsR7K-jabQvzB2EodgujQqc0RND_VLVkk2h3iDy8so3azN-964c2z5AiBGY6PVtWKYB-h0Z_VnzbebhDjzPLspEsANyQxaDX_ugOrf7BerQOtILWT5Vqs-A3745Bh0eTDFZpobmeENpANNhRE-yKwScjU8BDY9RimdrA2Z00V0lSliUQwnovWmtfdlbEpWObSFQAK7wCcNnUesV-jNZAUMrDkmTItPA9Z1Ks3NUbqdqMP3D6n99sy8DxQeFmbNQGYocYqI7QH24oNXODq0XB-2zpvCqy4T2jiBLgN_XEaZ5zTzEOzztpgMIWH1AUvEIyw";
 
         [Test]
-        public void ConstructorWithnullArgs()
+        public void ConstructorWithNullAuthFlowName_ThrowsNullArgumentException()
```
kyle-rader-msft, line 21: This test name could match the following test

src/MSALWrapper.Test/AuthFlow/BrokerTest.cs
```diff
@@ -94,7 +94,7 @@ public async Task BrokerAuthFlow_GetTokenSilent_ReturnsNull()
             this.pcaWrapperMock.VerifyAll();
             authFlowResult.TokenResult.Should().Be(null);
             authFlowResult.Errors.Should().BeEmpty();
-            authFlowResult.AuthFlowName.Should().Be("Broker");
+            authFlowResult.AuthFlowName.Should().Be("broker");
```
shalinikhare27, line 97: Wondering if the friendly names will break any of my Kusto queries. For any auth flow specific query, I do the following -
```
AppEvents
| where Name startswith "azureauth_authflow"
```

This should be okay since the commands should still start with `azureauth_authflow`.

kyle-rader-msft: Yes those queries won't be impacted - but the others will. We already have a feature on the backlog to go "re-up" all the Workbooks because the queries are already broken.

shalinikhare27: Which ones would break as my service queries could break too, so I wanted to be aligned. :)

kyle-rader-msft: I don't know off hand which ones are broken, but those impacted I expect to be those using the Name explicitly and queries looking at error_messages, since we've changed how the exceptions are logged. I don't think the aggregate queries look at those, and we can update the aggregates to exclude versions before 0.6.0 which is when the errors collection change was released.

reillysiemens (dismissed):

shalinikhare27 (approved):

reillysiemens (approved):

PR #256, opened 2023-03-28
Read registry key for application insights token

goagain:
# Summary
This PR will enable AzureAuth to read registry key `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\AzureAuth` with name `ApplicationInsightsIngestionToken` for Application Insights token.

# Order
AzureAuth will try to read env var first. If it is not set, AzureAuth will try to read the reg key on Windows Platform.

# Test
Tested on my Windows 11 laptop.

No unit tests available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `azureauth ado` : Prints the help for Azure Devops commands.
   - `azureauth ado pat` : Command for creating, and locally caching Azure Devops <abbr title="Personal Access Tokens">PAT</abbr>s.
   - `azureauth ado token` : Command for passing back a <abbr title="Personal Access Tokens">PAT</abbr> from an env var, or authenticating and returning an <abbr title="Azure Active Directory">AAD</abbr> access token.
-- Now supports reading data from the registry when the environment variable is not configured.
+- When the environment variable `AZUREAUTH_APPLICATION_INSIGHTS_INGESTION_TOKEN` is not configured,
+ regkey `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\AzureAuth\ApplicationInsightsIngestionToken` will be a back up on Windows for telemetry service.
 
 ## [0.7.3] - 2023-03-14
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `azureauth ado` : Prints the help for Azure Devops commands.
   - `azureauth ado pat` : Command for creating, and locally caching Azure Devops <abbr title="Personal Access Tokens">PAT</abbr>s.
   - `azureauth ado token` : Command for passing back a <abbr title="Personal Access Tokens">PAT</abbr> from an env var, or authenticating and returning an <abbr title="Azure Active Directory">AAD</abbr> access token.
+- Now supports reading data from the registry when the environment variable is not configured.
 
 ## [0.7.3] - 2023-03-14
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `azureauth ado pat` : Command for creating, and locally caching Azure Devops <abbr title="Personal Access Tokens">PAT</abbr>s.
   - `azureauth ado token` : Command for passing back a <abbr title="Personal Access Tokens">PAT</abbr> from an env var, or authenticating and returning an <abbr title="Azure Active Directory">AAD</abbr> access token.
 - When the environment variable `AZUREAUTH_APPLICATION_INSIGHTS_INGESTION_TOKEN` is not configured,
- regkey `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\AzureAuth\ApplicationInsightsIngestionToken` will be a back up on Windows for telemetry service.
+ regkey `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\AzureAuth\ApplicationInsightsIngestionToken` will be a back up on Windows for telemetry ingestion token config.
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `azureauth ado` : Prints the help for Azure Devops commands.
   - `azureauth ado pat` : Command for creating, and locally caching Azure Devops <abbr title="Personal Access Tokens">PAT</abbr>s.
   - `azureauth ado token` : Command for passing back a <abbr title="Personal Access Tokens">PAT</abbr> from an env var, or authenticating and returning an <abbr title="Azure Active Directory">AAD</abbr> access token.
+- Now supports reading data from the registry when the environment variable is not configured.
```
kyle-rader-msft, line 13: This changelog entry needs to be more specific. It is not clear that the thing being read is the telemetry key, and that it's a back up/alternative to the `AZUREAUTH_APPLICATION_INSIGHTS_TOKEN` env var.

README.md
```diff
@@ -79,6 +79,7 @@ Instructions on using AzureAuth CLI in your applications are available [here](do
 
 # Data Collection Is Off By Default
 No telemetry will be collected unless you set the `AZUREAUTH_APPLICATION_INSIGHTS_INGESTION_TOKEN` environment variable
+or the registry key `ApplicationInsightsKey` under `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\AzureAuth`
```
kyle-rader-msft, line 82: @goagain Is it too late to change the name of the key, or can we request they change the name through another CAB request? I think it would be better if the names of the env var reg key matched. Because the env var is called an Application Insights Ingestion Token, I'd expect/hope the reg key to be named the same, `ApplicationInsightsIngestionToken`. 

And I see that's actually what we called the variable name in the code ;)
* `ApplicationInsightsIngestionTokenRegKeyPath`
* `ApplicationInsightsIngestionTokenRegKeyName`

src/AzureAuth/Program.cs
```diff
@@ -35,14 +36,26 @@ private static void Main(string[] args)
...
             string applicationInsightsIngestionToken = Environment.GetEnvironmentVariable(EnvVars.ApplicationInsightsIngestionTokenEnvVar);
+
+            if (string.IsNullOrEmpty(applicationInsightsIngestionToken))
+            {
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    applicationInsightsIngestionToken = Win32.Registry.GetValue(
+                        EnvVars.ApplicationInsightsIngestionTokenRegKeyPath,
+                        EnvVars.ApplicationInsightsIngestionTokenRegKeyName,
+                        null) as string;
```
kyle-rader-msft, line 55: What is the behavior of this when the value is unset?

src/AzureAuth/Program.cs
```diff
@@ -35,14 +36,26 @@ private static void Main(string[] args)
...
             // `CommandLineApplication` to dynamically set telemetry configuration. Even if
             // it did, we can't guarantee that a failure in command line parsing wouldn't
             // trigger telemetry before we ever get to disable it.
             //
             // To disable telemetry a user need only leave this environment variable unset. It's off by default.
             string applicationInsightsIngestionToken = Environment.GetEnvironmentVariable(EnvVars.ApplicationInsightsIngestionTokenEnvVar);
+
+            if (string.IsNullOrEmpty(applicationInsightsIngestionToken))
+            {
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
kyle-rader-msft, line 50: Because this is Windows only, I think we should use `#ifdef WINDOWS` instead of a runtime check. We don't need to both checking on other platforms, and in the future, we may have a different block to conditionally compile in the else case for Mac, and for Linux. I don't feel strongly about this, but I think it would be good to know ho many `ms` it actually takes to check the OS at runtime. I've seen that code memoized in many other codebases to avoid doing it more than once, which makes me worry it's longer than one might think.

src/AzureAuth/Program.cs
kyle-rader-msft: Is it worth having a try-catch here? Can this throw? I ask because this is outside the normal Lasso error catching boundary, and we shouldn't crash if we fail to get this. We can log errors to stderr, but we shouldn't crash.

kyle-rader-msft: With the potential advent of error handling around this, I think it would also be worth pulling this into it's own static method to keep main as small and easy to read as possible.

kyle-rader-msft (changes requested):
This looks like the right direction to me but I have a few small questions.

README.md
goagain: I will ask in CAB workitem

goagain: Updated

src/AzureAuth/Program.cs
goagain: If we use `if PlatformWindows`, we have to suppress CA1416.

#if PlatformWindows
#pragma warning disable CA1416 // Validate platform compatibility
                applicationInsightsIngestionToken = Win32.Registry.GetValue(
                    EnvVars.ApplicationInsightsIngestionTokenRegKeyPath,
                    EnvVars.ApplicationInsightsIngestionTokenRegKeyName,
                    null) as string;
#pragma warning restore CA1416 // Validate platform compatibility
#endif

src/AzureAuth/Program.cs
goagain: We don't need to add try-catch. If the key doesn't exist or is empty, `Registry.GetValue` will return null or empty string.

src/AzureAuth/Program.cs
kyle-rader-msft: Does this more properly may require more changes to our targeting configuration, so I'm fine staying with the runtime for now.

kyle-rader-msft (approved):
Approved, with 1 small suggestion for the CHANGELOG.

kyle-rader-msft:
@goagain this PR also requires an update.

kyle-rader-msft (approved):

shalinikhare27 (approved):

PR #259, opened 2023-03-31
Define PAT storage constants

reillysiemens:
This PR just adds constants that will be used for configuring encrypted PAT storage. You'll notice overlap between some of these names and those in `src\MSALWrapper\PCACache.cs`. They're both concerned with configuring a storage class from the MSAL Extensions library, but for different classes and different caches.


mvanchaa (approved):
The comments are helpful - clear and concise :)

Haard30 (approved):

PR #260, opened 2023-03-31
Implement azureauth ado pat subcommand

reillysiemens:
This PR adds a functional version of the `azureauth ado pat` subcommand with a few caveats.

- The logging is insufficient and will be addressed in a subsequent PR.
- The telemetry is insufficient and will be addressed in a subsequent PR.
- ~I haven't addressed a cross-platform cache location, but I _think_ I want to put it alongside the AzureAuth installation itself.~
  - [x] Unix: `~/.azureauth/azureauth-pat.cache`
  - [x] Windows: `C:\Users\<user>\AppData\Local\Programs\AzureAuth\azureauth-pat.cache`
- We probably want to do a persistence check before trying to use the cache, but I'm still reading up on how to do that.
- There are basically no tests for this code, but with few exceptions there aren't any conditional paths and most places in `CommandPat.cs` are just configuring and then calling into well-tested components.

The basic workflow here is
1. Get an Azure AD access token.
2. Create a [`PatCache`](https://github.com/AzureAD/microsoft-authentication-cli/blob/9c029269ce30a6d72d87d4be55a77d52b80e3f56/src/AdoPat/PatCache.cs), [`PatClient`](https://github.com/AzureAD/microsoft-authentication-cli/blob/9c029269ce30a6d72d87d4be55a77d52b80e3f56/src/AdoPat/PatClient.cs), and use them to create a [`PatManager`](https://github.com/AzureAD/microsoft-authentication-cli/blob/9c029269ce30a6d72d87d4be55a77d52b80e3f56/src/AdoPat/PatManager.cs).
3. Acquire a lock ensuring no other applications can use the cache.
4. Use the `PatManager` to get a PAT.
5. Display said PAT.

This is a draft PR while I search for answers to the few remaining `TODO`s, but I want to get early feedback while I'm working on that. I will update the draft PR to a full PR when the `TODO`s are resolved.

### DIFF
--- a/src/AzureAuth/Commands/Ado/CommandPat.cs
+++ b/src/AzureAuth/Commands/Ado/CommandPat.cs
@@ -155,7 +155,7 @@ You can use any combination with multiple instances of the --mode flag.
                 var pat = manager.GetPatAsync(this.PatOptions()).Result;
 
                 // Do not use logger to avoid printing PATs into log files.
-                Console.WriteLine(FormatPat(pat, this.Output));
+                Console.Write(FormatPat(pat, this.Output));
             }
 
             return 0;
@@ -164,12 +164,12 @@ You can use any combination with multiple instances of the --mode flag.
         private static string FormatPat(PatToken pat, OutputMode output) => output switch
         {
             OutputMode.None => string.Empty,
-            OutputMode.Status => $"\"{pat.DisplayName}\" valid until {pat.ValidTo:O}",
-            OutputMode.Token => pat.Token,
-            OutputMode.Base64 => pat.Token.Base64(),
-            OutputMode.Header => pat.Token.AsHeader(AzureAuth.Ado.Authorization.Basic),
-            OutputMode.HeaderValue => pat.Token.AsHeaderValue(AzureAuth.Ado.Authorization.Basic),
-            OutputMode.Json => pat.AsJson(),
+            OutputMode.Status => $"\"{pat.DisplayName}\" valid until {pat.ValidTo:O}\n",
+            OutputMode.Token => $"{pat.Token}\n",
+            OutputMode.Base64 => $"{pat.Token.Base64()}\n",
+            OutputMode.Header => $"{pat.Token.AsHeader(AzureAuth.Ado.Authorization.Basic)}\n",
+            OutputMode.HeaderValue => $"{pat.Token.AsHeaderValue(AzureAuth.Ado.Authorization.Basic)}\n",
+            OutputMode.Json => $"{pat.AsJson()}\n",
             _ => throw new ArgumentOutOfRangeException(nameof(output)),
         };
 
--- a/src/AzureAuth/Commands/Ado/CommandPat.cs
+++ b/src/AzureAuth/Commands/Ado/CommandPat.cs
@@ -117,20 +117,13 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Ado
                 return 1;
             }
 
-            var patOptions = new PatOptions
-            {
-                Organization = this.Organization,
-                DisplayName = this.DisplayName,
-                Scopes = this.Scopes,
-            };
-
             var cache = this.Cache();
             var client = this.Client(accessToken.Token);
             var manager = new PatManager(cache, client);
 
             using (new CrossPlatLock(LockfilePath))
             {
-                var pat = manager.GetPatAsync(patOptions).Result;
+                var pat = manager.GetPatAsync(this.PatOptions()).Result;
 
                 // Do not use logger to avoid printing PATs into log files.
                 Console.WriteLine(FormatPat(pat, this.Output));
@@ -160,6 +153,16 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Ado
                 AccessTokenPrompt,
                 AccessTokenTimeout,
                 eventData);
+        }
+
+        private PatOptions PatOptions()
+        {
+            return new PatOptions
+            {
+                Organization = this.Organization,
+                DisplayName = this.DisplayName,
+                Scopes = this.Scopes,
+            };
         }
 
         private IPatClient Client(string accessToken)
--- a/src/AzureAuth/Commands/Ado/CommandPat.cs
+++ b/src/AzureAuth/Commands/Ado/CommandPat.cs
@@ -85,7 +85,7 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Ado
         [Option(CommandAad.ModeOption, CommandAad.AuthModeHelperText, CommandOptionType.MultipleValue)]
         private IEnumerable<AuthMode> AuthModes { get; set; } = new[] { AuthMode.Default };
 
-        [Option(CommandAad.DomainOption, $"{CommandAad.DomainHelpText}\n[default: microsoft.com]", CommandOptionType.SingleValue)]
+        [Option(CommandAad.DomainOption, $"{CommandAad.DomainHelpText}\n[default: {AzureAuth.Ado.Constants.PreferredDomain}]", CommandOptionType.SingleValue)]
         private string Domain { get; set; } = AzureAuth.Ado.Constants.PreferredDomain;
 
         /// <summary>
--- a/src/AzureAuth/Commands/Ado/CommandPat.cs
+++ b/src/AzureAuth/Commands/Ado/CommandPat.cs
@@ -159,11 +159,11 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Ado
         {
             OutputMode.None => string.Empty,
             OutputMode.Status => $"\"{pat.DisplayName}\" valid until {pat.ValidTo:O}\n",
-            OutputMode.Token => $"{pat.Token}\n",
-            OutputMode.Base64 => $"{pat.Token.Base64()}\n",
-            OutputMode.Header => $"{pat.Token.AsHeader(AzureAuth.Ado.Authorization.Basic)}\n",
-            OutputMode.HeaderValue => $"{pat.Token.AsHeaderValue(AzureAuth.Ado.Authorization.Basic)}\n",
-            OutputMode.Json => $"{pat.AsJson()}\n",
+            OutputMode.Token => pat.Token,
+            OutputMode.Base64 => pat.Token.Base64(),
+            OutputMode.Header => pat.Token.AsHeader(AzureAuth.Ado.Authorization.Basic),
+            OutputMode.HeaderValue => pat.Token.AsHeaderValue(AzureAuth.Ado.Authorization.Basic),
+            OutputMode.Json => pat.AsJson(),
             _ => throw new ArgumentOutOfRangeException(nameof(output)),
         };
 
--- a/src/AzureAuth/Commands/Ado/CommandPat.cs
+++ b/src/AzureAuth/Commands/Ado/CommandPat.cs
@@ -3,10 +3,22 @@
 
 namespace Microsoft.Authentication.AzureAuth.Commands.Ado
 {
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
     using McMaster.Extensions.CommandLineUtils;
 
+    using Microsoft.Authentication.AdoPat;
+    using Microsoft.Authentication.AzureAuth.Ado;
+    using Microsoft.Authentication.MSALWrapper;
     using Microsoft.Extensions.Logging;
+    using Microsoft.Identity.Client.Extensions.Msal;
     using Microsoft.Office.Lasso.Telemetry;
+    using Microsoft.VisualStudio.Services.DelegatedAuthorization;
+    using Microsoft.VisualStudio.Services.OAuth;
+    using Microsoft.VisualStudio.Services.WebApi;
+
+    using PatStorageParameters = Microsoft.Authentication.AzureAuth.Ado.Constants.PatStorageParameters;
 
     /// <summary>
     /// An ADO Command for creating or fetching, and returning Azure Devops PATs.
@@ -25,6 +37,21 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Ado
 
         private const string OutputOption = "--output";
         private const string OutputHelp = "How PAT information is displayed. [default: token]\n[possible values: none, status, token, base64, header, headervalue, json]";
+
+        private const string DomainOption = "--domain";
+        private const string DomainHelp = "The preferred domain used when acquiring Azure Active Directory access tokens. [default: microsoft.com]";
+
+        private const string AccessTokenPrompt = "AzureAuth ADO PAT";
+        private static readonly IEnumerable<AuthMode> AccessTokenAuthModes = new[] { AuthMode.Default };
+        private static readonly TimeSpan AccessTokenTimeout = TimeSpan.FromMinutes(15);
+
+        private static readonly string LockfilePath = Path.Combine(Path.GetTempPath(), AzureAuth.Ado.Constants.PatLockfileName);
+
+        // TODO: This is currently Windows-specific. We should probably define this directory conditionally by platform.
+        private static readonly string CacheDirectory = Path.Combine(
+            Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData),
+            "Programs",
+            "AzureAuth");
 
         // The possible PAT output modes.
         private enum OutputMode
@@ -63,16 +90,101 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Ado
         [Option(OutputOption, OutputHelp, CommandOptionType.SingleValue)]
         private OutputMode Output { get; set; } = OutputMode.Token;
 
+        [Option(DomainOption, DomainHelp, CommandOptionType.SingleValue)]
+        private string Domain { get; set; } = "microsoft.com";
+
         /// <summary>
         /// Executes the command and returns a status code indicating the success or failure of the execution.
         /// </summary>
         /// <param name="logger">The <see cref="ILogger{T}"/> instance that is used for logging.</param>
+        /// <param name="publicClientAuth">An <see cref="IPublicClientAuth"/>.</param>
         /// <param name="eventData">Lasso injected command event data.</param>
         /// <returns>An integer status code. 0 for success and non-zero for failure.</returns>
-        public int OnExecute(ILogger<CommandPat> logger, CommandExecuteEventData eventData)
+        public int OnExecute(ILogger<CommandPat> logger, IPublicClientAuth publicClientAuth, CommandExecuteEventData eventData)
         {
-            logger.LogInformation("coming soon");
+            var accessToken = this.AccessToken(publicClientAuth, eventData);
+            if (accessToken == null)
+            {
+                logger.LogError("Failed to acquire an Azure DevOps access token. Re-run with '--verbosity debug' for more info.");
+                return 1;
+            }
+
+            var patOptions = new PatOptions
+            {
+                Organization = this.Organization,
+                DisplayName = this.DisplayName,
+                Scopes = this.Scopes,
+            };
+
+            var cache = this.Cache();
+            var client = this.Client(accessToken.Token);
+            var manager = new PatManager(cache, client);
+
+            using (new CrossPlatLock(LockfilePath))
+            {
+                var pat = manager.GetPatAsync(patOptions).Result;
+
+                // Do not use logger to avoid printing PATs into log files.
+                Console.WriteLine(FormatPat(pat, this.Output));
+            }
+
             return 0;
+        }
+
+        private static string FormatPat(PatToken pat, OutputMode output) => output switch
+        {
+            OutputMode.None => string.Empty,
+            OutputMode.Status => $"\"{pat.DisplayName}\" valid until {pat.ValidTo:O}",
+            OutputMode.Token => pat.Token,
+            OutputMode.Base64 => pat.Token.Base64(),
+            OutputMode.Header => pat.Token.AsHeader(AzureAuth.Ado.Authorization.Basic),
+            OutputMode.HeaderValue => pat.Token.AsHeaderValue(AzureAuth.Ado.Authorization.Basic),
+            OutputMode.Json => pat.AsJson(),
+            _ => throw new ArgumentOutOfRangeException(nameof(output)),
+        };
+
+        private TokenResult AccessToken(IPublicClientAuth publicClientAuth, CommandExecuteEventData eventData)
+        {
+            return publicClientAuth.Token(
+                AzureAuth.Ado.Constants.AdoParams,
+                AccessTokenAuthModes,
+                this.Domain,
+                AccessTokenPrompt,
+                AccessTokenTimeout,
+                eventData);
+        }
+
+        private IPatClient Client(string accessToken)
+        {
+            var baseUrl = new Uri($"{AzureAuth.Ado.Constants.BaseUrl}/{this.Organization}");
+            var credentials = new VssOAuthAccessTokenCredential(accessToken);
+            var connection = new VssConnection(baseUrl, credentials);
+            var tokensHttpClientWrapper = new TokensHttpClientWrapper(connection);
+            return new PatClient(tokensHttpClientWrapper);
+        }
+
+        private IPatCache Cache()
+        {
+            var storageProperties = new StorageCreationPropertiesBuilder(
+                PatStorageParameters.CacheFileName,
+                CacheDirectory)
+            .WithMacKeyChain(
+                PatStorageParameters.MacOSServiceName,
+                PatStorageParameters.MacOSAccountName)
+            .WithLinuxKeyring(
+                PatStorageParameters.LinuxKeyRingSchemaName,
+                PatStorageParameters.LinuxKeyRingCollection,
+                PatStorageParameters.LinuxKeyRingLabel,
+                PatStorageParameters.LinuxKeyRingAttr1,
+                PatStorageParameters.LinuxKeyRingAttr2)
+            .Build();
+
+            // TODO: We probably need to do a `storage.VerifyPersistence` check
+            // before using this. On Linux this won't work in a headless
+            // environment, so we'll need to find a fallback or fail early.
+            var storage = Storage.Create(storageProperties);
+            var storageWrapper = new StorageWrapper(storage);
+            return new PatCache(storageWrapper);
         }
     }
 }
--- a/src/AzureAuth/Commands/Ado/CommandPat.cs
+++ b/src/AzureAuth/Commands/Ado/CommandPat.cs
@@ -99,7 +99,7 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Ado
         private OutputMode Output { get; set; } = OutputMode.Token;
 
         [Option(DomainOption, DomainHelp, CommandOptionType.SingleValue)]
-        private string Domain { get; set; } = "microsoft.com";
+        private string Domain { get; set; } = AzureAuth.Ado.Constants.PreferredDomain;
 
         /// <summary>
         /// Executes the command and returns a status code indicating the success or failure of the execution.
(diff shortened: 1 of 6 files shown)


src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -63,16 +90,101 @@ private enum OutputMode
...
+                AccessTokenAuthModes,
+                this.Domain,
+                AccessTokenPrompt,
+                AccessTokenTimeout,
+                eventData);
+        }
+
+        private IPatClient Client(string accessToken)
+        {
+            var baseUrl = new Uri($"{AzureAuth.Ado.Constants.BaseUrl}/{this.Organization}");
```
kyle-rader-msft, line 159: Using a `string.Join` versus interpolation here may be a tad more readable/extendable.

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -63,16 +90,101 @@ private enum OutputMode
...
+            var credentials = new VssOAuthAccessTokenCredential(accessToken);
```
kyle-rader-msft, line 160: That's neat, I didn't know about `VssOAuthAccessTokenCredential` - we might want to recommend that usage in AzureAuth-Lib docs.

kyle-rader-msft (commented):
@reillysiemens Overall this is looking great.

Question on x-plat file location. I'm not sure that's a question we need to answer? My understanding of the MSAL x-plat storage abstraction was that it only uses an actual file on Windows. The file is never used on Mac or Linux, because on those platforms the other properties are used to create keychain entries.

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -26,6 +38,21 @@ public class CommandPat
         private const string OutputOption = "--output";
         private const string OutputHelp = "How PAT information is displayed. [default: token]\n[possible values: none, status, token, base64, header, headervalue, json]";
 
+        private const string DomainOption = "--domain";
+        private const string DomainHelp = "The preferred domain used when acquiring Azure Active Directory access tokens. [default: microsoft.com]";
+
+        private const string AccessTokenPrompt = "AzureAuth ADO PAT";
```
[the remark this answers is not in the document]
reillysiemens: That's a great point.
1. I forgot we already add an `AzureAuth` prefix, so I'll drop that.
2. In addition to creation we also do caching and regeneration, so it's not _just_ creation...

How about `Azure DevOps PAT CLI`?

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -63,16 +90,101 @@ private enum OutputMode
         [Option(OutputOption, OutputHelp, CommandOptionType.SingleValue)]
         private OutputMode Output { get; set; } = OutputMode.Token;
 
+        [Option(DomainOption, DomainHelp, CommandOptionType.SingleValue)]
+        private string Domain { get; set; } = "microsoft.com";
```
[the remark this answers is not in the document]
reillysiemens: I'm happy to see this pulled into `Microsoft.Authentication.AzureAuth.Ado.Constants`. I'll put it in there and you `CommandToken` can use it as desired.

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -26,6 +38,21 @@ public class CommandPat
         private const string OutputOption = "--output";
         private const string OutputHelp = "How PAT information is displayed. [default: token]\n[possible values: none, status, token, base64, header, headervalue, json]";
 
+        private const string DomainOption = "--domain";
+        private const string DomainHelp = "The preferred domain used when acquiring Azure Active Directory access tokens. [default: microsoft.com]";
+
+        private const string AccessTokenPrompt = "AzureAuth ADO PAT";
+        private static readonly IEnumerable<AuthMode> AccessTokenAuthModes = new[] { AuthMode.Default };
+        private static readonly TimeSpan AccessTokenTimeout = TimeSpan.FromMinutes(15);
```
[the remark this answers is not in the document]
reillysiemens: I think when we originally discussed this feature we talked about keeping those kinds of flags to a minimum and just using the defaults (which should provide the best experience). Timeout certainly makes sense to me, but should we give a lot of control over the access token mode if we already have sane defaults?

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -63,16 +90,101 @@ private enum OutputMode
...
+                logger.LogError("Failed to acquire an Azure DevOps access token. Re-run with '--verbosity debug' for more info.");
+                return 1;
+            }
+
+            var patOptions = new PatOptions
+            {
+                Organization = this.Organization,
+                DisplayName = this.DisplayName,
+                Scopes = this.Scopes,
+            };
```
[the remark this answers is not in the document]
reillysiemens: My thinking here was that `new PatOptions { ... }` practically _was_ that "small function", but I suppose it would be just as clean if I wrote
```csharp
var pat = manager.GetPatAsync(this.patOptions()).Result;
```
, so I'm happy to change that.

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -63,16 +90,101 @@ private enum OutputMode
...
+            var cache = this.Cache();
+            var client = this.Client(accessToken.Token);
+            var manager = new PatManager(cache, client);
+
+            using (new CrossPlatLock(LockfilePath))
+            {
+                var pat = manager.GetPatAsync(patOptions).Result;
+
+                // Do not use logger to avoid printing PATs into log files.
+                Console.WriteLine(FormatPat(pat, this.Output));
```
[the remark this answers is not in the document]
reillysiemens: I wondered if you'd comment on that. 😜 Glad you did since it was kind of bugging me.

Here's why I didn't do what you suggest. `FormatPat` returns a `string` after formatting and keeps all I/O as the domain of the calling method. From where I sand, to not write the newline requires conditional logic either in the calling method (to interpret `null` or `string.Empty` as a reason to avoid calling `Console.WriteLine`) or moving the printing into `FormatPat` itself, making it impure.

I'm actually happy to do either and don't have any strong preferences. Is there an approach you prefer, @kyle-rader-msft?

src/AzureAuth/Commands/Ado/CommandPat.cs
reillysiemens: Totally reasonable. I'll change this.

src/AzureAuth/Commands/Ado/CommandPat.cs
reillysiemens: I didn't know about it either until I went spelunking in the Visual Studio code. I don't think it's especially well documented. Happy to mention it wherever it seems appropriate.

reillysiemens:
@kyle-rader-msft:
> My understanding of the MSAL x-plat storage abstraction was that it only uses an actual file on Windows. The file is never used on Mac or Linux, because on those platforms the other properties are used to create keychain entries.

This is not the case. Unless, of course, I'm using the library wrong. My understanding is that the cross-platform storage abstraction is only concerned with setting up the encryption (the keys for which are stored in the keychain/keyring, but the cache itself is on disk regardless of OS.

src/AzureAuth/Commands/Ado/CommandPat.cs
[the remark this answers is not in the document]
kyle-rader-msft: I like that.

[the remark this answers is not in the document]
kyle-rader-msft: Thinking more: I think `CLI` could be dropped.

src/AzureAuth/Commands/Ado/CommandPat.cs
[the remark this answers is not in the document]
kyle-rader-msft: I think the main consideration for auth modes is headless. The default modes do not include device code since 99% of the time it won't work with how things use AzureAuth. So imagine a mac or linux ssh session wanting to get and cache a PAT. They need to be able to say `azureauth ado pat --mode devicecode`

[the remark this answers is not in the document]
kyle-rader-msft: I suppose for that scenario, one _could_ run `azureauth ado token --mode devicecode` first, to cache AAD auth, and then could call the pat command as a workaround.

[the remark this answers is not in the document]
reillysiemens: I hesitated to add something like this in the first place because using an access token to get a PAT is an implementation detail and I didn't want to give too much control over that at the top level.

Headless usage is a good consideration though. I don't see requiring certain users to make additional calls just to warm a cache as a good option. What would you call this flag? Just `--mode` for consistency? `--access-token-mode` for specificity? 🤔

src/AzureAuth/Commands/Ado/CommandPat.cs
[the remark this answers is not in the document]
reillysiemens: It could, but that makes it _awfully_ difficult to know what this is really about.

> `AzureAuth: Azure DevOps PAT` is requesting your credentials.
> Say what now?

vs.

> `AzureAuth: Azure DevOps PAT CLI` is requesting your credentials.
> Oh, OK, Some CLI for working with PATs wants me to auth. Sure.

src/AzureAuth/Commands/Ado/CommandPat.cs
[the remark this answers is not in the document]
reillysiemens: Personally, I think I lean towards giving up on functional purity, renaming `FormatPat` to `DisplayPat`, and just adding `Console.WriteLine`s to the `switch` expression within.

src/AzureAuth/Commands/Ado/CommandPat.cs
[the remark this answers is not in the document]
reillysiemens: We could also just add a required `--prompt` flag and let the user be responsible for telling us what's using it.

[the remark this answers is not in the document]
reillysiemens: `AzureAuth: Azure DevOps PAT CLI` can still be a default value.

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -26,6 +38,29 @@ public class CommandPat
         private const string OutputOption = "--output";
         private const string OutputHelp = "How PAT information is displayed. [default: token]\n[possible values: none, status, token, base64, header, headervalue, json]";
 
+        private const string DomainOption = "--domain";
+        private const string DomainHelp = "The preferred domain used when acquiring Azure Active Directory access tokens. [default: microsoft.com]";
```
shalinikhare27, line 42: Interesting! I did not know we could have a PAT generated for a non Microsoft domain.

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -26,6 +38,29 @@ public class CommandPat
...
+
+        private const string AccessTokenPrompt = "AzureAuth ADO PAT";
+        private static readonly IEnumerable<AuthMode> AccessTokenAuthModes = new[] { AuthMode.Default };
+        private static readonly TimeSpan AccessTokenTimeout = TimeSpan.FromMinutes(15);
+
+        private static readonly string LockfilePath = Path.Combine(Path.GetTempPath(), AzureAuth.Ado.Constants.PatLockfileName);
+
+        // On all platforms the PAT cache should be in the same directory as a typical AzureAuth installation.
+        //   - On Windows this is usually `%LOCALAPPDATA%\Programs\AzureAuth`.
```
shalinikhare27, line 51: You mentioned `usually`. What's our fallback if it's not found in the path you are expecting it to?

reillysiemens: I just mentioned "usually" because there are installations of AzureAuth that don't go to the normal directory. I can remove that word from this comment.

There isn't a fallback. This is where we expect it to be and if it's not there we can't guess at where else it should be.

src/AzureAuth/Commands/Ado/CommandPat.cs
reillysiemens: Presumably it could happen if you were using `azureauth ado pat` as part of a non-Microsoft organization.

src/AzureAuth/Commands/Ado/CommandPat.cs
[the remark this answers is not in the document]
mvanchaa: I like the option `--mode`

src/AzureAuth/Commands/Ado/CommandPat.cs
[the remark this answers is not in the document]
kyle-rader-msft: I'm leaning towards a required `--prompt` option for 2 reasons:
1. This is how we denote usage from different tools in telemetry.
2. people generally may not even know AzureAuth is being used or what it is. What they will know is the tool they ran is creating a PAT, so the consuming tool should define the prompt. AzureAuth-Lib already has the prompt as a required argument to the token method.

src/AzureAuth/Commands/Ado/CommandPat.cs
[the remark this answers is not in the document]
kyle-rader-msft: From our offline discussion, I am generally pro "don't add it until you need it" but I think keeping the controls for auth consistent across the commands that need to auth. Also because of the overhead of shipping new versions internally atm, I'd rather have it and not need it, then need it and not have it.

src/AzureAuth/Commands/Ado/CommandPat.cs
kyle-rader-msft: Technically we could see where the current executable is, but I think a better description is one Tucker told me that - the MSAL cache we use is also system wide. All versions of AzureAuth attempt to use the same cache to have the best experience, and that same concept applies here. Two different versions of AzureAUth across several different repos or a system install can all share the same PAT cache minimizing PAT creation notifications and potential auth prompts.

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -26,6 +38,29 @@ public class CommandPat
...
+        //   - On Unix-like platforms this is usually `~/.azureauth`.
+#if PlatformWindows
```
kyle-rader-msft, line 53: I'm curious why this is not generating the same warning that Rick sees in program.cs (or maybe it is and we're just not seeing it here) - about this being reachable on all platforms.

src/AzureAuth/Commands/Ado/CommandPat.cs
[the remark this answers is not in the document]
kyle-rader-msft: Yep that's my recommendation. Use Write, and let the format add newlines as needed.

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -63,16 +98,104 @@ private enum OutputMode
...
+                PatStorageParameters.MacOSAccountName)
+            .WithLinuxKeyring(
+                PatStorageParameters.LinuxKeyRingSchemaName,
+                PatStorageParameters.LinuxKeyRingCollection,
+                PatStorageParameters.LinuxKeyRingLabel,
+                PatStorageParameters.LinuxKeyRingAttr1,
+                PatStorageParameters.LinuxKeyRingAttr2)
+            .Build();
+
+            // TODO: We probably need to do a `storage.VerifyPersistence` check
```
kyle-rader-msft, line 193: In my experience, the cache persistency check has only failed on mac, when there is some unexplainable issue with the KeyChain. And in that case, I'm still torn on whether we should fail all up, or proceed with effectively no cache, which could result in an auth prompt and new PAT on every run.

mvanchaa (approved):

goagain:
Oh, I clicked the checkbox in the description, and it was edited by me.

src/AzureAuth/Commands/Ado/CommandPat.cs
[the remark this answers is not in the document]
reillysiemens: I've added a `--prompt-hint` flag (so named because it's consistent with `azureauth aad`) and I've also added a private method to validate required falgs.

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -63,16 +98,104 @@ private enum OutputMode
...
+            var connection = new VssConnection(baseUrl, credentials);
+            var tokensHttpClientWrapper = new TokensHttpClientWrapper(connection);
+            return new PatClient(tokensHttpClientWrapper);
+        }
+
+        private IPatCache Cache()
+        {
+            var storageProperties = new StorageCreationPropertiesBuilder(
+                PatStorageParameters.CacheFileName,
+                CacheDirectory)
```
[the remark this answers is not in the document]
reillysiemens: It's certainly a related value, but it's not _quite_ constant, IMO. It's both operating system and user-dependent. For example, given a user `foobar` they might have the following cache directories
- Windows: `C:\Users\foobar\AppData\Local\Programs\AzureAuth`
- MacOS: `/Users/foobar/.azureauth`
- Linux: `/home/foobar/.azureauth`

After being defined once at runtime they never change again for the lifetime of the application, but they're not as rigidly defined as other "true constants" in my mind.

@kyle-rader-msft discussed adding this value to its own static class and memoizing the value. What do you think about that, @mvanchaa?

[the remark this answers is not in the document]
shalinikhare27: I agree with Tucker here. Since we use some logic to create them and then they are constants from there on, they are not truly constants. But great point, @mvanchaa :) It's almost who you talk to and a preference thing.

src/AzureAuth/Commands/Ado/CommandPat.cs
shalinikhare27: Umm, ADO is not Microsoft only and we sell to customers? Good to know! :)

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -60,19 +97,149 @@ private enum OutputMode
...
+            {
+                return 1;
+            }
+
+            var accessToken = this.AccessToken(publicClientAuth, eventData);
+            if (accessToken == null)
+            {
+                logger.LogError("Failed to acquire an Azure DevOps access token. Re-run with '--verbosity debug' for more info.");
+                return 1;
+            }
```
shalinikhare27, line 128: We return after this step, so all we ask is to re-run `verbosity-debug`. So if I was the user, I would see that and email our team alias for help? Like what's the next thing to do from a user perspective?

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -60,19 +97,149 @@ private enum OutputMode
...
+
+            var cache = this.Cache();
+            var client = this.Client(accessToken.Token);
+            var manager = new PatManager(cache, client);
+
+            using (new CrossPlatLock(LockfilePath))
+            {
+                var pat = manager.GetPatAsync(this.PatOptions()).Result;
```
shalinikhare27, line 136: Reason we don't write as `await manager.GetPatAsync(this.PatOptions);`

await asynchronously unwraps the result of your task, whereas just using Result would block until the task had completed.

shalinikhare27 (commented):
`CommandPat` is a huge class with many private methods. That's the best way to break it?

src/AzureAuth/Commands/Ado/CommandPat.cs
reillysiemens: Yeah, users would need to reach out for help and provide this information. The program can't proceed at this point, so there's nothing more to be done in the code.

src/AzureAuth/Commands/Ado/CommandPat.cs
reillysiemens: If we addressed this I think it would be good to do as a follow up PR. We could make `PatManager.GetPatAsync` actually just be synchronous and then there wouldn't be a need to use `.Result` at this point.

reillysiemens: For now, this is working, so I'm inclined to leave it as is.

src/AzureAuth/Commands/Ado/CommandPat.cs
reillysiemens: No idea. Maybe we have a StyleCop configuration issue? 🤔

src/AzureAuth/Commands/Ado/CommandPat.cs
reillysiemens: If this were being called in parallel with the expectation that the user would get a single PAT from caching the first call I think we'd end up with a bad user experience if we ignored caching failures. It could lead to a "PAT explosion" where you wanted 1 PAT, but now you have 100 because none of them got cached between calls.

mvanchaa (approved):

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -23,9 +35,48 @@ public class CommandPat
         private const string ScopeOption = "--scope";
         private const string ScopeHelp = "A token scope for accessing Azure DevOps resources. Repeated invocations allowed.";
 
+        private const string PromptHintOption = "--prompt-hint";
+        private const string PromptHintHelp = "A prompt hint to contextualize prompts and identify uses in telemetry, when captured.";
```
kyle-rader-msft, line 39: These are already defined as `public const` in CommandAad for re-use by other commands needing to do auth. Same for AuthMode, and Domain. Happy to see the content improvements made in those shared consts :)

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -23,9 +35,48 @@ public class CommandPat
...
         private const string OutputOption = "--output";
         private const string OutputHelp = "How PAT information is displayed. [default: token]\n[possible values: none, status, token, base64, header, headervalue, json]";
 
+        private const string ModeOption = "--mode";
+#if PlatformWindows
+        /// <summary>
+        /// The help text for the <see cref="ModeOption"/> option.
+        /// </summary>
+        private const string ModeHelp = @"Authentication mode. Repeated invocations allowed. [default: iwa (Integrated Windows Auth), then broker, then web]
+[possible values: all, iwa, broker, web, devicecode]";
```
kyle-rader-msft, line 50:
```suggestion
        private const string ModeHelp = @"Authentication mode. Repeated invocations allowed. 
[default: iwa (Integrated Windows Auth), then broker, then web]
[possible values: all, iwa, broker, web, devicecode]";
```

kyle-rader-msft: Newlines are preserved in the help text rendering, and all of it is auto-wrapper on the CLI based on terminal width. So I generally try to use newlines liberally to make reading the help easier.

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -60,19 +111,161 @@ private enum OutputMode
...
+            IPatCache cache = null;
+            try
+            {
+                cache = this.Cache();
+            }
+            catch (MsalCachePersistenceException e)
+            {
+                logger.LogError($"Failed to validate cache persistence: {e.Message}");
+                return 1;
+            }
```
kyle-rader-msft, line 156: I'd prefer to see this catch closer to the potential throw, explicitly around the cache check in that method so it's clearer what can do the throw.

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -60,19 +111,161 @@ private enum OutputMode
...
+
+        private static string FormatPat(PatToken pat, OutputMode output) => output switch
+        {
+            OutputMode.None => string.Empty,
+            OutputMode.Status => $"\"{pat.DisplayName}\" valid until {pat.ValidTo:O}\n",
+            OutputMode.Token => $"{pat.Token}\n",
+            OutputMode.Base64 => $"{pat.Token.Base64()}\n",
+            OutputMode.Header => $"{pat.Token.AsHeader(AzureAuth.Ado.Authorization.Basic)}\n",
+            OutputMode.HeaderValue => $"{pat.Token.AsHeaderValue(AzureAuth.Ado.Authorization.Basic)}\n",
+            OutputMode.Json => $"{pat.AsJson()}\n",
```
kyle-rader-msft, line 180: These could avoid potential usage issues by not adding newlines because the intent is for use by other programs, and forgetting to trim your strings is an easy pitfall.
```suggestion
            OutputMode.Token => $"{pat.Token}",
            OutputMode.Base64 => $"{pat.Token.Base64()}",
            OutputMode.Header => $"{pat.Token.AsHeader(AzureAuth.Ado.Authorization.Basic)}",
            OutputMode.HeaderValue => $"{pat.Token.AsHeaderValue(AzureAuth.Ado.Authorization.Basic)}",
            OutputMode.Json => $"{pat.AsJson()}",
```

src/AzureAuth/Commands/Ado/CommandPat.cs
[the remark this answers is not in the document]
kyle-rader-msft: I posit that `static readonly` from a use perspective is a "Constant" that can be initialized. So I'm fine with moving it to (a new) `Microsoft.Authentication.AzureAuth.Constants` file and calling it `AppDirectory` or something...

kyle-rader-msft (changes requested):
Lookin good, just a few small notes.

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -60,19 +111,161 @@ private enum OutputMode
...
+            var tokensHttpClientWrapper = new TokensHttpClientWrapper(connection);
+            return new PatClient(tokensHttpClientWrapper);
+        }
+
+        private IPatCache Cache()
+        {
+            var storageProperties = new StorageCreationPropertiesBuilder(
+                PatStorageParameters.CacheFileName,
+                CacheDirectory)
+            .WithMacKeyChain(
```
Haard30, line 253: Do we need to always provide `MacKeyChain` and `WithLinuxKeyring` or we can do a compile time check here and not provide for windows?

src/AzureAuth/Commands/Ado/CommandPat.cs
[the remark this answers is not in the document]
reillysiemens: Done.

src/AzureAuth/Commands/Ado/CommandPat.cs
reillysiemens: I think this is a great question! Right now I've just borrowed from existing code, so I'm inclined to keep this as is for now. However, we probably _don't_ need to be providing these values on platforms where they don't make sense. Would love to come back around to this issue in a subsequent PR and resolve it for all places in the code where this is used.

reillysiemens: I've filed #266 to remind us to do this.

src/AzureAuth/Commands/Ado/CommandPat.cs
reillysiemens: Done. I made some improvements to the shared consts as well.

src/AzureAuth/Commands/Ado/CommandPat.cs
reillysiemens: I went ahead and removed the persistence check for now and I'll come back to it in a follow up PR to improve all error handling for this subcommand.

mvanchaa (approved):

shalinikhare27 (approved):

kyle-rader-msft (approved):

PR #261, opened 2023-04-05
🏗️ [1/n] Create static `CachedAuth` to house cached auth logic

kyle-rader-msft:
This PR is the first to address an internal bug we have around doing cached auth before "real" auth.

Every auth flow (should) follow the same basic pattern:
1. Attempt to use the cache (call GetTokenSilent)
2. If the cache did not yield a usable token, then call the real method for that auth flow.

## The Bug
When there is no cache, or there are no accounts found in it attempting to call `GetTokenSilent` will always throw an `MSALUiRequiredException` because we didn't tell it which account to attempt to use silently.

This is really an MSAL usage error, and would be better represented by an `ArgumentNullException` indicating, passing null is not supported.

The effect is that we've been tracking this Exception like a normal UI Required exception which contributes to our internal monitoring, when it should not. We know that if we have a null account, we can skip cached auth and go one to the real auth method.

This first PR creates a new class that handles this case of null account by early returning null. The following PRs will update the actual auth flows to use this class.

## Upcoming Bugs in IWA (next PR)
In making this small refactor, I've discovered some other bugs which the refactor forces us to fix:
* IWA mode does not currently call IWA if the GetTokenSilent method returned `null`. It should still attempt IWA in this scenario because it's a form of cached auth not working.
* If the CompleteWithin wrapper call to GetTokenSilent returns null, it adds a null token result exception. This incorrectly causes timeouts to have a `NullTokenResultException` added which does not reflect what happened, they already add an exception to the error list for the timeout.

### DIFF
--- a/src/MSALWrapper/AuthFlow/CachedAuth.cs
+++ b/src/MSALWrapper/AuthFlow/CachedAuth.cs
@@ -49,6 +49,7 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
             catch (MsalUiRequiredException ex)
             {
                 errors.Add(ex);
+                logger.LogDebug($"Cached Auth failed:\n{ex.Message}");
             }
 
             return tokenResult;
--- a/src/MSALWrapper/AuthFlow/CachedAuth.cs
+++ b/src/MSALWrapper/AuthFlow/CachedAuth.cs
@@ -10,8 +10,9 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
     using Microsoft.Extensions.Logging;
     using Microsoft.Identity.Client;
 
-    using static Microsoft.Authentication.MSALWrapper.MsalWrapper;
-
+    /// <summary>
+    /// This class provides an abstraction over how to use the MSAL Cached Auth (known as GetTokenSilent).
+    /// </summary>
     public static class CachedAuth
     {
         /// <summary>
--- a/src/MSALWrapper/AuthFlow/CachedAuth.cs
+++ b/src/MSALWrapper/AuthFlow/CachedAuth.cs
@@ -15,6 +15,8 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
     /// </summary>
     public static class CachedAuth
     {
+        private static readonly TimeSpan CachedAuthTimeout = TimeSpan.FromSeconds(30);
+
         /// <summary>
         /// Try to get a token silently.
         /// </summary>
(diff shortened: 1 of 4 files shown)


src/MSALWrapper/AuthFlow/CachedAuth.cs
```diff
@@ -0,0 +1,58 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Identity.Client;
+
+    using static Microsoft.Authentication.MSALWrapper.MsalWrapper;
+
+    public static class CachedAuth
```
reillysiemens, line 15: This is a public class and should probably have some documentation.

reillysiemens (changes requested):

reillysiemens (approved):

src/MSALWrapper/AuthFlow/CachedAuth.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// Try to get a token silently.
+        /// </summary>
+        /// <param name="logger">An <see cref="ILogger"/>.</param>
+        /// <param name="cachedAuthTimeout">The timeout for getting a cached token.</param>
+        /// <param name="scopes">Auth Scopes.</param>
+        /// <param name="account"><see cref="IAccount"/> object to use or null if no account has been found.</param>
+        /// <param name="pcaWrapper">An <see cref="IPCAWrapper"/>.</param>
+        /// <param name="errors">List of errors to report any Exceptions into.</param>
+        /// <returns>A <see cref="Task{TResult}"/> or null if no cached token was acquired.</returns>
+        public static async Task<TokenResult> TryCachedAuthAsync(ILogger logger, TimeSpan cachedAuthTimeout, IEnumerable<string> scopes, IAccount account, IPCAWrapper pcaWrapper, IList<Exception> errors)
```
Haard30, line 28: Can we simplify this name to `TryAsync` or `ExecuteAsync`  as the class is already called `CachedAuth`?

kyle-rader-msft: A comment after my own heart. Delightful call out.

reillysiemens: Brilliant suggestion! Thanks, @Haard30!

Haard30, line 28: I feel this method has too many arguments. I don't know if it fits the scope of PR but we should look to simplify

reillysiemens: `GetTokenAsync` for parity with other proper flows? 🤔

kyle-rader-msft: Yeah, GetTokenAsync matches up nicely with where this might go soon :)

kyle-rader-msft: That's a great question. I built this fairly mechanically to extract the code as it was. Given this is always trying to use the cache, we could at the very least not require a cached auth timeout, and instead define that constant on this class. The rest would require a lot more thought about how to get that data over here.

kyle-rader-msft: A potentially easy thing to reach for would have been putting this method on the AuthFlowBase class, to be shared by all implemtnors, and move the logger, and errors into it. But, inheritance used for behavior sharing (rather than enforcement) can lead to the wrong abstraction.

This class sets us up nicely in the future to actually move the cached auth into it's own AuthFlow.

Haard30: I agree. I am fine with removing cached auth timeout parameter for now

Haard30 (approved):

reillysiemens (approved):

PR #262, opened 2023-04-05
Release 0.7.4

goagain:
This version is to enable MDM policy, and it is only for plain AzureAuth use and won't be merged to AzureAuth-lib or OMR.


mvanchaa (approved):

kyle-rader-msft (changes requested):
Holding until we unflight the ado token command.

kyle-rader-msft (approved):

PR #267, opened 2023-04-07
Prepare for Release 0.8.0

kyle-rader-msft:
# Prepare for Release `0.8.0`

This includes
* flighting the `ado` sub-commands, `token` and `pat`, and
* Removing the backwards compatibility of the root command, and printing help text instead.

### DIFF
--- a/src/AzureAuth/Commands/CommandAzureAuth.cs
+++ b/src/AzureAuth/Commands/CommandAzureAuth.cs
@@ -1,16 +1,14 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using System.IO.Abstractions;
-
-using McMaster.Extensions.CommandLineUtils;
-
-using Microsoft.Extensions.Logging;
-using Microsoft.Office.Lasso.Interfaces;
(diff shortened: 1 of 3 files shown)


src/AzureAuth/Commands/CommandAzureAuth.cs
```diff
@@ -1,34 +1,32 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-namespace Microsoft.Authentication.AzureAuth.Commands
-{
-    using System.IO.Abstractions;
```
goagain, line 6: Using should be inside namespace in our codebase, is it?

goagain (approved):

src/AzureAuth/Commands/CommandAzureAuth.cs
kyle-rader-msft: Yeah good catch.

mvanchaa (approved):

goagain (approved):
🚀🚀🚀🚀🚀🚀

PR #269, opened 2023-04-07
Ensure PAT cache keys are deterministic regardless of scope order

reillysiemens:
Before this PR, it's possible that `PatOptions` with a different ordering of `Scopes` could result in different cache keys. This is not desirable, so we now sort the scopes before generating a cache key.

### DIFF
--- a/src/AdoPat.Test/PatOptionsTest.cs
+++ b/src/AdoPat.Test/PatOptionsTest.cs
@@ -1,0 +1,50 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.AdoPat.Test
+{
+    using FluentAssertions;
+    using NUnit.Framework;
+
+    public class PatOptionsTest
+    {
+        private const string Organization = "contoso";
+        private const string DisplayName = "Test PAT";
+
+        [Test]
+        public void CacheKey_Uses_Organization_DisplayName_And_Scopes()
+        {
+            var patOptions = new PatOptions
+            {
+                Organization = Organization,
+                DisplayName = DisplayName,
+                Scopes = new[] { "test.scope.a", "test.scope.b", "test.scope.c" },
+            };
+            var expected = $"{Organization} {DisplayName} test.scope.a test.scope.b test.scope.c";
+
+            patOptions.CacheKey().Should().BeEquivalentTo(expected);
+        }
+
+        [Test]
+        public void CacheKey_Scope_Order_Is_Deterministic()
+        {
+            // These two PAT options have the same organization, display name,
+            // and scopes, but the scopes are in different orders. They should
+            // still be considered to have the same PAT cache key.
+            var patOptions1 = new PatOptions
+            {
+                Organization = Organization,
+                DisplayName = DisplayName,
+                Scopes = new[] { "test.scope.a", "test.scope.b", "test.scope.c" },
+            };
+            var patOptions2 = new PatOptions
+            {
+                Organization = Organization,
+                DisplayName = DisplayName,
+                Scopes = new[] { "test.scope.b", "test.scope.a", "test.scope.c" },
+            };
+
+            patOptions1.CacheKey().Should().BeEquivalentTo(patOptions2.CacheKey());
+        }
+    }
+}
--- a/src/AdoPat.Test/PatOptionsTest.cs
+++ b/src/AdoPat.Test/PatOptionsTest.cs
@@ -20,9 +20,9 @@ namespace Microsoft.Authentication.AdoPat.Test
                 DisplayName = DisplayName,
                 Scopes = new[] { "test.scope.a", "test.scope.b", "test.scope.c" },
             };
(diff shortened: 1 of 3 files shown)


shalinikhare27 (approved):

mvanchaa (approved):

mvanchaa (dismissed):
Discussed this offline, but commenting here. What if the org name or display names have spaces? Tucker has a solution for this - to hash them and then concatenate.

kyle-rader-msft:
> Discussed this offline, but commenting here. What if the org name or display names have spaces? Tucker has a solution for this - to hash them and then concatenate.

It looks like org name should not have spaces, but the display name certainly could

src/AdoPat/PatOptions.cs
```diff
@@ -33,9 +36,23 @@ public record PatOptions
         /// <returns>The cache key.</returns>
         public string CacheKey()
         {
-            var strings = new List<string> { this.Organization, this.DisplayName };
-            strings.AddRange(this.Scopes);
-            return string.Join(" ", strings);
+            // We concatenate scopes with the empty string after sorting and
+            // deduplicating them to ensure the cache key is always the same.
+            var sortedScopes = string.Concat(this.Scopes.ToImmutableSortedSet());
```
goagain, line 41: I love the SortedSet.

kyle-rader-msft (approved):

goagain:
I have a small question. Where does the `displayName` come from? If it is only for display purpose, should we consider it not a part of cache key?

src/AdoPat/PatOptions.cs
shalinikhare27, line 41: This is doing both sorting and deduplication, correct? So we have an empty string as a scope then it will sort with the empty string first, correct?

src/AdoPat.Test/PatOptionsTest.cs
```diff
@@ -0,0 +1,52 @@
...
+        public void CacheKey_Scope_Order_Is_Deterministic()
+        {
+            // These two PAT options have the same organization, display name,
+            // and scopes, but the scopes are in different orders. They should
+            // still be considered to have the same PAT cache key.
+            var patOptions1 = new PatOptions
+            {
+                Organization = Organization,
+                DisplayName = DisplayName,
+                Scopes = new[] { "test.scope.a", "test.scope.b", "test.scope.c" },
```
shalinikhare27, line 40: Should we consider writing a test where one of the scope is an empty string?

src/AdoPat/PatOptions.cs
Haard30, line 41: Will this take care of case sensitivity as well?

kyle-rader-msft: Yes, `ToImmutableSortedSet` is a LINQ method. I've also asked about scope validation, so I'd be happy to see handling empty string scopes in a follow up PR about scope validation.

Haard30: We should write a test for checking that as well

kyle-rader-msft: This is also a great scope _validation_ question, and I'd love to see it handled in a follow up PR.

src/AdoPat.Test/PatOptionsTest.cs
reillysiemens: I think this is good validation to do, but I'd like to do it at a higher level in the code, closer to where we parse the original command line arguments, which would ultimately eliminate the need for such a test.

reillysiemens: Rather, I don't think we should allow the empty string, but this PR doesn't address that. I'd like to handle it in a follow up.

src/AdoPat/PatOptions.cs
reillysiemens: It won't, but I think we should. Like with my comment to @shalinikhare27 earlier, I think this is probably validation that belongs at an early stage in the process. I think it's a good idea to do, but maybe not here in the code.

reillysiemens: I _think_ that's the case, but also we probably just shouldn't allow the empty string to be a scope.

Haard30: By scope validation do you mean we will not allow `Test.scope.a` or that we will convert that to lowercase there in early stage process?

reillysiemens:
@goagain: The display name is incorporated into the cache key to allow for logically separate PATs which happen to have the same org and scopes. In theory two PATs with separate display names, but the same org/scopes are the same PAT. They allow the same access. But users might like to name them differently to show what the PAT is _intended_ to be used for.

src/AdoPat/PatOptions.cs
Haard30: Either way, resolving this as we will be doing that in follow -up PR.

Haard30 (approved):

goagain (approved):

shalinikhare27 (approved):

PR #270, opened 2023-04-11
Remove uploading codeql result in release pipeline

goagain:
As [Microsoft Security User Story](security.microsoft.net), <abbr title="static code analysis">SCA</abbr> is required.
I was supposed to upload the SCA result to artifacts in release pipeline, but
1. It is not required. In ADO, CodeQL is auto injected but it doesn't upload the result either.
2. It doesn't work. In the [latest release pipeline](https://github.com/AzureAD/microsoft-authentication-cli/actions/runs/4640483566/jobs/8212474007), under `build (any platform)/Upload CodeQL Analysis Result`, it shows 
```
Warning: No files were found with the provided path: results/csharp.sarif. No artifacts will be uploaded.
```

We know this doesn't work, so we are going to remove it.
I am also trying to do some test with [Upload-sarif](https://github.com/github/codeql-action/tree/v2/upload-sarif) in a forked repo. If it works, I would like to add it again.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -74,12 +74,6 @@ jobs:
     - name: Perform CodeQL Analysis
       uses: github/codeql-action/analyze@v2
 
-    - name: Upload CodeQL Analysis Results
-      uses: actions/upload-artifact@v3
-      with:
-        name: azureauth-static-code-analysis-${{ matrix.runtime }}.sarif
-        path: results/csharp.sarif
-
     - name: Build artifacts
       run: dotnet publish src/AzureAuth/AzureAuth.csproj -p:Version=${{ github.event.inputs.version }} --configuration release --self-contained true --runtime ${{ matrix.runtime }} --output dist/${{ matrix.runtime }}
       env:


shalinikhare27:
@goagain I would love to understand the background on how you figured that it is not needed and wouldn't work.

goagain:
> @goagain I would love to understand the background on how you figured that it is not needed and wouldn't work.

@shalinikhare27 - Thank you for asking, I've added more details in the description.

goagain:
> @goagain your description is still somewhat confusing on why/how we figured this out. IIRC we saw in the workflow that the CodeQL task _did also upload_ the results, and so our extra task to attempt to upload them is not needed. But, I do not see the CodeQl results uploaded as an artifact of the last release workflow that ran for 0.8.0. So the auto-upload isn't working either?

Correct, that's what I meant. CodeQL has its own step to create a result abstraction. We wanted to make the artifacts contain the SCA result but failed.

kyle-rader-msft:
@goagain then the last thing I want to clarify is the "It's not required" I thought we did want to have the analysis uploaded with the release, otherwise - what's the value of having it? Will that task fail if it scans something important?

kyle-rader-msft (approved):

kyle-rader-msft:
@goagain I've approved, please use the "Update branch" option on GitHub to update this PR so that it doesn't dismiss approvals ;)

reillysiemens (approved):

PR #271, opened 2023-04-12
Update Microsoft.NET.Test.Sdk to 17.3.3 in test projs

Haard30:
Newtonsoft.Json prior to version 13.0.1 is vulnerable to Insecure Defaults due to improper handling of expressions with high nesting level that lead to StackOverFlow exception or high CPU and RAM usage.

Our test projects AdoPat.Test, AzureAuth.Test, MASLWrapper.Test uses `Microsoft.NET.Test.Sdk Version < 17.3.3` which uses a Newtonsoft.Json ` 9.0.1` having the above-mentioned vulnerability.

 ## Test
Run all tests in Windows and Mac to ensure nothing breaks.

More info about releases of `Microsoft.NET.Test.Sdk Version` can be found in their [project](https://github.com/microsoft/vstest/blob/main/docs/releases.md)

### DIFF
--- a/src/AdoPat.Test/AdoPat.Test.csproj
+++ b/src/AdoPat.Test/AdoPat.Test.csproj
@@ -22,7 +22,7 @@
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="6.1.0" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.2" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.3" />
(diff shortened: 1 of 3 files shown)


varshini-elayappen (approved):

goagain (approved):

mvanchaa (approved):

mvanchaa:
@Haard30 are we also using this version for `azureauth-lib`?

goagain:
> @Haard30 are we also using this version for `azureauth-lib`?

AzureAuth-Lib doesn't have such alert in CG.

PR #277, opened 2023-04-20
🤖 [2/n] Null Accounts: Update Broker

kyle-rader-msft:
# Pre Reqs
Review #261 

This PR updates the `Broker` auth flow to use the new `CachedAuth` from #261. As expected only 1 test needs to be updated since we no longer expect to call GetTokenSilent with a null account.

## Test Updates
* Move pcaMock verification into a `[TearDown]` method.
* The mock setup method `SilentAuthUIRequired` used to call `SetupWithPromptHint` for some reason. That method is not directly related as it's a part of the calling of interactive auth. So for example if silent auth returned an exception that bubbles out of the flow - the WithPromptHint mock would be set incorrectly. Separating the two makes the mock setups clearer for each test. 
* Mock setup order: I've re-ordered the calls to mock the calls in the order we expect them to happen.
* `BrokerAuthFlow_GetTokenSilent_ReturnsNull` on line 89 previously asserted after silent auth returned null that no actual auth happened, and now we do proceed with doing auth.

### DIFF
--- a/src/MSALWrapper.Test/AuthFlow/BrokerTest.cs
+++ b/src/MSALWrapper.Test/AuthFlow/BrokerTest.cs
@@ -5,19 +5,21 @@ namespace Microsoft.Authentication.MSALWrapper.Test
 {
     using System;
     using System.Collections.Generic;
-    using System.Runtime.CompilerServices;
     using System.Threading;
     using System.Threading.Tasks;
+
     using FluentAssertions;
+
     using Microsoft.Authentication.MSALWrapper;
     using Microsoft.Authentication.TestHelper;
-    using Microsoft.Extensions.DependencyInjection;
     using Microsoft.Extensions.Logging;
     using Microsoft.Identity.Client;
     using Microsoft.IdentityModel.JsonWebTokens;
+
     using Moq;
-    using NLog.Extensions.Logging;
+
     using NLog.Targets;
+
     using NUnit.Framework;
 
     public class BrokerTest
--- a/src/MSALWrapper.Test/AuthFlow/BrokerTest.cs
+++ b/src/MSALWrapper.Test/AuthFlow/BrokerTest.cs
@@ -69,7 +69,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         public AuthFlow.Broker Subject() => new AuthFlow.Broker(this.logger, ClientId, TenantId, this.scopes, pcaWrapper: this.pcaWrapperMock.Object, promptHint: this.promptHint);
 
         [Test]
-        public async Task BrokerAuthFlow_HappyPath()
+        public async Task BrokerAuthFlow_CachedAuth()
         {
             this.MockAccount();
             this.SilentAuthResult();
@@ -128,6 +128,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         {
             this.MockAccount();
             this.SilentAuthUIRequired();
+            this.SetupWithPromptHint();
             this.InteractiveAuthResultReturnsNullWithoutClaims();
 
             // Act
@@ -180,8 +181,8 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         {
             this.MockAccount();
             this.SilentAuthTimeout();
-            this.SetupWithPromptHint();
             this.InteractiveAuthResult();
+            this.SetupWithPromptHint();
 
             // Act
             AuthFlow.Broker broker = this.Subject();
@@ -255,6 +256,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         {
             this.MockAccount();
             this.SilentAuthUIRequired();
+            this.SetupWithPromptHint();
             this.InteractiveAuthExtraClaimsRequired();
             this.InteractiveAuthResultReturnsNullWithClaims();
 
--- a/src/MSALWrapper.Test/AuthFlow/BrokerTest.cs
+++ b/src/MSALWrapper.Test/AuthFlow/BrokerTest.cs
@@ -71,9 +71,8 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         [Test]
         public async Task BrokerAuthFlow_HappyPath()
         {
-            this.SilentAuthResult();
-
             this.MockAccount();
+            this.SilentAuthResult();
 
             // Act
             AuthFlow.Broker broker = this.Subject();
@@ -91,7 +90,7 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         {
             this.MockAccount();
             this.SilentAuthReturnsNull();
-            this.SetupInteractiveAuthWithPromptHint();
+            this.SetupWithPromptHint();
             this.InteractiveAuthResult();
 
             // Act
@@ -107,10 +106,10 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         [Test]
         public async Task BrokerAuthFlow_MsalUIException()
         {
-            this.SilentAuthUIRequired();
-            this.InteractiveAuthResult();
-
             this.MockAccount();
+            this.SilentAuthUIRequired();
+            this.SetupWithPromptHint();
+            this.InteractiveAuthResult();
 
             // Act
             AuthFlow.Broker broker = this.Subject();
@@ -127,10 +126,9 @@ namespace Microsoft.Authentication.MSALWrapper.Test
         [Test]
         public async Task BrokerAuthFlow_MsalUIException_InteractiveAuthResultReturnsNullWithoutClaims()
         {
+            this.MockAccount();
             this.SilentAuthUIRequired();
             this.InteractiveAuthResultReturnsNullWithoutClaims();
-
-            this.MockAccount();
 
             // Act
             AuthFlow.Broker broker = this.Subject();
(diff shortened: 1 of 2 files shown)


src/MSALWrapper.Test/AuthFlow/BrokerTest.cs
```diff
@@ -172,27 +176,26 @@ public async Task BrokerAuthFlow_GetTokenSilent_MsalServiceException()
...
-            this.pcaWrapperMock.VerifyAll();
             authFlowResult.TokenResult.Should().Be(null);
             authFlowResult.Errors.Should().HaveCount(1);
             authFlowResult.Errors[0].Should().BeOfType(typeof(MsalServiceException));
             authFlowResult.AuthFlowName.Should().Be("broker");
         }
 
         [Test]
-        public async Task BrokerAuthFlow_GetTokenSilent_OperationCanceledException()
+        public async Task BrokerAuthFlow_GetTokenSilent_OperationCanceledException_ThenSucceeds()
```
reillysiemens, line 186: Should we have a corresponding `ThenFails` test? Or maybe a `ThenMSALReturnsNull`?

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -153,7 +142,7 @@ protected override async Task<(TokenResult, IList<Exception>)> GetTokenInnerAsyn
                 this.errors.Add(ex);
             }
 
-            return (null, this.errors);
+            return (tokenResult, this.errors);
```
reillysiemens, line 145: I think that we could just return `tokenResult` and modify the `AuthFlowBase` class to always return `this.errors`. This is a separate change though, not entirely related to this PR, so I could see it done in a follow up.

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -84,58 +84,47 @@ protected override async Task<(TokenResult, IList<Exception>)> GetTokenInnerAsyn
...
-                try
+                tokenResult = await CachedAuth.TryCachedAuthAsync(
+                    this.logger,
+                    this.silentAuthTimeout,
+                    this.scopes,
+                    account,
+                    this.pcaWrapper,
+                    this.errors);
+
+                if (tokenResult == null)
```
reillysiemens, line 99: You could early return here and avoid nesting the giant `try` block below that way. Just something to consider.

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -84,58 +84,47 @@ protected override async Task<(TokenResult, IList<Exception>)> GetTokenInnerAsyn
...
+                        tokenResult = await TaskExecutor.CompleteWithin(
                             this.logger,
-                            this.silentAuthTimeout,
-                            "Get Token Silent",
-                            (cancellationToken) => this.pcaWrapper.GetTokenSilentAsync(this.scopes, account, cancellationToken),
+                            this.interactiveAuthTimeout,
+                            "Interactive Auth",
+                            (cancellationToken) => this.pcaWrapper
+                            .WithPromptHint(this.promptHint)
+                            .GetTokenInteractiveAsync(this.scopes, account, cancellationToken),
```
reillysiemens, line 109: The indentation here doesn't make it clear that these methods are being called on the `pcaWrapper`. This might also be moved into a helper method to make this code block smaller and more readable.

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -84,58 +84,47 @@ protected override async Task<(TokenResult, IList<Exception>)> GetTokenInnerAsyn
...
                             this.errors)
                             .ConfigureAwait(false);
```
reillysiemens, line 111: The indentation of `.ConfigureAwait(false)` here doesn't make it clear at which layer it's being called. Might help to dedent it one level.

src/MSALWrapper/AuthFlow/Broker.cs
kyle-rader-msft: Discussed offline - we like this idea, but will address in a separate PR.

src/MSALWrapper/AuthFlow/Broker.cs
kyle-rader-msft: Will address in the follow up to not have to return the tuple.

src/MSALWrapper.Test/AuthFlow/BrokerTest.cs
kyle-rader-msft: The primary difference is already tested where silent auth doesn't work - and interactive auth also doesn't work, returning null or throwing exceptions - both of those are covered in tests below.

reillysiemens (approved):

reillysiemens (dismissed):

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -84,57 +77,52 @@ protected override async Task<(TokenResult, IList<Exception>)> GetTokenInnerAsyn
...
+                        $"{this.Name()} interactive auth",
+                        getTokenInteractive,
+                        this.errors).ConfigureAwait(false);
                 }
                 catch (MsalUiRequiredException ex)
                 {
                     this.errors.Add(ex);
-                    this.logger.LogDebug($"Silent auth failed, re-auth is required.\n{ex.Message}");
-                    var tokenResult = await TaskExecutor.CompleteWithin(
+                    this.logger.LogDebug($"initial {this.Name()} auth failed. Trying again with claims from exception.\n{ex.Message}");
```
shalinikhare27, line 113:
```suggestion
                    this.logger.LogDebug($"Initial {this.Name()} auth failed. Trying again with claims from exception.\n{ex.Message}");
```

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -84,57 +77,52 @@ protected override async Task<(TokenResult, IList<Exception>)> GetTokenInnerAsyn
...
-                            "Interactive Auth",
-                            (cancellationToken) => this.pcaWrapper
-                            .WithPromptHint(this.promptHint)
-                            .GetTokenInteractiveAsync(this.scopes, account, cancellationToken),
-                            this.errors)
-                            .ConfigureAwait(false);
+                    return (tokenResult, this.errors);
+                }
+
+                Func<CancellationToken, Task<TokenResult>> getTokenInteractive = (cancellationToken) =>
```
shalinikhare27, line 96: To further clean things up, can we refactor this to its own private Func?

```
private Func<CancellationToken, Task<TokenResult>> getTokenInteractive = (cancellationToken) => 
                        this.pcaWrapper
                        .WithPromptHint(this.promptHint)
                        .GetTokenInteractiveAsync(this.scopes, account, cancellationToken);
```

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -84,57 +77,52 @@ protected override async Task<(TokenResult, IList<Exception>)> GetTokenInnerAsyn
...
                 catch (MsalUiRequiredException ex)
                 {
                     this.errors.Add(ex);
-                    this.logger.LogDebug($"Silent auth failed, re-auth is required.\n{ex.Message}");
-                    var tokenResult = await TaskExecutor.CompleteWithin(
+                    this.logger.LogDebug($"initial {this.Name()} auth failed. Trying again with claims from exception.\n{ex.Message}");
+
+                    Func<CancellationToken, Task<TokenResult>> getTokenInteractiveWithClaims = (cancellationToken) =>
+                        this.pcaWrapper
+                            .WithPromptHint(this.promptHint)
```
shalinikhare27, line 117: Same comment. We can probably move this as a private Func.

kyle-rader-msft: Yes we could define each of these as normal funcs.

src/MSALWrapper/AuthFlow/Broker.cs
shalinikhare27, line 113: We can assert these logs in the tests.

src/MSALWrapper/AuthFlow/Broker.cs
kyle-rader-msft: Oh, I missed that these capture the account, so no this is a little more complicated. Trying to define a function would require either giving it access to state, which is possible, or passing the account to create a lambda function and return that. I'll play with a little.

kyle-rader-msft: Yes, we can do this by passing the account and returning the Func, and I like it. I'll push shortly.

src/MSALWrapper/AuthFlow/Broker.cs
kyle-rader-msft: We can, but we do not currently. I agree it would be good to make log assertions here and I'd like to add those after the impending test refactors since _most_ of roughly half of the tests will be moved or deleted.

kyle-rader-msft: @shalinikhare27 does that work for you?

kyle-rader-msft: I've opened #282 to track that.

shalinikhare27 (approved):

reillysiemens (approved):

PR #278, opened 2023-04-20
🕸️ [3/n] Null Accounts: Update Web to use CachedAuth

kyle-rader-msft:
# PreReqs
Review #261 

## Update to use Cached Auth

This upgrades Web to use the Cached Auth. 
I have re-ordered the test mocks to setup the mocks in the order we expect them to be called. This doesn't have a functional impact but I think makes reasoning about the tests easier.

### DIFF
--- a/src/MSALWrapper/AuthFlow/Web.cs
+++ b/src/MSALWrapper/AuthFlow/Web.cs
@@ -63,35 +63,24 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
         /// <inheritdoc/>
         protected override async Task<(TokenResult, IList<Exception>)> GetTokenInnerAsync()
         {
-            IAccount account = await this.pcaWrapper.TryToGetCachedAccountAsync(this.preferredDomain) ?? null;
-
-            if (account != null)
-            {
-                this.logger.LogDebug($"Using cached account '{account.Username}'");
-            }
+            IAccount account = await this.pcaWrapper.TryToGetCachedAccountAsync(this.preferredDomain);
+            TokenResult tokenResult = null;
 
             try
             {
                 try
                 {
-                    try
-                    {
-                        var tokenResult = await TaskExecutor.CompleteWithin(
-                            this.logger,
-                            this.silentAuthTimeout,
-                            "Get Token Silent",
-                            (cancellationToken) => this.pcaWrapper.GetTokenSilentAsync(this.scopes, account, cancellationToken),
-                            this.errors)
-                            .ConfigureAwait(false);
-                        tokenResult.SetSilent();
+                    tokenResult = await CachedAuth.TryCachedAuthAsync(
+                        this.logger,
+                        this.silentAuthTimeout,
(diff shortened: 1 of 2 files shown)


src/MSALWrapper/AuthFlow/Web.cs
```diff
@@ -117,8 +104,6 @@ protected override async Task<(TokenResult, IList<Exception>)> GetTokenInnerAsyn
                         .GetTokenInteractiveAsync(this.scopes, ex.Claims, cancellationToken),
                         this.errors)
                         .ConfigureAwait(false);
-
-                    return (tokenResult, this.errors);
                 }
             }
             catch (MsalServiceException ex)
```
reillysiemens, line 109: Can we collapse these catch statements into a single one?

reillysiemens (commented):
My comments from #277 are also applicable here, but none of them are blocking.

src/MSALWrapper/AuthFlow/Web.cs
shalinikhare27: I think I did not fully understand what you are proposing. We are catching different exceptions to add to the error list. Please elaborate. :)

src/MSALWrapper/AuthFlow/Web.cs
```diff
@@ -63,62 +54,52 @@ public Web(ILogger logger, Guid clientId, Guid tenantId, IEnumerable<string> sco
...
+                        $"{this.Name()} interactive auth",
+                        interactiveAuth,
+                        this.errors).ConfigureAwait(false);
                 }
                 catch (MsalUiRequiredException ex)
                 {
                     this.errors.Add(ex);
-                    this.logger.LogDebug($"Silent auth failed, re-auth is required.\n{ex.Message}");
-                    var tokenResult = await TaskExecutor.CompleteWithin(
+                    this.logger.LogDebug($"Initial ${this.Name()} auth failed. Trying again with claims.\n{ex.Message}");
```
shalinikhare27, line 91: This log assertion is missing in tests.

src/MSALWrapper/AuthFlow/Web.cs
kyle-rader-msft: This will be handled in one of the following PRs (it's already done)

src/MSALWrapper/AuthFlow/Web.cs
kyle-rader-msft: It always has been, I'd like to forgo adding it here, until after the test revamp since many of the tests will move or go away (Same for the other 3 active PRs to use CachedAuth.GetTokenAsync)

shalinikhare27 (approved):

reillysiemens (approved):

src/MSALWrapper/AuthFlow/Web.cs
kyle-rader-msft: Resolving with the follow up item #282

PR #279, opened 2023-04-20
🪟 [4/n] Null Accounts: Update IWA to use CachedAuth

kyle-rader-msft:
# PreReqs
Review #261 

# Bug Fixes
* The base bug fix of no longer collecting an extra msal ui required exception
* IWA was previously not actually calling IWA in the event that cached auth returned null or timed out because neither of those are MSALUiRequired Exceptions.

### DIFF
--- a/src/MSALWrapper.Test/AuthFlow/IntegratedWindowsAuthenticationTest.cs
+++ b/src/MSALWrapper.Test/AuthFlow/IntegratedWindowsAuthenticationTest.cs
@@ -7,17 +7,18 @@ namespace Microsoft.Authentication.MSALWrapper.Test
     using System.Collections.Generic;
     using System.Threading;
     using System.Threading.Tasks;
+
     using FluentAssertions;
+
     using Microsoft.Authentication.MSALWrapper;
     using Microsoft.Authentication.MSALWrapper.AuthFlow;
     using Microsoft.Authentication.TestHelper;
-    using Microsoft.Extensions.DependencyInjection;
     using Microsoft.Extensions.Logging;
     using Microsoft.Identity.Client;
     using Microsoft.IdentityModel.JsonWebTokens;
+
     using Moq;
-    using NLog.Extensions.Logging;
-    using NLog.Targets;
+
     using NUnit.Framework;
 
     public class IntegratedWindowsAuthenticationTest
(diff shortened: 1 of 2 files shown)


reillysiemens (commented):
The comments from #277 apply here as well, otherwise I don't see any glaring issues.

src/MSALWrapper.Test/AuthFlow/IntegratedWindowsAuthenticationTest.cs
```diff
@@ -81,36 +86,33 @@ public async Task GetCachedToken_ReturnsNull()
...
             // Act
             AuthFlow.IntegratedWindowsAuthentication iwa = this.Subject();
             var authFlowResult = await iwa.GetTokenAsync();
 
             // Assert
-            this.pcaWrapperMock.VerifyAll();
-            authFlowResult.TokenResult.Should().Be(null);
-            authFlowResult.Errors.Should().HaveCount(1);
+            authFlowResult.TokenResult.Should().Be(this.tokenResult);
+            authFlowResult.Errors.Should().BeEmpty();
```
shalinikhare27, line 97: We should also assert the following:
`authFlowResult.TokenResult.IsSilent.Should().BeTrue();`

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -53,51 +52,46 @@ public IntegratedWindowsAuthentication(ILogger logger, Guid clientId, Guid tenan
...
+                    iwaAuth,
+                    this.errors).ConfigureAwait(false);
+
+                // If IWA worked, it was 100% silent.
+                tokenResult.SetSilent();
             }
             catch (MsalUiRequiredException ex)
             {
                 this.errors.Add(ex);
+                this.logger.LogDebug(ex.Message);
```
shalinikhare27, line 88: Consider adding log assertions in tests.

kyle-rader-msft: Will do post test refactor

reillysiemens (approved):

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
kyle-rader-msft: Resolving with #282

PR #281, opened 2023-04-21
Cleanly log catastrophic errors

reillysiemens:
Right now the `azureauth ado pat` subcommand lets users see the full stack trace for a catastrophic error. This PR presents that in a cleaner fashion and also makes sure that errors are included in telemetry when applicable.


reillysiemens:
Actually, this can also be resolved by just updating our internal dependency on the Lasso framework, so I'll do that instead.

kyle-rader-msft (approved):
🪅

shalinikhare27 (approved):

PR #283, opened 2023-04-26
Update vulnerable python libraries

mvanchaa:
The python script `trigger_azure_pipelines.py` is using libraries `oauthlib==3.2.1` and `certifi==2022.9.24`. They need to be updated because they are found to be security vulnerabilities. This PR addresses that change.


Haard30 (approved):

reillysiemens (changes requested):
The diff here shows as `+0,-0` for me, which seems strange. Any idea what's up with that, @mvanchaa?

kyle-rader-msft:
Something happened to the requirements.txt file to make Git treat it as `BIN` (binary) file. It should be a normal text file with UTF-8 encoding.

kyle-rader-msft (changes requested):
Fix encoding or git treatment of requirements.txt file to be text.

mvanchaa:
> The diff here shows as `+0,-0` for me, which seems strange. Any idea what's up with that, @mvanchaa?

@reillysiemens @kyle-rader-msft this happened before and I raised a feedback with GitHub long ago. The contents under `bin` are probably being considered as binary files. We also git ignored the contents of the bin. Hence, we cannot see it here. I remember there was a reason why we git ignored it but I forgot why. @reillysiemens should I remove the contents of bin from the .gitignore file?

mvanchaa:
So this file was initially utf-16 encoded and I converted it to utf-8 but I still see it as binary file. :(

.gitattributes
```diff
@@ -1,2 +1,4 @@
 # Do not do any text (line ending) normalization
-* -text
\ No newline at end of file
+* -text
+# Generate diff of .txt files as text file instead of binary file
+*.txt diff
```
kyle-rader-msft, line 4: This should happen by default I think? Was the cause for being treated as a binary the fact that it was in the /bin folder? or was the encoding UTF-16 or something?

Also do we intend to move the location of the requirements files?

mvanchaa: It could be due to being ignored. Because I changed the encoding to UTF-8. About moving the requirements file - I was testing if it shows the diff if requirements is moved out of bin and yes it does show diff properly as text.

mvanchaa: Even if I excluded top level `bin` folder from gitgnore, it still shows up as binary :(

kyle-rader-msft: Other files under `bin` are not treated as binary. I expect being UTF-16 originally is the cause. 
I would try removing the file, forcing that with `git rm -f` and then re-adding it as utf-8 in a new commit so that it's a new file.

PR #284, opened 2023-04-27
[2/n] Make CachedAuth an AuthFlow

kyle-rader-msft:
Make `CachedAuth` inherit from `AuthFlowBase` and it's own "auth flow". 
This new flow is added to flows in the AuthFlowFactory, always, regardless of mode, to maintain existing behavior, where no matter what mode you choose, you get cached auth first.

This change allows us to apply the common error handling of the `AuthFlowBase` to the cached auth flow the same way it's applied to all other flows.

## `AuthFlowTestBase`
This is a new class in the test project to house the common setup for testing auth flows. 
Following PRs will re-use this to help dedupe the large amount of duplicated test setup we have accumlated across the auth flows, similar to how the error handling being moved into the base class has simplified all the authflows.


goagain (approved):

mvanchaa (approved):

Haard30 (approved):
I like this refactoring :)

PR #285, opened 2023-04-27
Improve ADO PAT debug logging

reillysiemens:
There was previously no debug logging within the `ado pat` subcommand. This PR addresses that and gives users information with `--verbosity debug` about

- Whether a PAT was found in the cache.
- Whether a PAT was inactive.
- Whether a PAT was expiring soon.
- Whether a PAT was created.
- Whether a PAT was regenerated.

In the event that we do suffer failures this information should help us better diagnose the errors. Catastrophic failures are already sent in telemetry (if and **only** if telemetry is enabled).

### DIFF
--- a/src/AdoPat.Test/PatManagerTest.cs
+++ b/src/AdoPat.Test/PatManagerTest.cs
@@ -9,6 +9,7 @@ namespace Microsoft.Authentication.AdoPat.Test
     using System.Threading.Tasks;
     using FluentAssertions;
     using Microsoft.Authentication.TestHelper;
+    using Microsoft.Extensions.Logging;
     using Microsoft.VisualStudio.Services.DelegatedAuthorization;
     using Moq;
     using NLog.Targets;
@@ -34,7 +35,7 @@ namespace Microsoft.Authentication.AdoPat.Test
 
         private readonly string cacheKey = string.Join('-', OrganizationHash, DisplayNameHash, ScopeHash);
 
-        private Extensions.Logging.ILogger logger;
+        private ILogger logger;
         private MemoryTarget logTarget;
 
         // Common mocks which are configured via the Setup and Teardown methods.
--- a/src/AdoPat.Test/PatManagerTest.cs
+++ b/src/AdoPat.Test/PatManagerTest.cs
@@ -8,8 +8,10 @@ namespace Microsoft.Authentication.AdoPat.Test
     using System.Threading;
     using System.Threading.Tasks;
     using FluentAssertions;
+    using Microsoft.Authentication.TestHelper;
     using Microsoft.VisualStudio.Services.DelegatedAuthorization;
     using Moq;
+    using NLog.Targets;
     using NUnit.Framework;
 
     public class PatManagerTest
@@ -30,6 +32,11 @@ namespace Microsoft.Authentication.AdoPat.Test
         // This is a dummy authorization ID, not valid in any real contexts.
         private readonly Guid authorizationId = new Guid("ee0c5586-a96f-4a44-b1d9-8613028b1078");
 
+        private readonly string cacheKey = string.Join('-', OrganizationHash, DisplayNameHash, ScopeHash);
+
+        private Extensions.Logging.ILogger logger;
+        private MemoryTarget logTarget;
+
         // Common mocks which are configured via the Setup and Teardown methods.
         private Mock<IPatCache> cache;
         private Mock<IPatClient> client;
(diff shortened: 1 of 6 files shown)


goagain (dismissed):

src/AdoPat.Test/PatManagerTest.cs
```diff
@@ -176,6 +198,13 @@ public async Task GetPatAsync_RegeneratesPatWhenPatHasExpired()
 
             // Assert
             pat.Should().BeEquivalentTo(expectedPat);
+
+            this.logTarget.Logs.Count.Should().Be(5);
+            this.logTarget.Logs[0].Should().Be($"Checking for PAT in cache with key '{this.cacheKey}'");
```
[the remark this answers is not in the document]
reillysiemens: TL;DR - I don't think this is an issue, but if you disagree it's good to pause and think about it more.

The cache key need not be confidential because it consists only of information that is supplied on the command line when the program is run, namely `organization`, `displayName`, and `scopes`. The formula for creating the cache key is to take the SHA256 hash of the aforementioned pieces of information and concatenate them with a `-`.

https://github.com/AzureAD/microsoft-authentication-cli/blob/cc9319143a6b15d8523616d9d3f0f478ff18150e/src/AdoPat/PatOptions.cs#L37-L56

Even if it were important to safeguard that information it is my understanding that the [SHA256](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.sha256?view=net-8.0) hashing algorithm is cryptographically secure. The original use for hashing the cache key was not actually security, but rather to ensure that the characters were fixed length, valid JSON keys, and to prevent accidental overlap where two similar keys might conflict.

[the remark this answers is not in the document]
reillysiemens: I should also note that the cache itself is still encrypted and locked with the user's credentials.

Haard30 (approved):

src/AdoPat/PatManager.cs
```diff
@@ -19,19 +20,24 @@ public class PatManager
         private const int ValidToExtensionDays = 7;
         private const int NearingExpirationDays = 2;
 
+        // Note: Logging in this class intentionally avoids referencing any PAT details lest those leak into log files.
```
mvanchaa, line 23: @reillysiemens I did not get this comment. Can you elaborate this a little?

mvanchaa (dismissed):

src/AdoPat/PatManager.cs
reillysiemens: Sure. Lemme take a stab a rephrasing this and if it's suitable to you I'll update the comment.
```csharp
// Note: Do NOT use this logger instance to log any fields on a PAT, ESPECIALLY the token.
```

shalinikhare27: Yes, it's simpler english I think :)

shalinikhare27: Basically you are saying do not log PAT values?

src/AdoPat.Test/PatManagerTest.cs
```diff
@@ -30,6 +32,11 @@ public class PatManagerTest
         // This is a dummy authorization ID, not valid in any real contexts.
         private readonly Guid authorizationId = new Guid("ee0c5586-a96f-4a44-b1d9-8613028b1078");
 
+        private readonly string cacheKey = string.Join('-', OrganizationHash, DisplayNameHash, ScopeHash);
+
+        private Extensions.Logging.ILogger logger;
```
shalinikhare27, line 37: Can you not just say `ILogger logger` and put the rest in usings?

shalinikhare27 (approved):

src/AdoPat.Test/PatManagerTest.cs
reillysiemens: Uhh... maybe? I just copied what was done in another similar test class.

https://github.com/AzureAD/microsoft-authentication-cli/blob/cc9319143a6b15d8523616d9d3f0f478ff18150e/src/MSALWrapper.Test/LockedTest.cs#L23

src/AdoPat/PatManager.cs
reillysiemens: Yes, exactly. Anything logged could end up in a log file. Best to just never log any part of a PAT.

Haard30 (approved):

mvanchaa (approved):

shalinikhare27 (approved):

goagain (approved):
🚀🚀🚀

PR #286, opened 2023-05-02
[1/n] Consolidate Error Handling into `AuthFlowBase`

kyle-rader-msft:
After moving the cached auth into a static function, we can start to see the outer MSAL error handling that is common to all authflows. This PR moves that common handling into the base class along with the ownership of the `errors` list.

Namely:
* We add an `IList<Exception> errors` to the `AuthFlowBase` class.
* We remove it from each auth flow so as to not "shadow" the base member.
* We extend the GetTokenAsync method to include the error handling so that all logging, and error collection is done in that 1 place for the suite of MSAL errors thrown by any AuthFlow.

I recommend viewing these diffs using "Side by Side" view.

### DIFF
--- a/src/MSALWrapper/AuthFlow/AuthFlowBase.cs
+++ b/src/MSALWrapper/AuthFlow/AuthFlowBase.cs
@@ -13,11 +13,17 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
     /// </summary>
     public abstract class AuthFlowBase : IAuthFlow
     {
+        /// <summary>
+        /// The list of errors encountered during token acquisition.
+        /// </summary>
+        protected IList<Exception> errors = new List<Exception>();
+
         /// <inheritdoc/>
         public async Task<AuthFlowResult> GetTokenAsync()
         {
-            (var result, var errors) = await this.GetTokenInnerAsync();
-            return new AuthFlowResult(result, errors, this.Name());
+            this.errors = new List<Exception>();
+            var result = await this.GetTokenInnerAsync();
+            return new AuthFlowResult(result, this.errors, this.Name());
         }
 
         /// <summary>
@@ -25,7 +31,7 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
         /// Performs token acquisition.
         /// </summary>
         /// <returns>a tuple of <see cref="TokenResult"/> and <see cref="IList{Exception}"/>.</returns>
-        protected abstract Task<(TokenResult result, IList<Exception> errors)> GetTokenInnerAsync();
+        protected abstract Task<TokenResult> GetTokenInnerAsync();
 
         /// <summary>
         /// The name of this AuthFlow.
--- a/src/MSALWrapper/AuthFlow/AuthFlowBase.cs
+++ b/src/MSALWrapper/AuthFlow/AuthFlowBase.cs
@@ -27,6 +27,7 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
             typeof(MsalUiRequiredException),
             typeof(MsalClientException),
             typeof(MsalServiceException),
+            typeof(MsalException),
             typeof(NullReferenceException),
         };
 
--- a/src/MSALWrapper/AuthFlow/AuthFlowBase.cs
+++ b/src/MSALWrapper/AuthFlow/AuthFlowBase.cs
@@ -3,6 +3,9 @@
 
 namespace Microsoft.Authentication.MSALWrapper.AuthFlow
 {
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Identity.Client;
+
     using System;
     using System.Collections.Generic;
     using System.Threading.Tasks;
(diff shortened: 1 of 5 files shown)


shalinikhare27 (approved):

goagain (approved):

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -73,65 +70,46 @@ private enum GetAncestorFlags
...
+            catch (MsalUiRequiredException ex)
             {
-                this.logger.LogWarning($"Msal unexpected null reference! (Not Expected)\n{ex.Message}");
                 this.errors.Add(ex);
+                this.logger.LogDebug($"initial {this.Name()} auth failed. Trying again with claims from exception.\n{ex.Message}");
+
+                tokenResult = await TaskExecutor.CompleteWithin(
+                    this.logger,
+                    this.interactiveAuthTimeout,
+                    $"{this.Name()} interactive auth (with extra claims)",
```
goagain, line 107: Now we have a base class. Is it a good time to change it to a property `this.Name`?

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -73,65 +70,46 @@ private enum GetAncestorFlags
...
+                    this.interactiveAuthTimeout,
+                    $"{this.Name()} interactive auth",
+                    this.GetTokenInteractive(account),
+                    this.errors).ConfigureAwait(false);
             }
-            catch (NullReferenceException ex)
+            catch (MsalUiRequiredException ex)
             {
-                this.logger.LogWarning($"Msal unexpected null reference! (Not Expected)\n{ex.Message}");
                 this.errors.Add(ex);
```
mvanchaa, line 101: Do we still need to add exceptions to the local errors list? I'm trying to figure out how these errors are propagated to the base class. I see that any `MsalUiRequiredException` from here wouldn't propagate up. Am I missing something?

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -73,65 +70,46 @@ private enum GetAncestorFlags
...
-                        this.interactiveAuthTimeout,
-                        $"{this.Name()} interactive auth (with extra claims)",
-                        this.GetTokenInteractiveWithClaims(ex.Claims),
-                        this.errors).ConfigureAwait(false);
-                }
-            }
-            catch (MsalServiceException ex)
+            if (tokenResult != null)
             {
-                this.logger.LogWarning($"MSAL Service Exception! (Not expected)\n{ex.Message}");
```
mvanchaa, line 120: Is it okay if we miss out on providing the warnings like this?

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -73,65 +70,46 @@ private enum GetAncestorFlags
...
-                this.errors.Add(ex);
+                tokenResult = await TaskExecutor.CompleteWithin(
+                    this.logger,
+                    this.interactiveAuthTimeout,
+                    $"{this.Name()} interactive auth",
+                    this.GetTokenInteractive(account),
+                    this.errors).ConfigureAwait(false);
             }
-            catch (NullReferenceException ex)
+            catch (MsalUiRequiredException ex)
```
mvanchaa, line 99: Line 101, do we still need to add exceptions to the local errors list? I'm trying to figure out how these errors are propagated to the base class. I see that any `MsalUiRequiredException` from here wouldn't propagate up. Am I missing something?

src/MSALWrapper/AuthFlow/Broker.cs
shalinikhare27: @mvanchaa - In this case, we have a token so we don't need any warning. We can return early directly. The logic for specific exceptions have been moved below.

src/MSALWrapper/AuthFlow/AuthFlowBase.cs
```diff
@@ -13,19 +16,46 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
...
+
+        /// <summary>A <see cref="ILogger"/> to use for logging.</summary>
+        protected ILogger logger;
+
+        private static readonly HashSet<Type> ExceptionsToCatch = new HashSet<Type>()
+        {
+            typeof(MsalUiRequiredException),
+            typeof(MsalClientException),
+            typeof(MsalServiceException),
+            typeof(MsalException),
```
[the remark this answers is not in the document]
kyle-rader-msft: Not necessarily. They are the ones were we catching in each authflow though, that we are aware of/have encountered in the wild.

src/MSALWrapper/AuthFlow/Broker.cs
kyle-rader-msft: Abstract properties do look to be possible! we can make that change in a follow up PR! That would simplify each authflow a tad.

src/MSALWrapper/AuthFlow/Broker.cs
kyle-rader-msft: We still warn, in the base class's method. So all warnings done consistently.

src/MSALWrapper/AuthFlow/Broker.cs
kyle-rader-msft: Previously we had two levels of `catch (MsalUIRequiredException)` directly nested. We still have that same concept, but now the outer catch happens in the base class method `GetTokenAsync`. This inner catch is still required for `broker` and `web` modes where we might need to try doing interactive auth twice, the second time with extra claims returned in the exception. When we call again with those extra claims, an MSAL UI Required exception can be thrown again, if say, the error was not about claims, but instead that MFA was required, and they failed to pass it.

We could probably streamline this better in the future by checking if there are claims in the exception - and only do this extra attempt if there are claims. If not, we could rethrow and let the base class handle catching and logging it.

Haard30 (approved):

src/MSALWrapper/AuthFlow/Broker.cs
kyle-rader-msft: @mvanchaa I'm going to merge to keep going on these, but circle back please if this doesn't answer your question and we can discuss :)

mvanchaa: Thanks for explaining @kyle-rader-msft. You answered my question.

src/MSALWrapper/AuthFlow/Broker.cs
mvanchaa: I only see the debug statements though :o @kyle-rader-msft

PR #287, opened 2023-05-03
[1/🗑️] Remove `CachedAuth` usage from IWA

kyle-rader-msft:
Now that `CachedAuth` is its own authflow. Each other authflow (except `Broker`) can remove the initial use of CachedAuth. This PR starts that process with IWA. The other's will follow with their own PRs given how much change that induces in each test.

## Broker
So the Broker authflow is a little special. It still needs to call the "CachedAuth" pattern, because when broker is enabled this CachedAuth (GetTokenSilent) actually can succeed, even if there was no valid MSAL file cache entry. Because the broker has it's own separate internal caching.

So, we leave the call to CachedAuth there, which works nicely with the static method entry popint for cached auth, since we can pass in the broker enabled PCA.

### DIFF
--- a/src/MSALWrapper/AuthFlow/Web.cs
+++ b/src/MSALWrapper/AuthFlow/Web.cs
@@ -80,7 +80,6 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
                 this.errors.Add(ex);
                 this.logger.LogDebug($"Initial ${this.Name()} auth failed. Trying again with claims.\n{ex.Message}");
 
-
                 tokenResult = await TaskExecutor.CompleteWithin(
                     this.logger,
                     this.interactiveAuthTimeout,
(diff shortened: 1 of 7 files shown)


src/MSALWrapper/AuthFlow/Web.cs
```diff
@@ -80,7 +80,6 @@ protected override async Task<TokenResult> GetTokenInnerAsync()
                 this.errors.Add(ex);
                 this.logger.LogDebug($"Initial ${this.Name()} auth failed. Trying again with claims.\n{ex.Message}");
 
-
```
shalinikhare27, line 83: This change can be reverted. It's seem line removed by mistake.

shalinikhare27 (approved):

src/MSALWrapper/AuthFlow/Web.cs
kyle-rader-msft: I will address that in `[9/🧹]` - a final linting PR to catch up on all the small linting warnings.

goagain (approved):

mvanchaa (approved):

PR #289, opened 2023-05-03
[2/🗑️] Remove `CachedAuth` usage from Web.

kyle-rader-msft:
This removes the `CachedAuth` usage from `Web` and removes all the GetTokenSilent related tests from WebTest.

Most of the setup methods for mocking behavior on a PCA are not specific to Web and are also usable by broker, and device code modes so these are moved to the `AuthFlowTestBase` class.

### DIFF
--- a/src/MSALWrapper/AuthFlow/AuthFlowBase.cs
+++ b/src/MSALWrapper/AuthFlow/AuthFlowBase.cs
@@ -54,6 +54,7 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
         /// The inner method required to be implemented by an AuthFlow.
         /// Performs token acquisition.
         /// </summary>
+        /// <param name="account">The <see cref="IAccount"/> to attempt to use.</param>
         /// <returns>a tuple of <see cref="TokenResult"/> and <see cref="IList{Exception}"/>.</returns>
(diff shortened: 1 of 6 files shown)


src/MSALWrapper/AuthFlow/AuthFlowBase.cs
```diff
@@ -54,6 +54,7 @@ public async Task<AuthFlowResult> GetTokenAsync()
         /// The inner method required to be implemented by an AuthFlow.
         /// Performs token acquisition.
         /// </summary>
+        /// <param name="account">The <see cref="IAccount"/> to attempt to use.</param>
```
shalinikhare27, line 57: Is this added by mistake? I do not see any param as `account`.

shalinikhare27 (approved):

src/MSALWrapper/AuthFlow/AuthFlowBase.cs
kyle-rader-msft: Yes, removing in `[9/🧹]` :), good catch!

Haard30 (approved):
🚀

mvanchaa (approved):

goagain (approved):

PR #296, opened 2023-05-04
[5/🧹] Use `AuthParameters` in IWA.

kyle-rader-msft:
Addresses the auth params data clump in `IntegratedWindowsAuthentication`.

### DIFF
--- a/src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
+++ b/src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
@@ -16,7 +16,6 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
     /// </summary>
     public class IntegratedWindowsAuthentication : AuthFlowBase
     {
-        private const string NameValue = "iwa";
         private readonly IEnumerable<string> scopes;
         private readonly string preferredDomain;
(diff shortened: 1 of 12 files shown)


src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -17,6 +17,8 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
     public class IntegratedWindowsAuthentication : AuthFlowBase
     {
         private readonly IEnumerable<string> scopes;
+
+        // TODO: We don't currently warn if the account returned from IWA oesn't match the preferred domain.
```
shalinikhare27, line 21:
```suggestion
        // TODO: We don't currently warn if the account returned from IWA doesn't match the preferred domain.
```

Did you log as a warning?

kyle-rader-msft: Yes, right now the domain is unused in this auth flow, so if someone was trying to log in with their Torus account, or maybe some other account that IWA would succeed for, if the domain doesn't match the given domain that's likely something worth warning about. We can open a separate bug for this.

mvanchaa (approved):

Haard30 (approved):

PR #303, opened 2023-05-12
Upgrade Lasso to 2023.5.11.1

goagain:
The latest Lasso will not generate log files in sub-process for async telemetry mode.
It will mitigate the number of files in temp folder (which has a maximum at 65535 on DotNet 7.0 and lower).

https://learn.microsoft.com/en-us/dotnet/api/system.io.path.gettempfilename?view=net-7.0#:~:text=On%20.NET%207,any%20operating%20system.

### DIFF
--- a/src/AzureAuth/AzureAuth.csproj
+++ b/src/AzureAuth/AzureAuth.csproj
@@ -33,7 +33,7 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


reillysiemens (approved):

kyle-rader-msft (approved):
Thanks Rick!

PR #309, opened 2023-06-22
Update release workflow to release Linux binaries.

goagain:
In #211 we created a workflow to enable Linux release, but there was a bug in ADO side and we reverted it in #226.

Now the bug in ADO has been fixed, and we want to enable it again.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -148,6 +148,38 @@ jobs:
           name: azureauth-${{ github.event.inputs.version }}-${{ matrix.runtime }}
           path: azureauth-${{ github.event.inputs.version }}-${{ matrix.runtime }}
 
+  # Build and sign Linux binaries on Azure DevOps and publish them to GitHub and packages.microsoft.com.
+  linux_release:
+    runs-on: ubuntu-latest
+    needs: [validate]
+    env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -148,6 +148,38 @@ jobs:
           name: azureauth-${{ github.event.inputs.version }}-${{ matrix.runtime }}
           path: azureauth-${{ github.event.inputs.version }}-${{ matrix.runtime }}
 
+  # Build and sign Linux binaries on Azure DevOps and publish them to GitHub and packages.microsoft.com.
```
mvanchaa, line 151: @goagain on line 42, we build azureauth for linux which currently doesn't build or create any debian package. We can skip doing that because we are building the package on ADO pipeline. I made that change in the PR #211 (https://github.com/AzureAD/microsoft-authentication-cli/blob/ed3cd91207f9b4180d0cc995a54b32492eb9c291/.github/workflows/release.yml#L42C1-L42C1)

goagain: Removed linux build

reillysiemens (approved):

mvanchaa (approved):

PR #311, opened 2023-06-29
Allow pat and token subcommands to use custom tenant

reillysiemens:
Currently neither the `ado pat` nor `ado token` subcommands pass through a custom tenant (although the `ado token` subcommand had the flag for it). This PR addresses that issue for both subcommands.

### DIFF
--- a/src/AzureAuth/Commands/Ado/CommandPat.cs
+++ b/src/AzureAuth/Commands/Ado/CommandPat.cs
@@ -81,6 +81,9 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Ado
         [Option(CommandAad.PromptHintOption, CommandAad.PromptHintHelpText, CommandOptionType.SingleValue)]
         private string PromptHint { get; set; } = null;
 
+        [Option(CommandAad.TenantOption, Description = "The Azure Tenant ID to use for authentication. Defaults to Microsoft.")]
+        private string Tenant { get; set; } = AzureAuth.Ado.Constants.Tenant.Microsoft;
+
         [Option(CommandAad.ModeOption, CommandAad.AuthModeHelperText, CommandOptionType.MultipleValue)]
         private IEnumerable<AuthMode> AuthModes { get; set; } = new[] { AuthMode.Default };
 
@@ -201,7 +204,7 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Ado
         private TokenResult AccessToken(IPublicClientAuth publicClientAuth, CommandExecuteEventData eventData)
         {
             return publicClientAuth.Token(
-                AzureAuth.Ado.Constants.AdoParams,
+                AzureAuth.Ado.AuthParameters.AdoParameters(this.Tenant),
                 this.AuthModes,
                 this.Domain,
                 this.PromptHint,
--- a/src/AzureAuth/Ado/AuthParameters.cs
+++ b/src/AzureAuth/Ado/AuthParameters.cs
@@ -1,0 +1,20 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Authentication.AzureAuth.Ado
+{
+    /// <summary>Azure DevOps-specific authentication parameters.</summary>
+    public static class AuthParameters
+    {
+        /// <summary>Create Azure DevOps-specific authentication parameters.</summary>
+        /// <param name="tenant">The tenant. Defaults to Microsoft.</param>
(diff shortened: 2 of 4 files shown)


src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -200,8 +203,12 @@ private bool ValidOptions(ILogger logger)
 
         private TokenResult AccessToken(IPublicClientAuth publicClientAuth, CommandExecuteEventData eventData)
         {
+            var authParams = new AuthParameters(
+                AzureAuth.Ado.Constants.Client.VisualStudio,
+                this.Tenant,
+                new[] { AzureAuth.Ado.Constants.Scope.AzureDevOpsDefault });
             return publicClientAuth.Token(
-                AzureAuth.Ado.Constants.AdoParams,
```
mvanchaa, line 204: Does this mean we can get rid of this constant?

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -200,8 +203,12 @@ private bool ValidOptions(ILogger logger)
 
         private TokenResult AccessToken(IPublicClientAuth publicClientAuth, CommandExecuteEventData eventData)
         {
+            var authParams = new AuthParameters(
+                AzureAuth.Ado.Constants.Client.VisualStudio,
+                this.Tenant,
+                new[] { AzureAuth.Ado.Constants.Scope.AzureDevOpsDefault });
```
mvanchaa, line 209: Do we not need custom scope to pass through here?

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -200,8 +203,12 @@ private bool ValidOptions(ILogger logger)
 
         private TokenResult AccessToken(IPublicClientAuth publicClientAuth, CommandExecuteEventData eventData)
         {
+            var authParams = new AuthParameters(
```
mvanchaa, line 206: Totally unrelated to this PR, but related to the previous one. Sorry I just thought about it - we have two constants file one of which is introduced in the previous PR - https://github.com/AzureAD/microsoft-authentication-cli/blob/main/src/AzureAuth/Ado/Constants.cs  and this one https://github.com/AzureAD/microsoft-authentication-cli/blob/main/src/AdoPat/Constants.cs 

Do we need both the files or can we merge them?

mvanchaa (changes requested):

src/AzureAuth/Commands/Ado/CommandPat.cs
reillysiemens: Rather than get rid of it we might want to find a better way to use it. Perhaps a builder-style method on that allows overriding the tenant... 🤔

src/AzureAuth/Commands/Ado/CommandPat.cs
reillysiemens: Confusingly, that's a different scope. Azure AD scopes are not the same as PAT scopes for Azure DevOps. In this case it's the former and that's the only one we need in this case, so there's no reason to do pass through.

src/AzureAuth/Commands/Ado/CommandPat.cs
reillysiemens: I'm not sure it's prudent to merge them without merging the rest. The constants in `AzureAuth.Ado.Constants` are ones which are important for the CLI. The ones in `AdoPat.Constants` are ones which are important for the ADO PAT library (which is used internally by the CLI). There's a subtle, but important difference there, I think. To me, it makes sense to keep them separate for now.

goagain:
Will we also add `clear` option in ADO? I felt painful when I was testing then clearing cache on Linux.

goagain (dismissed):

src/AzureAuth/Commands/Ado/CommandPat.cs
mvanchaa: I like that idea :)

reillysiemens:
@goagain: Your suggestion of a `clear` subcommand (or flag) is a good one, but not within the scope of this PR, I think. It's not something that anyone has asked for just yet, so I don't think it's urgent when we can delete a file during testing. Still, it's something for us to consider down the road.

mvanchaa (approved):

Haard30 (dismissed):

goagain (dismissed):

Haard30 (dismissed):

goagain (approved):

Haard30 (approved):

PR #317, opened 2023-08-03
Handle SIGINT (Ctrl+C)

goagain:
AzureAuth may hang forever when receiving SIGINT signal. This PR will handle the signal and exit gracefully.

### DIFF
--- a/src/AzureAuth/Program.cs
+++ b/src/AzureAuth/Program.cs
@@ -97,6 +97,12 @@ namespace Microsoft.Authentication.AzureAuth
             services.AddSingleton<IPublicClientAuth, PublicClientAuth>();
             services.AddSingleton<ILogger>(logger);
 
+            Console.CancelKeyPress += (_, _) =>
+            {
+                logger.LogWarning("Received SIGINT (Ctrl+C), exiting...");
+                Environment.Exit(2);
+            };
+
             new Lasso(app, options, services).Execute(args);
         }
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- AzureAuth now can handle SIGINT(Ctrl+C) correctly and return 2.
 
 ## [0.8.2] - 2023-07-06
 ### Added


mvanchaa (dismissed):

kyle-rader-msft:
Ideally, catching the CTRL-C would allow us to abort the auth in progress, and finish executing, returning telemetry that it was cancelled.

src/AzureAuth/Program.cs
```diff
@@ -97,6 +97,12 @@ private static void Main(string[] args)
             services.AddSingleton<IPublicClientAuth, PublicClientAuth>();
             services.AddSingleton<ILogger>(logger);
 
+            Console.CancelKeyPress += (_, _) =>
```
shalinikhare27, line 100: I tried testing this via your branch and behavior was same in main / when checking out your branch.

I tried running `azureauth ado token` and abruptly pressed CTRL+C (version 0.8.2) and no message was printed. What am I missing?

goagain:
> Ideally, catching the CTRL-C would allow us to abort the auth in progress, and finish executing, returning telemetry that it was cancelled. (That may be a lot harder/more complex though).

It seems not possible or very hard to send telemetry data, because the handle will exit the process immediately and it is not in Lasso execution.

The scenario that user interrupts AzureAuth should be very rare. Maybe we don't need to collect such telemetry data?

src/AzureAuth/Program.cs
goagain: Maybe you interrupt AzureAuth too quick, and at that time logger hasn't been initialized. You can add option `--verbosity debug`, and press Ctrl+C after you see the first message.

src/AzureAuth/Program.cs
```diff
@@ -97,6 +97,12 @@ private static void Main(string[] args)
             services.AddSingleton<IPublicClientAuth, PublicClientAuth>();
             services.AddSingleton<ILogger>(logger);
 
+            Console.CancelKeyPress += (_, _) =>
+            {
+                logger.LogWarning("Received SIGINT (Ctrl+C), exiting...");
```
kyle-rader-msft, line 102: `logger?`?

goagain: `logger` should not be null because it is initialized in line 93
```
            var logger = loggerFactory.CreateLogger("AzureAuth");
```

And the handler is registered after that.

src/AzureAuth/Program.cs
```diff
@@ -97,6 +97,12 @@ private static void Main(string[] args)
             services.AddSingleton<IPublicClientAuth, PublicClientAuth>();
             services.AddSingleton<ILogger>(logger);
 
+            Console.CancelKeyPress += (_, _) =>
+            {
+                logger.LogWarning("Received SIGINT (Ctrl+C), exiting...");
+                Environment.Exit(-1);
```
kyle-rader-msft, line 103: I'm not familiar with returning a negative exit code. Is that a thing? I would have expected to see `1` rather than `-1`.

kyle-rader-msft: After doing a brief search around I don't think any systems expect a negative exit code and this would be similar to returning 256, so I think we want `1` here.

kyle-rader-msft (changes requested):

src/AzureAuth/Program.cs
goagain: I thought -1 was like 255, which was the exit code when I interrupt a program in some debuggers.😅

kyle-rader-msft: If you want 255, you should use 255, not -1, which would rely on consumers to wrap around.

From https://unix.stackexchange.com/questions/110348/how-do-i-get-the-list-of-exit-codes-and-or-return-codes-and-meaning-for-a-comm, 

it looks like in Bash at least, generally `130` indicates interrupted.

kyle-rader-msft (approved):

src/AzureAuth/Program.cs
goagain: Seems like I should use exit code 2. 130=128+2 where 128 means the process was force closed by external signals on Linux.

goagain: For example, in this post: https://stackoverflow.com/questions/66321338/exit-code-130-on-linux-and-2-on-windows-on-sigint

kyle-rader-msft (approved):

shalinikhare27 (approved):

Issue #178, opened 2022-12-15
Change WS-Trust endpoint warning message


reillysiemens:
The warning in this section of code introduced in #153 

https://github.com/AzureAD/microsoft-authentication-cli/blob/85a51c3c67e7d911a6a7471fcda26cf3b959dea3/src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs#L111-L114

could probably be more generic and refer to a non-Microsoft solution. How else could we phrase this?

kyle-rader-msft:
@reillysiemens we could change "Corp Net" to something more generic like "corporate AD backed" networks. Though we should make sure it's accurate.

PR #328, opened 2023-08-22
Downgrade IWA warning message

goagain:
Many customers report that the warning message "IWA only works on Corp Net, please turn on VPN." from IWA is very confusing. We can be clearer by downgrading it to debug level and changing it to 
```
IWA only works on corporate AD backed network, AzureAuth is trying to use other auth flows.
If you specify the IWA mode, please turn on VPN.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - AzureAuth now can handle SIGINT(Ctrl+C) correctly and return 2.
 
 ### Changed
-- Optimize the warning message from IWA auth flow when VPN is not connected.
+- Optimize the warning message from IWA auth flow when VPN is not connected, and downgrade it to debug level.
 
 ## [0.8.2] - 2023-07-06
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - AzureAuth now can handle SIGINT(Ctrl+C) correctly and return 2.
 
+### Changed
+- When attempting the IWA authentication flow without a VPN connection, the warning message "IWA only works on Corp Net, please turn on VPN." will be downgraded to info level.
+
 ## [0.8.2] - 2023-07-06
 ### Added
 - The `azureauth ado` subcommands now support a `--tenant` flag.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - AzureAuth now can handle SIGINT(Ctrl+C) correctly and return 2.
 
 ### Changed
-- When attempting the IWA authentication flow without a VPN connection, the warning message "IWA only works on Corp Net, please turn on VPN." will be downgraded to info level.
+- Optimize the warning message from IWA auth flow when VPN is not connected.
 
 ## [0.8.2] - 2023-07-06
 ### Added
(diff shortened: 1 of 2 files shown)


src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -72,7 +72,7 @@ protected override async Task<TokenResult> GetTokenInnerAsync()
             }
             catch (MsalClientException ex) when (ex.Message.Contains("WS-Trust endpoint not found"))
             {
-                this.logger.LogWarning($"IWA only works on Corp Net, please turn on VPN.");
+                this.logger.LogInformation($"IWA only works on Corp Net, please turn on VPN.");
```
Haard30, line 75: Related to #178, should we also consider changing the wording too?

goagain: I think we can be more specific, and provide a clear instruction

shalinikhare27 (approved):

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -72,7 +72,8 @@ protected override async Task<TokenResult> GetTokenInnerAsync()
             }
             catch (MsalClientException ex) when (ex.Message.Contains("WS-Trust endpoint not found"))
             {
-                this.logger.LogWarning($"IWA only works on Corp Net, please turn on VPN.");
+                this.logger.LogWarning($"IWA only works on corporate AD backed network, AzureAuth is trying to use other auth flows.");
+                this.logger.LogWarning($"If you specify the IWA mode, please turn on VPN.");
```
Haard30, line 76: Should we instead say, "Turn on VPN for IWA mode to succeed". The user might not have externally specified this mode and it could lead to confusion?

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -72,7 +72,8 @@ protected override async Task<TokenResult> GetTokenInnerAsync()
             }
             catch (MsalClientException ex) when (ex.Message.Contains("WS-Trust endpoint not found"))
             {
-                this.logger.LogWarning($"IWA only works on Corp Net, please turn on VPN.");
+                this.logger.LogWarning($"IWA only works on corporate AD backed network, AzureAuth is trying to use other auth flows.");
```
Haard30, line 75: This might not be always true if user has selected only IWA mode for their auth. We shouls also try to be consistent with using modes v/s auth flows. We can something similar to "...AzureAuth will use other modes if available"

kyle-rader-msft, line 75: I think the problem here is that this is logged as Warning, and it generally doesn't need to be a warning given, most users will either succeed here, or in web mode. My recommendation is to change the verbosity level to debug, so it's till present in the logs, but not shown on the CLI during normal usage. The text itself, could be left alone.

Haard30 (dismissed):

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - AzureAuth now can handle SIGINT(Ctrl+C) correctly and return 2.
 
+### Changed
+- Optimize the warning message from IWA auth flow when VPN is not connected.
```
Haard30, line 15: We should change this too as we have downgraded the warning to debug log

Haard30 (approved):

shalinikhare27 (approved):

goagain:
resolve #178

PR #331, opened 2023-08-29
Change the top-level help text to make sense

reillysiemens:
Saying CLI interface is like saying ATM machine. The `I` is for `interface`. This just spells it out by using all the words instead of the abbreviation.

### DIFF
--- a/src/AzureAuth/Commands/CommandAzureAuth.cs
+++ b/src/AzureAuth/Commands/CommandAzureAuth.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Authentication.AzureAuth.Commands
     /// <summary>
     /// The command main class parses commands and dispatches to the corresponding methods.
     /// </summary>
-    [Command(Name = "azureauth", Description = "A CLI interface to MSAL (Microsoft Authentication Library).")]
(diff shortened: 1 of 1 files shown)


shalinikhare27 (approved):

goagain (approved):

src/AzureAuth/Commands/CommandAzureAuth.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.Authentication.AzureAuth.Commands
     /// <summary>
     /// The command main class parses commands and dispatches to the corresponding methods.
     /// </summary>
-    [Command(Name = "azureauth", Description = "A CLI interface to MSAL (Microsoft Authentication Library).")]
+    [Command(Name = "azureauth", Description = "A command line interface to MSAL (Microsoft Authentication Library).")]
```
Haard30, line 15: Should we also cover the ability to generate PATs in the text?

reillysiemens: That's covered by one of the subcommands. A whole overhaul of our help text might be worthwhile, but I think that should be a separate PR. 😆

Haard30 (approved):

PR #333, opened 2023-09-01
Fix the release issue `Pattern 'azureauth-0.8.3-linux-x64.deb' does not match any files`

goagain:
## Background
We got errors from our latest release
**🤔 Pattern 'azureauth-0.8.3-linux-x64.deb' does not match any files.
🤔 Pattern 'azureauth-0.8.3-linux-arm64.deb' does not match any files.**

## Issue
The artifact name doesn't mean the file name. In the last step, the artifact was downloaded with the raw name `azureauth_0.8.3-1_amd64.deb`, that's why the file cannot be matched.

## Solution
Rename the deb package before uploading to artifact.

## Validation
https://github.com/goagain/microsoft-authentication-cli/actions/runs/6043610356
https://github.com/goagain/microsoft-authentication-cli/releases/tag/0.8.3

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -172,16 +172,22 @@ jobs:
         ADO_AZUREAUTH_LINUX_PIPELINE_ID: ${{ secrets.ADO_AZUREAUTH_LINUX_PIPELINE_ID }}
         ADO_AZUREAUTH_LINUX_STAGE_ID: ${{ vars.ADO_AZUREAUTH_LINUX_STAGE_ID }}
         VERSION: ${{ github.event.inputs.version }}
-    - name: Upload linux-x64 artifact
+
+    - name: Rename linux-x64 artifact
+      run: |
+        mv ${{ env.ADO_LINUX_ARTIFACT_DOWNLOAD_PATH }}/${{ env.ADO_LINUX_ARTIFACT_NAME }}/azureauth_${{ github.event.inputs.version }}-${{ env.DEBIAN_REVISION }}_amd64.deb \
+        azureauth-${{ github.event.inputs.version }}-linux-x64.deb
+    - name: Rename linux-arm64 artifact
+      run: |
+        mv ${{ env.ADO_LINUX_ARTIFACT_DOWNLOAD_PATH }}/${{ env.ADO_LINUX_ARTIFACT_NAME }}/azureauth_${{ github.event.inputs.version }}-${{ env.DEBIAN_REVISION }}_arm64.deb \
+        azureauth-${{ github.event.inputs.version }}-linux-arm64.deb
+    - name: Upload linux artifact
       uses: actions/upload-artifact@v3
       with:
-        name: azureauth-${{ github.event.inputs.version }}-linux-x64.deb
-        path: ${{ env.ADO_LINUX_ARTIFACT_DOWNLOAD_PATH }}/${{ env.ADO_LINUX_ARTIFACT_NAME }}/azureauth_${{ github.event.inputs.version }}-${{ env.DEBIAN_REVISION }}_amd64.deb
-    - name: Upload linux-arm64 artifact
-      uses: actions/upload-artifact@v3
-      with:
-        name: azureauth-${{ github.event.inputs.version }}-linux-arm64.deb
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -249,11 +253,7 @@ jobs:
     - name: Download linux-x64 artifact
       uses: actions/download-artifact@v3
       with:
-        name: azureauth-${{ github.event.inputs.version }}-linux-x64.deb
-    - name: Download linux-arm64 artifact
-      uses: actions/download-artifact@v3
-      with:
-        name: azureauth-${{ github.event.inputs.version }}-linux-arm64.deb
+        name: azureauth-linux
```
reillysiemens, line 256: Huh, so this allows both `.deb` artifacts to be referred to by a single name (`azureauth-linux`) and downloads both of them into place?

.github/workflows/release.yml
```diff
@@ -172,16 +172,22 @@ jobs:
...
-    - name: Upload linux-arm64 artifact
+
+    - name: Rename linux-x64 artifact
+      run: |
+        mv ${{ env.ADO_LINUX_ARTIFACT_DOWNLOAD_PATH }}/${{ env.ADO_LINUX_ARTIFACT_NAME }}/azureauth_${{ github.event.inputs.version }}-${{ env.DEBIAN_REVISION }}_amd64.deb \
+        azureauth-${{ github.event.inputs.version }}-linux-x64.deb
+    - name: Rename linux-arm64 artifact
+      run: |
+        mv ${{ env.ADO_LINUX_ARTIFACT_DOWNLOAD_PATH }}/${{ env.ADO_LINUX_ARTIFACT_NAME }}/azureauth_${{ github.event.inputs.version }}-${{ env.DEBIAN_REVISION }}_arm64.deb \
+        azureauth-${{ github.event.inputs.version }}-linux-arm64.deb
```
reillysiemens, line 183: We do the other actions involving these two artifacts in a single step. Can we do the same here as well?

.github/workflows/release.yml
```diff
@@ -172,16 +172,22 @@ jobs:
...
-        path: ${{ env.ADO_LINUX_ARTIFACT_DOWNLOAD_PATH }}/${{ env.ADO_LINUX_ARTIFACT_NAME }}/azureauth_${{ github.event.inputs.version }}-${{ env.DEBIAN_REVISION }}_amd64.deb
-    - name: Upload linux-arm64 artifact
+
+    - name: Rename linux-x64 artifact
+      run: |
+        mv ${{ env.ADO_LINUX_ARTIFACT_DOWNLOAD_PATH }}/${{ env.ADO_LINUX_ARTIFACT_NAME }}/azureauth_${{ github.event.inputs.version }}-${{ env.DEBIAN_REVISION }}_amd64.deb \
+        azureauth-${{ github.event.inputs.version }}-linux-x64.deb
+    - name: Rename linux-arm64 artifact
+      run: |
+        mv ${{ env.ADO_LINUX_ARTIFACT_DOWNLOAD_PATH }}/${{ env.ADO_LINUX_ARTIFACT_NAME }}/azureauth_${{ github.event.inputs.version }}-${{ env.DEBIAN_REVISION }}_arm64.deb \
```
reillysiemens, line 182: Names like this are pretty long and make it easy to miss what's being done. Especially if more than one line is used. How do you feel about doing
```yaml
env:
    DEB_SOURCE: ${{ env.ADO_LINUX_ARTIFACT_DOWNLOAD_PATH }}/${{ env.ADO_LINUX_ARTIFACT_NAME }}/azureauth_${{ github.event.inputs.version }}-${{ env.DEBIAN_REVISION }}_arm64.deb
    DEB_TARGET: azureauth-${{ github.event.inputs.version }}-linux-arm64.deb
run: |
    mv ${{ env.DEB_SOURCE }} ${{ env.DEB_TARGET}}
```
or something similar?

.github/workflows/release.yml
goagain: Right. Artifacts are similar as a zip file. When downloading an artifact, it actually downloads then extracts the inside files.

reillysiemens (approved):

shalinikhare27 (approved):

Issue #121, opened 2022-06-21
Provide an uninstall utility


reillysiemens:
This could either be done with a flag to the install script or with a separate set of platform-specific uninstall scripts. Either way, though we'd be sad to see them go, users might appreciate this.

PR #340, opened 2023-09-28
AzureAuth Windows uninstall script

Emilio0404:
Added a script to uninstall AzureAuth easily. The script handles the removal of the default AzureAuth installation folder, and updates the PATH to not contain any references to AzureAuth.

Right now, it was tested manually, but integration tests are coming soon.

From issue: https://github.com/AzureAD/microsoft-authentication-cli/issues/121


shalinikhare27:
@Emilio0404 - We can just edit the title as `AzureAuth uninstall script` :)

install/uninstall.ps1
```diff
@@ -0,0 +1,95 @@
...
+
+    if ($deleted) {
+        Write-Output "Uninstalled AzureAuth!"
+    } else {
+        Write-Output "There were no AzureAuth installations to be deleted."
+    }
+}
+
+function Get-UninstallDirectory {
+    $directory = if (![string]::IsNullOrEmpty($Env:AZUREAUTH_INSTALL_DIRECTORY)) {
```
mvanchaa, line 22: What if the env `AZUREAUTH_INSTALL_DIRECTORY` refers to the `%localappdata%/Programs/AzureAuth`?

install/uninstall.ps1
```diff
@@ -0,0 +1,95 @@
...
+    param ([string]$uninstallDirectory)
+
+    $registryPath = 'Registry::HKEY_CURRENT_USER\Environment'
+    $currentPath = (Get-ItemProperty -Path $registryPath -Name PATH -ErrorAction SilentlyContinue).Path
+
+    # We try to find any other AzureAuth installations that are not in default location.
+    # Installations in custom locations that are not listed in PATH cannot be found.
+    $azureauthsInPath = (Get-Command -Name azureauth -ErrorAction SilentlyContinue -CommandType Application -All).Source
+    $otherDirectories = [System.Collections.ArrayList]@()
+    ForEach($az in $azureauthsInPath) {
```
mvanchaa, line 50: Are we removing custom location installed azureauth paths from PATH here?

install/uninstall.ps1
```diff
@@ -0,0 +1,95 @@
...
+        if (!$az.Contains($uninstallDirectory)) {
```
mvanchaa, line 51: Are we removing custom location installed azureauth paths from PATH here?

Emilio0404: I realized that we can only remove the custom locations that are listed in the PATH, so I removed those. If there are custom locations NOT listed in the PATH, there is no way to find those installations. I'll update the comments/messages to be more clear.

install/uninstall.ps1
Emilio0404: Good point, I'll make a check for that.

install/uninstall.ps1
```diff
@@ -0,0 +1,95 @@
...
+    if ($deleted) {
+        Write-Output "Uninstalled AzureAuth!"
+    } else {
+        Write-Output "There were no AzureAuth installations to be deleted."
+    }
+}
+
+function Get-UninstallDirectory {
+    $directory = if (![string]::IsNullOrEmpty($Env:AZUREAUTH_INSTALL_DIRECTORY)) {
+        Write-Error "Uninstalling from custom location is not supported"
```
shalinikhare27, line 23: [super nit]
```suggestion
        Write-Error "Uninstalling from custom location is not supported."
```

install/uninstall.ps1
shalinikhare27: Definitely can edge case we handle. However, our [doc ](https://github.com/AzureAD/microsoft-authentication-cli/blob/c1296b3d9777f8ace89e3621de0d433275252d98/README.md#windows)clearly specifies that we install at `%localappdata%/Programs/AzureAuth` by default unless the env var is set. So if someone sets that env var to default location, that would be weird, but worth handling. :)

install/uninstall.ps1
```diff
@@ -0,0 +1,95 @@
...
+    }
+}
+
+function Remove-FromPath {
+    param ([string]$uninstallDirectory)
+
+    $registryPath = 'Registry::HKEY_CURRENT_USER\Environment'
+    $currentPath = (Get-ItemProperty -Path $registryPath -Name PATH -ErrorAction SilentlyContinue).Path
+
+    # We try to find any other AzureAuth installations that are not in default location.
```
shalinikhare27, line 46: So if there is azureauth in custom location, we do not support uninstalling it. However, we do support uninstalling from custom location if it is listed in PATH?

install/uninstall.ps1
```diff
@@ -0,0 +1,95 @@
...
+            elseif (!$path.Equals("")) {
+                Write-Verbose "Removing '${path}' from `$env:PATH"
+            }
+        }
+        $updatedPath = ($pathArr -join ";") + ";"
+    }
+
+    Set-ItemProperty -Path $registryPath -Name PATH -Value $updatedPath
+    Send-SettingChange
+    Write-Verbose "Custom installations that were not in `$env:Path could not be deleted."
```
shalinikhare27, line 78:
```suggestion
    Write-Verbose "Custom installations that were not in $env:Path could not be deleted."
```

install/uninstall.ps1
shalinikhare27, line 46: [nit]
```suggestion
    # We try to find and uninstall any other AzureAuth installations that are not in default location.
```

install/uninstall.ps1
```diff
@@ -0,0 +1,95 @@
...
+}
+
+function Remove-FromPath {
+    param ([string]$uninstallDirectory)
+
+    $registryPath = 'Registry::HKEY_CURRENT_USER\Environment'
+    $currentPath = (Get-ItemProperty -Path $registryPath -Name PATH -ErrorAction SilentlyContinue).Path
+
+    # We try to find any other AzureAuth installations that are not in default location.
+    # Installations in custom locations that are not listed in PATH cannot be found.
```
shalinikhare27, line 47:
```suggestion
    # Installations in custom locations that are not listed in PATH cannot be found and uninstalled.
```

install/uninstall.ps1
mvanchaa: True that!

install/uninstall.ps1
mvanchaa: Just had a conversation with Emilio regarding the same :)
We should either modify the error message on line 23 or just uninstall default azureauth installation and it's location from the PATH.

I would suggest the first approach by mentioning something like `Attempting to remove AzureAuth from all the installation locations found`

Emilio0404: Yeah. I will change "not support uninstalling from custom location" to "cannot uninstall custom location if not listed in PATH".

This is because if there happens to be more than one AzureAuth installed, for example one in default and another one in custom location (both listed in PATH), if we only delete the AzureAuth in the default location, then the AzureAuth in a custom location is used. To the user, it would appear as if it wasn't uninstalled.

mvanchaa: 100% agree with what you said about multiple installations. Awesome you handled it! :D

Haard30: Thanks for the explanation, Emilio. I have some more questions - Wouldn't removing from PATH and not removing from custom location leave the machine in weird state. For example, there is some `AzureAuth` installation folder lying around but now we do not know its location because we removed that from PATH. Would this take away our ability in future to support custom location uninstallation?

mvanchaa: He is also deleting the folder before removing from the PATH, correct me if I'm wrong, Emilio.

install/uninstall.ps1
```diff
@@ -0,0 +1,95 @@
...
+    $azureauthsInPath = (Get-Command -Name azureauth -ErrorAction SilentlyContinue -CommandType Application -All).Source
+    $otherDirectories = [System.Collections.ArrayList]@()
+    ForEach($az in $azureauthsInPath) {
+        if (!$az.Contains($uninstallDirectory)) {
+            $additionalDirectory = (Get-Item $az).Directory.FullName
+            Write-Verbose "Additional installation found in ${additionalDirectory}"
+            $_ = $otherDirectories.Add($additionalDirectory)
+            Remove-InstallationFolder ($additionalDirectory)
```
Haard30, line 55: Should we re-design this so `Remove-FromPath` only updates the path and does not perform any folder deletion? We can first gather these custom installations and call `Remove-InstalllationFolder`?

reillysiemens, line 55: From a separation of concerns perspective, this function becomes rather more complicated by doing this. At face value, by looking at the name `Remove-FromPath`, I expect this to only manipulate `$PATH`, not to _also_ manipulate the filesystem.

**If** we did go ahead with the notion of uninstalling from custom locations, I would consider reorganizing the logic like so

1. Discover **all** locations where AzureAuth might be installed (always include the default directory).
2. Then remove all directories in a single place in the script.
3. Then remove all related `$PATH` entries because they can't point to anything anymore.

The output of step 1 can be fed into step 2, which can then be fed into step 3.

I think I _now_ understand why `$deleted` is a global, but I don't think it would need to be if the steps were pipelined as above.

reillysiemens (changes requested):
@Emilio0404, this is a great first AzureAuth PR. I have a couple of thoughts/questions for you, but overall I think you're on the right track. I think the biggest thing to sort out will be whether we can reasonably support uninstalling from custom locations.

install/uninstall.ps1
Emilio0404: I will go ahead with this redesign since now two people have mentioned it. I will ping request re-review once I'm done

Emilio0404: Will do, thank you!

install/uninstall.ps1
Emilio0404: Added the check for it.

install/uninstall.ps1
```diff
@@ -0,0 +1,95 @@
...
+    } 
+    return ([System.IO.Path]::Combine($Env:LOCALAPPDATA, "Programs", "AzureAuth"))
+}
+
+function Remove-InstallationFolder {
+    param ([string]$directory)
+
+    if (Test-Path -Path $directory) {
+        Write-Verbose "Removing installations at '${directory}'"
+        Remove-Item -Force -Recurse $directory
```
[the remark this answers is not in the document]
Emilio0404: The files get locked so they don't get deleted. I will reuse the code in the install script to close running instances.

install/uninstall.ps1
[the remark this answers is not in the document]
Emilio0404: It was a leftover from previous drafts. Got rid of it.

install/uninstall.ps1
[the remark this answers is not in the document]
Emilio0404: I decided to go with `Write-Warning`, since it seems better to not stop the script. Thank you for letting me know about it :)

install/uninstall.ps1
Emilio0404: Now we only warn about custom installations, we won't delete them.

install/uninstall.ps1
```diff
@@ -0,0 +1,125 @@
...
+[CmdletBinding()]
+
+# Halt script execution at the first failed command.
+$script:ErrorActionPreference='Stop'
+
+$azureauthDefaultLocation = ([System.IO.Path]::Combine($Env:LOCALAPPDATA, "Programs", "AzureAuth"))
+
+if (![string]::IsNullOrEmpty($Env:AZUREAUTH_INSTALL_DIRECTORY) `
+        -And $Env:AZUREAUTH_INSTALL_DIRECTORY -ne $azureauthDefaultLocation) {
+    Write-Warning "Ignoring AZUREAUTH_INSTALL_DIRECTORY environment variable. Custom uninstallation locations are not supported."
```
shalinikhare27, line 11:
```suggestion
    Write-Warning "Ignoring AZUREAUTH_INSTALL_DIRECTORY environment variable. Uninstallation for custom locations is not supported."
```

install/uninstall.ps1
```diff
@@ -0,0 +1,125 @@
...
+            }
+        }
+        $updatedPath = ($pathArr -join ";") + ";"
+    }
+
+    Set-ItemProperty -Path $registryPath -Name PATH -Value $updatedPath
+    Send-SettingChange
+}
+
+function Close-AzureauthInstances {
```
shalinikhare27, line 95: Umm, I am curious what others think. Should we rename this method to `Kill-AzureAuthInstances`? We use `taskkill`, so that might be more appropriate here.

install/uninstall.ps1
```diff
@@ -0,0 +1,125 @@
...
+    Close-AzureauthInstances
+
+    $locations = Get-AzureauthsInPath
+    Remove-AzureauthDirectories $locations
+    Remove-FromPath $locations
+
+    Write-Output "Uninstalled AzureAuth!"
+}
+
+function Get-AzureauthsInPath {
```
shalinikhare27, line 24: We can consistently update AzureAuth in pascal casing.
```suggestion
function Get-AzureAuthsInPath {
```

install/uninstall.ps1
```diff
@@ -0,0 +1,125 @@
...
+        -And $Env:AZUREAUTH_INSTALL_DIRECTORY -ne $azureauthDefaultLocation) {
+    Write-Warning "Ignoring AZUREAUTH_INSTALL_DIRECTORY environment variable. Custom uninstallation locations are not supported."
+}
+
+function Uninstall {
+    Close-AzureauthInstances
+
+    $locations = Get-AzureauthsInPath
+    Remove-AzureauthDirectories $locations
+    Remove-FromPath $locations
```
shalinikhare27, line 19: We can be consistent here and rename to `Remove-AzureAuthFromPath`. Or because this is the `AzureAuth` uninstallation script, we can rename your methods to be the following:

`Kill-Process`
`Get-InPath`
`Remove-Directories`
`Remove-FromPath`

I think it will be less overloaded.

install/uninstall.ps1
```diff
@@ -0,0 +1,125 @@
...
+
+function Get-AzureauthsInPath {
+    $allLocations = [System.Collections.ArrayList]@()
+    $azureauthsInPath = (Get-Command -Name azureauth -ErrorAction SilentlyContinue -CommandType Application -All).Source
+
+    # Uninstallation is only supported from the default AzureAuth location.
+    # We warn the user of any custom locations that are listed in the PATH.
+    # Installations in custom locations that are not listed in PATH cannot
+    # be found and uninstalled.
+    # We still...
```
shalinikhare27, line 32: Is this unintentionally added?

install/uninstall.ps1
```diff
@@ -0,0 +1,125 @@
...
+    ForEach($az in $azureauthsInPath) {
+        $tempDirectory = (Get-Item $az).Directory.FullName
+        if (!$az.Contains($azureauthDefaultLocation)) {
+            Write-Warning "Additional installation found in ${tempDirectory}"
```
shalinikhare27, line 36: Should we expect users to take any action in this case? If so, we should print that. Additionally, I see you have added a comment `Installations in custom locations that are not listed in PATH cannot be found and uninstalled.` I think we should explicitly warn the user with this comment if there are no actions they need to take.

install/uninstall.ps1
```diff
@@ -0,0 +1,125 @@
...
+        } else {
+            # We add the PATH location to the list and not the default location to later
+            # remove it from the PATH (default location is the whole parent folder).
```
shalinikhare27, line 39: I think we can consider removing `and not the default location`. We are basically adding all the azureauths in PATH to this list to later remove it from PATH. It's pretty straightforward to get from what this method intends to do. 

`We add the PATH location to the list to later remove it from the PATH. Note, the default location is the parent folder.` ?

install/uninstall.ps1
```diff
@@ -0,0 +1,125 @@
...
+}
+
+function Remove-AzureauthDirectories {
+    param([System.Collections.ArrayList]$locationsToDelete)
+
+    ForEach ($location in $locationsToDelete) {
+        Remove-InstallationFolder $location
+    }
+    
+    # We also delte AzureAuth parent folder
```
shalinikhare27, line 54:
```suggestion
    # We also delete AzureAuth parent folder.
```

install/uninstall.ps1
```diff
@@ -0,0 +1,125 @@
...
+}
+
+function Remove-InstallationFolder {
+    param ([string]$directory)
+
+    if (Test-Path -Path $directory) {
+        Write-Verbose "Removing installations at '${directory}'"
+        Remove-Item -Force -Recurse $directory
+    } else {
+        Write-Verbose "There were no installations found at '${directory}'"
```
shalinikhare27, line 65: I am wondering if we will ever enter this `else` statement. We created a list of azureauth directories that exist already. Am I missing an edge case?

shalinikhare27 (changes requested):

install/uninstall.ps1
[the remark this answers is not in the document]
reillysiemens: Yeah, apologies, I asked a rhetorical question, but I could have just told you that it won't work. 😅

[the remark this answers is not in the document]
Emilio0404: No worries! It was a good exercise for me :)

install/uninstall.ps1
Emilio0404: User could manually delete the AzureAuth folder and not from the PATH. Though I could change the comment to "There were no installations found at '${directory}', removing reference from PATH"

install/uninstall.ps1
reillysiemens: Kill does seem like a more appropriate verb. This is not a polite request. The `/f` flag is being used.

install/uninstall.ps1
reillysiemens: I think @Emilio0404's correct about this edge case. Just because we discover something on the user's `PATH` doesn't mean that it's actually backed by anything on disk still.

I would be wary of crossing purposes in your output though. Saying that you're removing a reference from `PATH` in the `Remove-InstallationFolder` function assumes knowledge from another function or how the whole pipeline is put together. If you're truly making these functions individual components then this function shouldn't be responsible for that message.

Logging is also part of your API, not just the return value.

install/uninstall.ps1
```diff
@@ -0,0 +1,125 @@
...
+
+    $locations = Get-AzureauthsInPath
+    Remove-AzureauthDirectories $locations
+    Remove-FromPath $locations
+
+    Write-Output "Uninstalled AzureAuth!"
+}
+
+function Get-AzureauthsInPath {
+    $allLocations = [System.Collections.ArrayList]@()
```
reillysiemens, line 25: Unless I'm mistaken this doesn't get used for **all** locations, only for those which are subdirectories of the default installation directory. Other directories are iterated over, but ultimately ignored.

Consider renaming this variable accordingly.

reillysiemens (changes requested):
@Emilio0404: I appreciate your willingness to keep iterating on this. It seems like you're still working through some of this, so feel free to consider a draft PR if you're not ready for more review.

I've left some more granular, but minor feedback after your last changes. At a high level I want to suggest that maybe you could just delete the _entire_ default installation location instead of individually removing subdirectories thereof.

Basically, this:
1. Discover all paths where AzureAuth might be.
2. Just delete the entire `%LOCALAPPDATA%\Programs\AzureAuth` directory.
3. Clean up paths you deleted from `%PATH%` (these will be subdirectories of the above).
4. Output that you've uninstalled AzureAuth.
5. As a courtesy, output non-overlapping paths from step 1 (things you can't actually safely delete).

I think this might simplify some of your current logic.

@mvanchaa, @shalinikhare27, @Haard30: Would love your feedback on the above steps. ☝️

install/uninstall.ps1
Emilio0404: Changed the naming

install/uninstall.ps1
Emilio0404: Changed to kill

install/uninstall.ps1
```diff
@@ -0,0 +1,125 @@
...
+    $azureauthsInPath = (Get-Command -Name azureauth -ErrorAction SilentlyContinue -CommandType Application -All).Source
+
+    # Uninstallation is only supported from the default AzureAuth location.
+    # We warn the user of any custom locations that are listed in the PATH.
+    # Installations in custom locations that are not listed in PATH cannot
+    # be found and uninstalled.
+    # We still...
+    ForEach($az in $azureauthsInPath) {
+        $tempDirectory = (Get-Item $az).Directory.FullName
```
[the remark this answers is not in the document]
Emilio0404: Changed the name to $azureauthParent

install/uninstall.ps1
Emilio0404: We will just be warning of custom locations and not deleting them.

install/uninstall.ps1
Emilio0404: Removed unintended the comment.

install/uninstall.ps1
Emilio0404: Removed the comment

install/uninstall.ps1
Emilio0404: Left the comment as is.

install/uninstall.ps1
Emilio0404: Remove-FromPath doesn't do folder deletions anymore.

install/uninstall.ps1
```diff
@@ -0,0 +1,125 @@
...
+
+# Halt script execution at the first failed command.
+$script:ErrorActionPreference='Stop'
+
+$azureauthDefaultLocation = ([System.IO.Path]::Combine($Env:LOCALAPPDATA, "Programs", "AzureAuth"))
+
+if (![string]::IsNullOrEmpty($Env:AZUREAUTH_INSTALL_DIRECTORY) `
+        -And $Env:AZUREAUTH_INSTALL_DIRECTORY -ne $azureauthDefaultLocation) {
+    Write-Warning "Ignoring AZUREAUTH_INSTALL_DIRECTORY environment variable. Custom uninstallation locations are not supported."
+}
```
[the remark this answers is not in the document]
Emilio0404: Done

install/uninstall.ps1
Emilio0404: I added the message at the end that @reillysiemens suggested.

install/uninstall.ps1
Emilio0404: Changed the name.

mvanchaa (approved):

shalinikhare27 (approved):

reillysiemens (approved):

Haard30 (dismissed):

reillysiemens (approved):

shalinikhare27 (approved):

mvanchaa (approved):

PR #353, opened 2023-11-07
Just one more YAML fix, one more...

reillysiemens:
This makes the `variables` section compliant with the suggested documentation for migrating these pipelines.


Haard30 (approved):

goagain (approved):
😅

PR #356, opened 2023-11-15
Change target framework to .net 8.0

goagain:
Dotnet 8.0 was released recently. It is the latest LTS version of dotnet. We should change the framework to .net 8.0 and see whether it can bring performance and other benefits to us.

### DIFF
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -16,10 +16,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - name: Setup .NET 6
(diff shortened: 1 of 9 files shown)


goagain:
dotnet 8.0 has many breaking changes especially for trimming.
`CommandLineUtils`, an upstream of Lasso, is currently not compatible with dotnet 8.0.

kyle-rader-msft:
@mijpeterson @varshini-elayappen It would be nice to have azureauth on dotnet 8 :)

mijpeterson:
Closing as this PR is stale and we've already updated to .NET 8.

PR #357, opened 2023-11-15
Remove ESRP NuGet feed

goagain:
Office is a downstream of ESRP stream, which means we can always download `Microsoft.EsrpClient` from existing Office feed. By removing the feed and redirecting to Office, we no longer need to maintain 2 separate NuGet token. `ADO_TOKEN` will be enough.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -104,10 +104,10 @@ jobs:
       - name: Download ESRPClient.exe
         env:
           ESRP_VERSION: ${{ secrets.ESRP_VERSION }}
-          NUGET_CREDENTIALS: ${{ secrets.NUGET_CREDENTIALS }}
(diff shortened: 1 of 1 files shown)


shalinikhare27:
The only source of truth we have that this works is our Build pipeline in ADO indicating it can download ESRP stuff from office feed? We can't test in Github until we actually try to make a release?

goagain:
> The only source of truth we have that this works is our Build pipeline in ADO indicating it can download ESRP stuff from office feed? We can't test in Github until we actually try to make a release?

We can run release pipeline on this branch to validate. Remember don't hit the button.
image

shalinikhare27:
Well, that will be great validation to do. Make sure we let everyone in the team know to avoid hitting the button.

mvanchaa (approved):

reillysiemens (approved):
Great find here, @goagain!

Haard30 (approved):

PR #364, opened 2024-01-05
Reclassify CG pipeline as nonproduction

reillysiemens:
While important, it doesn't appear as though this pipeline requires a `production` classification and can instead be classified as `nonproduction`.


goagain (approved):

shalinikhare27 (approved):

rewrlution (commented):
LGTM.

PR #366, opened 2024-02-15
Add "ado pat scopes" sub-command to print available scopes

kyle-rader-msft:
# azureauth ado pat scopes
## Motivation
It's hard to remember where or find the actual list of scopes in the ADO documentation. After 5 searches this morning, I couldn't find it online and had to come here to the source to figure out what the actual scope was to request while using the `ado pat` command.

This adds a new subcommand onto the existing `ado pat` command path to list the available scopes and the link to the documentation making it much easier to use the command.

## Usage
```
C:\r\microsoft-authentication-cli> .\dist\azureauth.exe ado pat -h                                                                      
1.0.0.0
Create and cache Azure Devops Personal Access Tokens (PATs) using encrypted local storage.

Usage: azureauth ado pat [command] [options]

Options:
  --organization  The name of the Azure DevOps organization.
  --display-name  The PAT name.
  --scope         A token scope for accessing Azure DevOps resources. Repeated invocations allowed.
  --output        How PAT information is displayed. [default: token]
                  [possible values: none, status, token, base64, header, headervalue, json]
  --prompt-hint   A prompt hint to contextualize prompts and identify uses in telemetry, when captured.
  --tenant        The Azure Tenant ID to use for authentication. Defaults to Microsoft.
  --mode          Authentication mode. Repeated invocations allowed.
                  [default: iwa (Integrated Windows Auth), then broker, then web]
                  [possible values: all, iwa, broker, web, devicecode]
  --domain        Preferred domain for filtering cached accounts.
                  Skips launching an account picker if only one cached account matches the preferred domain.
                  [default: microsoft.com]
  --timeout       The number of minutes before authentication times out.
                  [default: 15 minutes]
  -?|-h|--help    Show help information.
  --version       Show version information.
  --debug         Turn off telemetry reporting.
  --verbosity     Configure the minimum log level to show.
                  Allowed values are: trace, debug, info, information, warn, warning, error, critical, fatal, none, off.

Commands:
  scopes          List the valid ado pat scopes

Run 'pat [command] -?|-h|--help' for more information about a command.
```

Note the new Commands section at the bottom 👆.

```
.\dist\azureauth.exe ado pat scopes                                                                  02/15/24 14:51:55 PMvso.agentpools
vso.agentpools_manage
vso.analytics
vso.auditlog
vso.build
vso.build_execute
vso.code
vso.code_full
vso.code_manage
vso.code_status
vso.code_write
vso.dashboards
vso.dashboards_manage
vso.drop
vso.drop_manage
vso.drop_write
vso.entitlements
vso.environment_manage
vso.extension
vso.extension_manage
vso.extension.data
vso.extension.data_write
vso.gallery
vso.gallery_acquire
vso.gallery_manage
vso.gallery_publish
vso.graph
vso.graph_manage
vso.identity
vso.identity_manage
vso.loadtest
vso.loadtest_write
vso.machinegroup_manage
vso.memberentitlementmanagement
vso.memberentitlementmanagement_write
vso.notification
vso.notification_diagnostics
vso.notification_manage
vso.notification_write
vso.packaging
vso.packaging_manage
vso.packaging_write
vso.profile
vso.profile_write
vso.project
vso.project_manage
vso.project_write
vso.release
vso.release_execute
vso.release_manage
vso.security_manage
vso.serviceendpoint
vso.serviceendpoint_manage
vso.serviceendpoint_query
vso.settings
vso.settings_write
vso.symbols
vso.symbols_manage
vso.symbols_write
vso.taskgroups_manage
vso.taskgroups_read
vso.taskgroups_write
vso.test
vso.test_write
vso.tokenadministration
vso.tokens
vso.variablegroups_manage
vso.variablegroups_read
vso.variablegroups_write
vso.wiki
vso.wiki_write
vso.work
vso.work_full
vso.work_write
```

### DIFF
--- a/src/AzureAuth/Commands/Ado/CommandPat.cs
+++ b/src/AzureAuth/Commands/Ado/CommandPat.cs
@@ -25,6 +25,7 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Ado
     /// An ADO Command for creating or fetching, and returning Azure Devops PATs.
     /// </summary>
     [Command("pat", Description = "Create and cache Azure Devops Personal Access Tokens (PATs) using encrypted local storage.")]
+    [Subcommand(typeof(Pat.CommandScopes))]
     public class CommandPat
     {
         private const string OrganizationOption = "--organization";
--- a/src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
+++ b/src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
@@ -10,7 +10,8 @@ using System.Linq;
 namespace Microsoft.Authentication.AzureAuth.Commands.Ado.Pat
 {
     /// <summary>
-    /// Command to print the list of avaialable scopes
+    /// Command to print the list of available scopes
+
     /// </summary>
     [Command("scopes", Description = "List the valid ado pat scopes")]
     public class CommandScopes
--- a/src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
+++ b/src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
@@ -4,6 +4,7 @@ using McMaster.Extensions.CommandLineUtils;
 using Microsoft.Authentication.AdoPat;
 using Microsoft.Extensions.Logging;
 
+using System;
 using System.Linq;
 
 namespace Microsoft.Authentication.AzureAuth.Commands.Ado.Pat
@@ -33,17 +34,12 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Ado.Pat
         /// <returns></returns>
         public int OnExecute()
         {
-            // Print at the top and bottom because it's a long list.
-            logger.LogInformation(UrlMessage + "\n");
-
             var scopes = Scopes.ValidScopes.ToList();
             scopes.Sort();
 
-            foreach (var scope in scopes)
-            {
-                logger.LogInformation(scope);
-            }
-
+            // Print URL at the top and bottom because it's a long list of scopes.
+            logger.LogInformation(UrlMessage + "\n");
+            logger.LogInformation(string.Join(Environment.NewLine, scopes));
             logger.LogInformation("\n" + UrlMessage);
             return 0;
         }
--- a/src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
+++ b/src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
@@ -3,7 +3,6 @@ using McMaster.Extensions.CommandLineUtils;
 
 using Microsoft.Authentication.AdoPat;
 using Microsoft.Extensions.Logging;
-using Microsoft.Office.Lasso.Telemetry;
 
 using System.Linq;
 
@@ -16,6 +15,8 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Ado.Pat
     public class CommandScopes
     {
         private readonly ILogger logger;
+
+        private readonly string UrlMessage = $"See {AdoPat.Constants.PatListURL} for details.";
 
         /// <summary>
         /// Create a CommandScopes
@@ -32,6 +33,9 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Ado.Pat
         /// <returns></returns>
         public int OnExecute()
         {
+            // Print at the top and bottom because it's a long list.
+            logger.LogInformation(UrlMessage + "\n");
+
             var scopes = Scopes.ValidScopes.ToList();
             scopes.Sort();
 
@@ -40,6 +44,7 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Ado.Pat
                 logger.LogInformation(scope);
             }
 
+            logger.LogInformation("\n" + UrlMessage);
             return 0;
         }
     }
--- a/src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
+++ b/src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
@@ -1,0 +1,46 @@
+
+using McMaster.Extensions.CommandLineUtils;
+
+using Microsoft.Authentication.AdoPat;
+using Microsoft.Extensions.Logging;
+using Microsoft.Office.Lasso.Telemetry;
+
+using System.Linq;
+
+namespace Microsoft.Authentication.AzureAuth.Commands.Ado.Pat
+{
+    /// <summary>
+    /// Command to print the list of avaialable scopes
+    /// </summary>
+    [Command("scopes", Description = "List the valid ado pat scopes")]
+    public class CommandScopes
+    {
+        private readonly ILogger logger;
+
+        /// <summary>
+        /// Create a CommandScopes
+        /// </summary>
+        /// <param name="logger"></param>
+        public CommandScopes(ILogger<CommandScopes> logger)
+        {
+            this.logger = logger;
+        }
+
+        /// <summary>
+        /// Executes the Scopes command listing the available scopes
+        /// </summary>
+        /// <returns></returns>
+        public int OnExecute()
+        {
+            var scopes = Scopes.ValidScopes.ToList();
+            scopes.Sort();
+
+            foreach (var scope in scopes)
+            {
+                logger.LogInformation(scope);
+            }
+
+            return 0;
+        }
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Added new sub-command `azureauth ado pat scopes` to list the set of actual scopes the `pat` command validates against and print the short-link to the pat scopes docs.
 
 ## [0.8.4] - 2023-09-05
 ### Changed


src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        /// <summary>
+        /// Executes the Scopes command listing the available scopes
+        /// </summary>
+        /// <returns></returns>
+        public int OnExecute()
+        {
+            var scopes = Scopes.ValidScopes.ToList();
+            scopes.Sort();
```
shalinikhare27, line 36: Is it possible to have duplicate scopes in the list and so instead we use `HashSet` to only print unique scopes?

src/AzureAuth/Commands/Ado/CommandPat.cs
```diff
@@ -25,6 +25,7 @@ namespace Microsoft.Authentication.AzureAuth.Commands.Ado
     /// An ADO Command for creating or fetching, and returning Azure Devops PATs.
     /// </summary>
     [Command("pat", Description = "Create and cache Azure Devops Personal Access Tokens (PATs) using encrypted local storage.")]
+    [Subcommand(typeof(Pat.CommandScopes))]
```
shalinikhare27, line 28: Just out of curiosity, what is the purpose of adding this change? Was this a user feedback or how did you come up with solving this scenario and extending to a new subcommand

src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
```diff
@@ -0,0 +1,46 @@
+
+using McMaster.Extensions.CommandLineUtils;
+
+using Microsoft.Authentication.AdoPat;
+using Microsoft.Extensions.Logging;
+using Microsoft.Office.Lasso.Telemetry;
+
+using System.Linq;
+
+namespace Microsoft.Authentication.AzureAuth.Commands.Ado.Pat
```
shalinikhare27, line 10: In order to make this functionality available, we will need to release a new version of AzureAuth. This leads to the whole OMR update churn that our team has to deal with. I am assuming this is not a breaking change? I will test and verify shortly too. :)

src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
kyle-rader-msft: I believe the motivation for using a HashSet is for constant time lookup for the validation of a set of scopes when running the `ado pat` command. Since HashSet is not inherently ordered, I want to print these alphabetically sorted for ease of use and readability.

src/AzureAuth/Commands/Ado/CommandPat.cs
kyle-rader-msft: I've run into this problem a few times, and again this morning while showing another engineer the command, but then not knowing exactly what the scope was that we needed and couldn't find the docs for it either :(. But I know we had validation of them, so it'd be nice if we could list them. Listing valid options is a very common action in this kind of scenario (e.g. `nvm ls-remote`, `rustup target list`, etc)

src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
kyle-rader-msft: Correct, it is a new sub-command and not one that does authentication. This is not a breaking change, nor is it urgent. I would be happy to help with the update if needed.

src/AzureAuth/Commands/Ado/CommandPat.cs
mijpeterson, line 28: I'm very unfamiliar with CLI standards, so genuinely just curious. Is there a reason for picking a subcommand as opposed to creating a new argument, or having this live under -h?

kyle-rader-msft: Great question, I intentionally did not add this as an option to the existing pat command, because that would create a divergence in it's logic, complicating it's option validation and the main execution code all together. It also would violate single responsibility. 

I didn't put this in the help, because generally you want your help text to be as succinct as possible. It's not full-on documentation and should be quick to read. The output list of scopes is quite long and would obfuscate the rest of the help, which for this command, does have several required options.

src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <returns></returns>
+        public int OnExecute()
+        {
+            // Print at the top and bottom because it's a long list.
+            logger.LogInformation(UrlMessage + "\n");
+
+            var scopes = Scopes.ValidScopes.ToList();
+            scopes.Sort();
+
+            foreach (var scope in scopes)
```
reillysiemens, line 42: Can we just join these on the platform newline and log it once? There are a lot of scopes.

Haard30 (commented):
A bit tangential, how about adding `--search` or `--exists` to this subcommand? As a user I would like if I can just type and find out (may be scopes that match a substring?).  Could be a future improvement though so not blocking this PR.

kyle-rader-msft:
> A bit tangential, how about adding `--search` or `--exists` to this subcommand? As a user I would like if I can just type and find out (may be scopes that match a substring?). Could be a future improvement though so not blocking this PR.

Great idea Haard! This is not an urgent need, I just think it would be really helpful. Adding an option, or even just taking positional arguments for scopes to search for is a good idea too for quick searching. How about filtering against positional args if given?

src/AzureAuth/Commands/Ado/CommandPat.cs
mijpeterson: That makes perfect sense! Thanks for the explanation 😊

shalinikhare27 (approved):

src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
shalinikhare27: I guess duplication will never happen as we are just printing from here - https://github.com/AzureAD/microsoft-authentication-cli/blob/6c80faf99de594f0e9958ccd0276fa2c44c2cf45/src/AdoPat/Scopes.cs#L16 :)

src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
```diff
@@ -0,0 +1,47 @@
...
+using Microsoft.Authentication.AdoPat;
+using Microsoft.Extensions.Logging;
+
+using System;
+using System.Linq;
+
+namespace Microsoft.Authentication.AzureAuth.Commands.Ado.Pat
+{
+    /// <summary>
+    /// Command to print the list of avaialable scopes
```
reillysiemens, line 13:
```suggestion
    /// Command to print the list of available scopes

```

shalinikhare27:
> > A bit tangential, how about adding `--search` or `--exists` to this subcommand? As a user I would like if I can just type and find out (may be scopes that match a substring?). Could be a future improvement though so not blocking this PR.
> 
> Great idea Haard! This is not an urgent need, I just think it would be really helpful. Adding an option, or even just taking positional arguments for scopes to search for is a good idea too for quick searching. How about filtering against positional args if given?

This seems useful. However, we do print the URL in the beginning and at the end so that users can read what that scope means. Personally, if I know about a valid scope already, I wouldn't be coming and searching here. And if I did not know about a valid scope, the subcommand output is enough for me to sneak peek the scope I care about and then go to the official scope URL https://aka.ms/azure-devops-pat-scopes to learn what permissions it grants me (specifically the `Description` column). But that's just me and there might be other scenarios I have not ran into.

src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
reillysiemens: Their order in `Scopes.ValidScopes` (as defined) is alphabetically by the _groups_ you see in [https://aka.ms/azure-devops-pat-scopes](https://aka.ms/azure-devops-pat-scopes), but we couldn't do that here without changing the data structure that's used.

Haard30:
> > > A bit tangential, how about adding `--search` or `--exists` to this subcommand? As a user I would like if I can just type and find out (may be scopes that match a substring?). Could be a future improvement though so not blocking this PR.
> > 
> > 
> > Great idea Haard! This is not an urgent need, I just think it would be really helpful. Adding an option, or even just taking positional arguments for scopes to search for is a good idea too for quick searching. How about filtering against positional args if given?
> 
> This seems useful. However, we do print the URL in the beginning and at the end so that users can read what that scope means. Personally, if I know about a valid scope already, I wouldn't be coming and searching here. And if I did not know about a valid scope, the subcommand output is enough for me to sneak peek the scope I care about and then go to the official scope URL https://aka.ms/azure-devops-pat-scopes to learn what permissions it grants me (specifically the `Description` column). But that's just me and there might be other scenarios I have not ran into.

As you said, if someone knows about valid scope, they might not use this subcommand too. And if they are using (or if I am using), I would appreciate tool providing me capability to search it rather than just print everything it knows (For me, it's about scrolling this growing list vs just searching what's relevant to me like we do in notepad/VS Code). But this me thinking purely from User Experience perspective. I see your point and I am fine with its current state too! :)

reillysiemens:
I would strongly prefer that people refer to [https://aka.ms/azure-devops-pat-scopes](https://aka.ms/azure-devops-pat-scopes) (if they can find it). I would like for the API documentation itself to remain the source of truth. We should probably update the `azureauth ado pat` subcommand help text to refer to that link so that it's easy to find even if you haven't found the `scopes` subcommand.

That said, I think having the locality of information by running `azureauth ado pat scopes` available to you is good and helpful. It's a quick reference of the scopes that **this tool** validates. I think that's worthwhile. However, I would be hesitant to add more features than just listing the scopes because I want folks to refer to the API docs. For example, should we also have a description of what each scope does? I'd rather just point folks to the docs.

TL;DR - This seems like a good change, but I'd be hesitant to add much more here.

Haard30:
> I would strongly prefer that people refer to https://aka.ms/azure-devops-pat-scopes (if they can find it). I would like for the API documentation itself to remain the source of truth. We should probably update the `azureauth ado pat` subcommand help text to refer to that link so that it's easy to find even if you haven't found the `scopes` subcommand.
> 
> That said, I think having the locality of information by running `azureauth ado pat scopes` available to you is good and helpful. It's a quick reference of the scopes that **this tool** validates. I think that's worthwhile. However, I would be hesitant to add more features than just listing the scopes because I want folks to refer to the API docs. For example, should we also have a description of what each scope does? I'd rather just point folks to the docs.
> 
> TL;DR - This seems like a good change, but I'd be hesitant to add much more here.

I think that makes sense. Personally, I would prefer search rather than long scroll (and most start with `vso` and make it difficult to spot). 🤷

Haard30 (approved):

mijpeterson (dismissed):

reillysiemens (dismissed):

src/AzureAuth/Commands/Ado/Pat/CommandScopes.cs
reillysiemens: @kyle-rader-msft, I hope you don't mind, I just applied this small typo fix for you.

shalinikhare27 (dismissed):

reillysiemens (dismissed):

kyle-rader-msft:
> > I would strongly prefer that people refer to https://aka.ms/azure-devops-pat-scopes (if they can find it). I would like for the API documentation itself to remain the source of truth. We should probably update the `azureauth ado pat` subcommand help text to refer to that link so that it's easy to find even if you haven't found the `scopes` subcommand.
> > That said, I think having the locality of information by running `azureauth ado pat scopes` available to you is good and helpful. It's a quick reference of the scopes that **this tool** validates. I think that's worthwhile. However, I would be hesitant to add more features than just listing the scopes because I want folks to refer to the API docs. For example, should we also have a description of what each scope does? I'd rather just point folks to the docs.
> > TL;DR - This seems like a good change, but I'd be hesitant to add much more here.
> 
> I think that makes sense. Personally, I would prefer search rather than long scroll (and most start with `vso` and make it difficult to spot). 🤷

An argument against adding search or filtering, is single responsibility. As a command line tool, if you want to search the output of _anything_, there are better tools for that than baking it in yourself. For example, I don't want to add fuzzy searching or regex. 

So instead, I think it is best to rely on the ecosystem of your shell and if a user wants to search, they search they way they hopefully normally do, with `ripgrep` or `grep`.

```
bin\win\azureauth.cmd ado pat scopes | rg code
vso.code
vso.code_full
vso.code_manage
vso.code_status
vso.code_write
```

mvanchaa (approved):

kyle-rader-msft:
@shalinikhare27 , @reillysiemens, @Haard30 would you mind re-reviewing. I left a counter argument for not adding any search and just have `list` do exactly that.

shalinikhare27:
> > > I would strongly prefer that people refer to https://aka.ms/azure-devops-pat-scopes (if they can find it). I would like for the API documentation itself to remain the source of truth. We should probably update the `azureauth ado pat` subcommand help text to refer to that link so that it's easy to find even if you haven't found the `scopes` subcommand.
> > > That said, I think having the locality of information by running `azureauth ado pat scopes` available to you is good and helpful. It's a quick reference of the scopes that **this tool** validates. I think that's worthwhile. However, I would be hesitant to add more features than just listing the scopes because I want folks to refer to the API docs. For example, should we also have a description of what each scope does? I'd rather just point folks to the docs.
> > > TL;DR - This seems like a good change, but I'd be hesitant to add much more here.
> > 
> > 
> > I think that makes sense. Personally, I would prefer search rather than long scroll (and most start with `vso` and make it difficult to spot). 🤷
> 
> An argument against adding search or filtering, is single responsibility. As a command line tool, if you want to search the output of _anything_, there are better tools for that than baking it in yourself. For example, I don't want to add fuzzy searching or regex.
> 
> So instead, I think it is best to rely on the ecosystem of your shell and if a user wants to search, they search they way they hopefully normally do, with `ripgrep` or `grep`.
> 
> ```
> bin\win\azureauth.cmd ado pat scopes | rg code
> vso.code
> vso.code_full
> vso.code_manage
> vso.code_status
> vso.code_write
> ```

This I totally agree with. I can use `grep` or leverage any other search capability if I am really looking for a scope filter. I think still adding a sub-command to print scopes make sense but anything beyond that is a stretch. :)

Haard30:
> @shalinikhare27 , @reillysiemens, @Haard30 would you mind re-reviewing. I left a counter argument for not adding any search and having this list action do exactly that, list only.

Thanks Kyle and Shalini. I believe all good applications extend beyond single responsibility to improve usability (an argument can be made on how social media applications have included reels and other stuff when there were already other platforms doing that). While I am not completely onboard with single responsibility here, I am completely onboard with not extending the functionality for this :). Approving!

Haard30 (approved):

PR #367, opened 2024-02-22
Adding comment to safeguard PAT validTo period

mijpeterson:
# Safeguarding PAT validTo period

## Background
The Azure DevOps team recently made a change to the PAT API such that any requests to generate PATs that last more than 7 days will fail. In the future, they intend to change the behavior such that a request for a greater than 7 day PAT will return a 7 day PAT.

## What did I do?
I dug through the code to ensure that we don't attempt to request for a longer living PAT, and was happy to see that we request for a 7 day PAT. Since people can change code, I decided to add a comment outlining the restriction, and why it's in place. Hopefully, in the future, people will read that comment before attempting to change the value.

### DIFF
--- a/src/AdoPat/PatManager.cs
+++ b/src/AdoPat/PatManager.cs
@@ -36,12 +36,7 @@ namespace Microsoft.Authentication.AdoPat
         /// <param name="client">Any class that implements the <see cref="IPatClient"/> interface.</param>
         /// <param name="now">A function for computing the current moment. Defaults to null, which
         /// uses <see cref="DateTime.UtcNow"/>. Overriding this should only be necessary in testing.</param>
-        public PatManager(
-            ILogger logger,
-            IPatCache cache,
-            IPatClient client,
-            Func<DateTime> now = null
-        )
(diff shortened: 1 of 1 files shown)


Haard30 (approved):

src/AdoPat/PatManager.cs
```diff
@@ -17,6 +17,7 @@ namespace Microsoft.Authentication.AdoPat
     /// </remarks>
     public class PatManager
     {
+        // The ValidToExtensionDays value cannot exceed 7. The Azure DevOps PAT API will not allow a PAT to be valid for more than 7 days, and may return an error if the value is higher.
```
rewrlution, line 20: might be a good idea to set a variable that holds this magic number?

rewrlution: Oh wait, this is already a const

rewrlution: nvm

rewrlution (approved):

shalinikhare27 (approved):

reillysiemens (approved):

PR #383, opened 2024-05-25
Switch out PATs for AAD Tokens in "Build and Test" workflow

mijpeterson:
As the title says, this PR updates the Build and Test workflow to now dynamically retrieve an access token to pass into dotnet restore, as opposed to relying on a manually maintained PAT.

We also modified the CodeQL actions to only run on Windows since CodeQL interferes with other jobs using Azure Powershell on a Mac agent. Since CodeQL is really just a code inspection tool, there isn't really a need to run it on multiple OSes (Windows, Mac, Linux)

### DIFF
--- a/.github/actions/get-ado-token/action.yml
+++ b/.github/actions/get-ado-token/action.yml
@@ -1,0 +1,34 @@
+name: Get Azure DevOps Access Token
+inputs:
+  client-id:
+    description: "The client ID of the application calling Azure DevOps"
+    required: true
+  tenant-id:
+    description: "The tenant ID of the application calling Azure DevOps"
+    required: true
+  organization:
+    description: "The Azure DevOps organization to authenticate with"
+    required: true
+outputs:
+  token:
+    description: "The access token to authenticate with Azure DevOps"
+    value: ${{ steps.ADOAuth.outputs.token }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: OIDC Login with AzPowershell
+      uses: azure/login@v2
+      with:
+        client-id: ${{ inputs.client-id }}
+        tenant-id: ${{ inputs.tenant-id }}
+        allow-no-subscriptions: true
+        enable-AzPSSession: true
+    - id: ADOAuth
+      name: Get ADO Access Token
+      uses: azure/powershell@v1
+      with:
+        azPSVersion: "latest"
+        inlineScript: |
+          $accessToken = (Get-AzAccessToken -ResourceUrl "https://${{inputs.organization}}.visualstudio.com").Token
+          "token=$accessToken" | Out-File -FilePath $env:GITHUB_OUTPUT -Append
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -44,7 +44,7 @@ jobs:
 
   analyze:
     name: Analyze
-    runs-on: windows-latest
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
     permissions:
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -9,6 +9,8 @@ on:
 jobs:
   test:
     runs-on: ${{ matrix.os }}
+    permissions:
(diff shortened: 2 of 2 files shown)


Haard30 (approved):

emilyyang-ms (approved):

.github/actions/get-ado-token/action.yml
```diff
@@ -0,0 +1,34 @@
...
+  organization:
+    description: "The Azure DevOps organization to authenticate with"
+    required: true
+outputs:
+  token:
+    description: "The access token to authenticate with Azure DevOps"
+    value: ${{ steps.ADOAuth.outputs.token }}
+
+runs:
+  using: "composite"
```
rewrlution, line 18: Can you tell me more about the `composite` keyword?

.github/actions/get-ado-token/action.yml
```diff
@@ -0,0 +1,34 @@
...
+  steps:
+    - name: OIDC Login with AzPowershell
+      uses: azure/login@v2
+      with:
+        client-id: ${{ inputs.client-id }}
+        tenant-id: ${{ inputs.tenant-id }}
+        allow-no-subscriptions: true
+        enable-AzPSSession: true
```
rewrlution, line 26: just curious, will the `AzPSSession` be closed automatically after usage?

.github/actions/get-ado-token/action.yml
```diff
@@ -0,0 +1,34 @@
...
+    - id: ADOAuth
+      name: Get ADO Access Token
+      uses: azure/powershell@v1
+      with:
+        azPSVersion: "latest"
+        inlineScript: |
+          $accessToken = (Get-AzAccessToken -ResourceUrl "https://${{inputs.organization}}.visualstudio.com").Token
+          "token=$accessToken" | Out-File -FilePath $env:GITHUB_OUTPUT -Append
```
rewrlution, line 34: Out of curiosity, is there another mode called `-Write`? Why do we want to do an `-Append` here?

.github/workflows/dotnet-test.yml
```diff
@@ -34,21 +44,28 @@ jobs:
 
   analyze:
     name: Analyze
-    runs-on: ubuntu-latest
+    runs-on: windows-latest
```
rewrlution, line 47: Any reason why we change the os?

rewrlution (commented):
The overall structure looks good to me. Just have some questions about the implementation detail.

.github/actions/get-ado-token/action.yml
mijpeterson: Great question! It will indeed be closed after usage. This task injects what's known as a "post" action at the end of the workflow run. This is visible [here](https://github.com/AzureAD/microsoft-authentication-cli/actions/runs/9231515558/job/25401557591?pr=383) under "Post Get Azure DevOps Access Token"

.github/workflows/dotnet-test.yml
mijpeterson: Another great question! I touched on it briefly in the description of the PR but not in very much detail.

The gist is that this os declaration was actually misleading before since it actually ran on three different OSes (ubuntu, macOS, and windows) due to the "matrix" declaration a few lines down. And we wouldn't have changed anything if it wasn't for the fact that we noticed that CodeQL messed with our Azure PowerShell tasks specifically on the macOS runner. Digging into it further, we came to the conclusion that CodeQL doesn't actually need to run on all OSes since it's a code inspection tool as opposed to one that inspects build outputs. So I ended up just picking windows to run CodeQL on.

mijpeterson: Testing it on Linux, it's actually much faster. So I'm going to convert it to ubuntu.

Haard30 (approved):

emilyyang-ms (approved):

.github/actions/get-ado-token/action.yml
rewrlution: great. good to know!

rewrlution (approved):
LGTM

PR #384, opened 2024-05-25
Removing PATs from Release Workflow

mijpeterson:
This PR substitutes PATs in the release workflow with dynamically generated access tokens.

We also separate out the CodeQL actions to a separate Windows-only job since CodeQL interferes with other jobs using Azure Powershell on a Mac agent. Since CodeQL is really just a code inspection tool, there isn't really a need to run it on multiple OSes (Windows, Mac, Linux)

### DIFF
--- a/bin/trigger_azure_pipelines.py
+++ b/bin/trigger_azure_pipelines.py
@@ -12,7 +12,7 @@ from azure.devops.connection import Connection
 from azure.devops.v6_0.pipelines.pipelines_client import PipelinesClient
 from azure.devops.v6_0.build.build_client import BuildClient
 from azure.devops.v6_0.build.models import TimelineRecord
-from msrest.authentication import BasicAuthentication
+from msrest.authentication import BasicTokenAuthentication
 from requests import Response
 
 # https://learn.microsoft.com/en-us/rest/api/azure/devops/build/timeline/get#taskresult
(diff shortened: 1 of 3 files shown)


Haard30 (dismissed):

Haard30 (approved):

bin/trigger_azure_pipelines.py
```diff
@@ -123,7 +123,7 @@ def main() -> None:
     try:
         # ADO PAT (Azure DevOps Personal Access Token) with "Build" scope.
```
emilyyang-ms, line 124: Shall we update this comment?

bin/trigger_azure_pipelines.py
```diff
@@ -123,7 +123,7 @@ def main() -> None:
     try:
         # ADO PAT (Azure DevOps Personal Access Token) with "Build" scope.
         # More information here - https://learn.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate?view=azure-devops&tabs=Windows#create-a-pat
```
emilyyang-ms, line 125: Update this comment?

emilyyang-ms (dismissed):

bin/trigger_azure_pipelines.py
Haard30: That's a good catch, Emily. I removed the comment as it is no longer relevant and var name is self-explanatory to represent ADO Token.

Haard30 (approved):

emilyyang-ms (approved):

PR #386, opened 2024-06-05
Moving Release Workflow to Azure Pipelines

mijpeterson:
The secure future initiative is making it increasingly difficult to continue supporting the build and release of "production" tools via GitHub Actions/Workflows. This appears to be because of the fact that GitHub itself is considered an untrusted "environment", especially from the perspective of our production tenants.

[This official Microsoft Open Source doc](https://docs.opensource.microsoft.com/security/azure/) explains those considerations in detail and are a good signal for us to move to Azure Pipelines.

This PR will be the first in a series designed to move us away from GitHub workflows to Azure Pipelines. It introduces a new pipeline that will live side-by-side with our existing workflow. This new pipeline will act as a testing ground for us, in which we'll iteratively port over steps from our workflow.

Unfortunately, as there is no way to test out these pipelines on a GitHub connected repo until they're merged into main, there will likely be many PRs introducing experimental changes to the pipeline.

### DIFF
--- a/.github/workflows/release-azure-pipelines.yml
+++ b/.github/workflows/release-azure-pipelines.yml
@@ -79,7 +79,7 @@ extends:
             inputs:
               command: restore
               feedsToUse: select
-              vstsFeed: $(feedId)
+              vstsFeed: $(vstsFeedId)
               includeNuGetOrg: false
               arguments: '--runtime ${{ matrix.runtime }}'
 
--- a/.github/workflows/release-azure-pipelines.yml
+++ b/.github/workflows/release-azure-pipelines.yml
@@ -17,7 +17,7 @@ resources:
     ref: refs/tags/release
 
 extends:
-  template: v1/Office.Official.PipelineTemplate.yml@CustomPipelineTemplates
+  template: v1/Office.Unofficial.PipelineTemplate.yml@CustomPipelineTemplates
   parameters:
     pool:
       name: Azure-Pipelines-1ESPT-ExDShared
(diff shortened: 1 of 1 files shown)


.github/workflows/release-azure-pipelines.yml
```diff
@@ -0,0 +1,103 @@
+# This pipeline will be triggered when either main branch is pushed or 2AM on workdays.
+variables:
+- name: tags
+  value: "nonproduction"
```
shalinikhare27, line 4: Are we temporarily calling this a non-prod pipeline?

mijpeterson: Great question! I am intentionally configuring this as non-production for now, while we are still experimenting with it. That's because I don't want to accidentally trigger any S360 KPIs while we're drafting up the final version of the pipeline. Once we have it closer to completion, I intend to switch it to production (and use the Official pipeline template)

shalinikhare27 (dismissed):

.github/workflows/release-azure-pipelines.yml
```diff
@@ -0,0 +1,103 @@
+# This pipeline will be triggered when either main branch is pushed or 2AM on workdays.
```
emilyyang-ms, line 1: Where is the 2am on workday stuff defined?

emilyyang-ms (dismissed):

.github/workflows/release-azure-pipelines.yml
mijpeterson: Oh you know what! I copied this over from the CG pipeline, and the comment is no longer relevant. I'll remove it. Thanks for catching that!

mijpeterson: Thinking about it more, I'm actually going to change it to only be manually triggered (to better replicate the existing workflow).

emilyyang-ms (approved):

Haard30 (approved):

PR #393, opened 2024-06-27
AzureAuth GitHub release through ADO pipeline 

Haard30:
This PR adds the ADO pipeline that releases AzureAuth to GitHub.

This pipeline has similar behavior to original [github release](https://github.com/AzureAD/microsoft-authentication-cli/blob/main/.github/workflows/release.yml), except for a few things:

1. We no longer require SP secrets and instead use federated creds.
2. We use the 1ES Pipeline template. (We will switch to `official` pipeline template and mark the pipeline as `prod` ~~in a follow-up PR~~). REVIEW the [PR](https://github.com/AzureAD/microsoft-authentication-cli/pull/394)
3. Code signing - Our existing code signing mechanism of downloading the cert from Azure KV and storing it locally on windows cert store is no longer compatible with 1ES PT. So instead, we use the built-in ADO task (similar to how other teams perform code signing with ESRP). We should now be able to remove [sign.py](https://github.com/AzureAD/microsoft-authentication-cli/blob/main/bin/sign.py) which leads to lesser code and dependencies to maintain (in a follow-up PR).

Here's the draft release created through this pipeline: https://github.com/AzureAD/microsoft-authentication-cli/releases/tag/untagged-c463728100f3301ee7c8

Testing:
1. Test end to end actual release of AureAuth using this pipeline.
2. Download, install and run AureAuth published through release.
3. Verify the code signing by looking at code signing summary generated through esrp task.
4. Reject/Approve the release and see behavior is correctly reflected.

### DIFF
--- a/.github/workflows/release-azure-pipelines.yml
+++ b/.github/workflows/release-azure-pipelines.yml
@@ -340,7 +340,6 @@ extends:
             tagSource: 'userSpecifiedTag'
             tag: ${{ parameters.version }}
             isPrerelease: ${{ parameters.prerelease }}
-            isDraft: true
             addChangeLog: false
             releaseNotesSource: 'inline'
             releaseNotesInline: "Release ${{ parameters.version }}. See [`CHANGELOG.md`](https://github.com/AzureAD/microsoft-authentication-cli/blob/${{ parameters.version }}/CHANGELOG.md) for updates."
--- a/.github/workflows/release-azure-pipelines.yml
+++ b/.github/workflows/release-azure-pipelines.yml
@@ -319,10 +319,7 @@ extends:
       jobs:
       - job: approval
         displayName: Manual Approval
-        pool:
-          name: Azure-Pipelines-1ESPT-ExDShared
-          image: ubuntu-latest
-          os: linux
+        pool: server
         timeoutInMinutes: 5760 # job times out in 4 days
         steps:
         - task: ManualValidation@0
--- a/.github/workflows/release-azure-pipelines.yml
+++ b/.github/workflows/release-azure-pipelines.yml
@@ -30,7 +30,7 @@ parameters:
   
 variables:
 - name: tags
-  value: "nonproduction"
+  value: "production"
   readonly: true
 - name: pythonVersion
   value: 3.10
--- a/.github/workflows/release-azure-pipelines.yml
+++ b/.github/workflows/release-azure-pipelines.yml
@@ -317,8 +317,22 @@ extends:
       displayName: Release
       dependsOn: package
       jobs:
+      - job: approval
+        displayName: Manual Approval
+        pool:
+          name: Azure-Pipelines-1ESPT-ExDShared
+          image: ubuntu-latest
+          os: linux
+        timeoutInMinutes: 5760 # job times out in 4 days
+        steps:
+        - task: ManualValidation@0
+          timeoutInMinutes: 4320 # task times out in 3 days
+          inputs:
+            notifyUsers: $(REVIEWER)
+            instructions: 'Review the AzureAuth GitHub Release.'
       - job: release
         displayName: Release
+        dependsOn: approval
         pool:
           name: Azure-Pipelines-1ESPT-ExDShared
           image: ubuntu-latest
--- a/.github/workflows/release-azure-pipelines.yml
+++ b/.github/workflows/release-azure-pipelines.yml
@@ -137,7 +137,6 @@ extends:
           displayName: Signing ${{ config.runtime }}
           pool:
             name: Azure-Pipelines-1ESPT-ExDShared
-            # This step has to run on Windows because ESRPClient.exe is currently only available for that platform.
             image: windows-latest
             os: windows
           templateContext:
--- a/.github/workflows/release-azure-pipelines.yml
+++ b/.github/workflows/release-azure-pipelines.yml
@@ -2,7 +2,6 @@
 parameters:
 - name: version
   type: string
-  default: 0.8.6
 - name: prerelease
   displayName: Prerelease?
   type: boolean
@@ -38,6 +37,7 @@ variables:
   readonly: true
 
 trigger: none
+
 pr: none
 
 resources:
@@ -66,16 +66,16 @@ extends:
       - job: validate
         displayName: Validate
         steps:
-          - checkout: self
-          - task: UsePythonVersion@0
-            displayName: Use Python $(pythonVersion)
-            inputs:
-              versionSpec: $(pythonVersion)
-          - task: Bash@3
-            inputs:
-              targetType: inline
-              script: |
-                echo ${{ parameters.version }} | python ./bin/version.py
+        - checkout: self
+        - task: UsePythonVersion@0
+          displayName: Use Python $(pythonVersion)
+          inputs:
+            versionSpec: $(pythonVersion)
+        - task: Bash@3
+          inputs:
+            targetType: inline
+            script: |
+              echo ${{ parameters.version }} | python ./bin/version.py
 
     - stage: build
       displayName: Build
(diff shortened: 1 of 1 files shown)


.github/workflows/release-azure-pipelines.yml
```diff
@@ -93,40 +93,268 @@ extends:
...
+          - task: EsrpCodeSigning@5
+            displayName: Sign artifacts win10-x64
+            condition: eq('${{ config.runtime }}', 'win10-x64')
+            inputs:
+              ConnectedServiceName: $(esrpKVServiceConnection)
+              AppRegistrationClientId: $(SIGNING_AAD_ID)
+              AppRegistrationTenantId: $(SIGNING_TENANT_ID)
+              AuthAKVName: $(AZURE_VAULT)
+              AuthCertName: $(AZURE_VAULT_ESRP_AAD_CERT_NAME)
+              AuthSignCertName: $(AZURE_VAULT_ESRP_REQ_CERT_NAME)
```
mijpeterson, line 161: Just confirming, but are all of these going to be Pipeline variables? If so, should we use consistent casing for variable names? (I'm not picky on which casing style we choose, so long as we stay consistent)

.github/workflows/release-azure-pipelines.yml
```diff
@@ -93,40 +93,268 @@ extends:
...
+    - stage: release
+      displayName: Release
+      dependsOn: package
+      jobs:
+      - job: approval
+        displayName: Manual Approval
+        pool: server
+        timeoutInMinutes: 5760 # job times out in 4 days
+        steps:
+        - task: ManualValidation@0
```
mijpeterson, line 324: Ooo this is a clever addition. I suspect that in the future, when release pipeline templates become a thing, that this could end up being a lockbox approval instead of a pipeline approval (if we're required to use MOBR). But we'll see what that looks like when we get there!

mijpeterson (changes requested):
Just the one comment on pipeline variable casing consistency. No preference which way we go.

.github/workflows/release-azure-pipelines.yml
```diff
@@ -93,40 +93,268 @@ extends:
...
+            displayName: Build artifacts
+            env:
+              ADO_TOKEN: $(System.AccessToken)
+            inputs:
+              command: publish
+              projects: src/AzureAuth/AzureAuth.csproj
+              arguments: -p:Version=${{ parameters.version }} --configuration release --self-contained true --runtime ${{ config.runtime }} --output dist/${{ config.runtime }}
+              publishWebProjects: false
+              zipAfterPublish: false
+              modifyOutputPath: true
```
mvanchaa, line 129: Sorry if this is a dumb question - Is the [Analyze](https://github.com/AzureAD/microsoft-authentication-cli/blob/ea5bf2cfd9db97e96394f5051a033aac4d32b23c/.github/workflows/release.yml#L88C3-L122C37) job which does CodeQL analysis already included in the 1ES pipeline template?

mvanchaa, line 129: Does this task generate the artifacts to the `$(Build.ArtifactStagingDirectory)` directory? And that's how the artifacts are being used in line 146?

.github/workflows/release-azure-pipelines.yml
```diff
@@ -93,40 +93,268 @@ extends:
...
+            displayName: Extract signed artifacts osx
+            condition: startsWith('${{ config.runtime }}', 'osx')
+            inputs:
+              archiveFilePatterns: $(Build.ArtifactStagingDirectory)/azureauth-${{ parameters.version }}-${{ config.runtime }}.zip
+              destinationFolder: $(Build.ArtifactStagingDirectory)/azureauth-${{ parameters.version }}-${{ config.runtime }}
+              cleanDestinationFolder: true
+              overwriteExistingFiles: true
+          # We rename the signed artifacts to avoid conflicts with the unsigned pipeline artifacts from the previous stage.
+          - task: PowerShell@2
+            displayName: Rename signed artifacts
```
mvanchaa, line 245: Is `ExtractFiles@1` going to replace the `azureauth-${{ parameters.version }}-${{ config.runtime }}` file generated during the build task (if any)? Is it possible to have the `destinationFolder` value be `azureauth-${{ parameters.version }}-${{ config.runtime }}-signed` instead and avoid the `PowerShell@2` task?

.github/workflows/release-azure-pipelines.yml
```diff
@@ -93,40 +93,268 @@ extends:
...
+            inputs:
+            - ${{ each config in parameters.buildConfigs }}:
+              - input: pipelineArtifact
+                artifactName: azureauth-${{ parameters.version }}-${{ config.runtime }}-signed
+                targetPath: $(Build.ArtifactStagingDirectory)/azureauth-${{ parameters.version }}-${{ config.runtime }}-signed
+            outputs:
+            - ${{ each config in parameters.buildConfigs }}:
+              - output: pipelineArtifact
+                artifactName: azureauth-${{ parameters.version }}-${{ config.runtime }}.${{ config.archiveExt }}
+                targetPath: $(Build.SourcesDirectory)/azureauth-${{ parameters.version }}-${{ config.runtime }}.${{ config.archiveExt }}
```
mvanchaa, line 273: If the `targetPath` here is used to place the final artifacts, you can try using `Build.BinariesDirectory` instead as `$(Build.SourcesDirectory)` should ideally be reserved for the source code files. See https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#build-variables-devops-services for more information on pre-defined pipeline variables.

.github/workflows/release-azure-pipelines.yml
```diff
@@ -93,40 +93,268 @@ extends:
...
+          displayName: Create AzureAuth GitHub Release
+          inputs:
+            gitHubConnection: $(githubReleaseServiceConnection)
+            repositoryName: 'AzureAD/microsoft-authentication-cli'
+            action: 'create'
+            target: $(Build.SourceVersion)
+            tagSource: 'userSpecifiedTag'
+            tag: ${{ parameters.version }}
+            isPrerelease: ${{ parameters.prerelease }}
+            isDraft: true
```
mvanchaa, line 353: Are you going to update `isDraft` in further PRs?

.github/workflows/release-azure-pipelines.yml
Haard30: That's a good question, Manuha. Yes, SDL Analysis is auto injected by 1ES PT.

Haard30: That's almost correct. But instead, it generates artifacts to `dist/{{config.runtime}}` and we mention that path in the `templateContext` output `targetpath`. The artifacts mentioned in output are then uploaded (this time from `$(Build.ArtifactStagingDirectory)`)so it can be downloaded by signing stage. Similarly for signing stage where signed artifacts are uploaded at the end of stage for packaging stage.

.github/workflows/release-azure-pipelines.yml
Haard30: The `extractFiles` task is only required for macos where we need to zip before sending to esrp. However, renaming is required for all the runtimes. So I have kept this as a separate task representing the intentional rename common for all runtimes.

.github/workflows/release-azure-pipelines.yml
Haard30: Correct, I will switch it to false when we use official template and switch pipeline to prod

.github/workflows/release-azure-pipelines.yml
Haard30: So the `targetPath` here refers to from which path should artifacts be uploaded at the end of stage. Since the packaging step compresses and puts it in sources directory, I have provided that path (as compressed file is only available at that path to upload it). I tried searching for specifying the target directory for compressed file itself so the pipeline can pick up from there to upload it but I was not able to find that location (It compresses and puts it in the same place from where it compressed). I can have an additional task to copy the compressed files to `BinariesDirectory`, but I am not sure it will add any value (as this is just the location to pick up from)

.github/workflows/release-azure-pipelines.yml
```diff
@@ -93,40 +93,268 @@ extends:
...
+              cleanDestinationFolder: true
+              overwriteExistingFiles: true
+          # We rename the signed artifacts to avoid conflicts with the unsigned pipeline artifacts from the previous stage.
+          - task: PowerShell@2
+            displayName: Rename signed artifacts
+            inputs:
+              workingDirectory: $(Build.ArtifactStagingDirectory)
+              targetType: 'inline'
+              script: |
+                mv "azureauth-${{ parameters.version }}-${{ config.runtime }}" "azureauth-${{ parameters.version }}-${{ config.runtime }}-signed"
```
mijpeterson, line 250: Do you run into conflicts if you use the same file name, but a different artifact name?

.github/workflows/release-azure-pipelines.yml
mijpeterson: That's so weird that mac has this one specific requirement for code-signing! I wonder why that is.

.github/workflows/release-azure-pipelines.yml
Haard30: Done!

.github/workflows/release-azure-pipelines.yml
Haard30: It is unfortunate! I tried signing without zipping and got `Incorrect file type used for command -z. It needs to be a .zip or .dmg file.\n`. Later I looked at other pipelines and realized we do need to zip.

.github/workflows/release-azure-pipelines.yml
mvanchaa: Awesome!

.github/workflows/release-azure-pipelines.yml
Haard30: Do you mean different artificat name in the `output`? If so, I tried doing that, but it couldn't find the path. It searches for the exact file.

mijpeterson: Yep I meant exactly that! And that's surprising! I always thought of artifacts as different "buckets" of files. I wonder if maybe there's a "bug" in how 1ES PT consume artifacts via the "inputs" template context. But either way, not a big deal :)

.github/workflows/release-azure-pipelines.yml
Haard30: Created PR to this branch https://github.com/AzureAD/microsoft-authentication-cli/pull/394

Haard30:
Please review the PR to switch to prod/official template https://github.com/AzureAD/microsoft-authentication-cli/pull/394

mijpeterson (approved):

.github/workflows/release-azure-pipelines.yml
Haard30: SBOM (when switching to prod pipeline) requires output to be a folder and not file (like we do zip/tar.gz here...sigh). I am going to change this so that we have `packaged` directory similarly to `signed` one. I will put this change in #394 PR

mvanchaa: Oh then I don't think it's worth doing an additional copy. It makes sense as the final compressed signed files are places in the same location of the original file. It would have been great if the signing task provided an option to specify the output file location.

mvanchaa (approved):

PR #396, opened 2024-07-08
Switch to official prod version

Haard30:


.github/workflows/release-azure-pipelines.yml
```diff
@@ -59,6 +59,8 @@ extends:
         name: Azure-Pipelines-1ESPT-ExDShared
         image: windows-latest
         os: windows
+      roslyn:
+        copyLogsOnly: true
```
reillysiemens, line 63: I think it's worth adding context about how this section (and the explicit build below) prevents an unnecessary second build.

.github/workflows/release-azure-pipelines.yml
```diff
@@ -267,18 +274,24 @@ extends:
...
+              script: |
+                mkdir azureauth-${{ parameters.version }}-packaged
           - task: ArchiveFiles@2
             displayName: Create win10-x64 archive
             inputs:
               rootFolderOrFile: $(Build.ArtifactStagingDirectory)/azureauth-${{ parameters.version }}-win10-x64-signed/AzureAuth
               includeRootFolder: false
               archiveType: zip
-              archiveFile: azureauth-${{ parameters.version }}-win10-x64.zip
+              archiveFile: $(Build.ArtifactStagingDirectory)/azureauth-${{ parameters.version }}-packaged/azureauth-${{ parameters.version }}-win10-x64.zip
```
reillysiemens, line 294: Is it possible for
```
$(Build.ArtifactStagingDirectory)/azureauth-${{ parameters.version }}-packaged/azureauth-${{ parameters.version }}
```
to be defined once somewhere?

.github/workflows/release-azure-pipelines.yml
```diff
@@ -267,18 +274,24 @@ extends:
...
             outputs:
-            - ${{ each config in parameters.buildConfigs }}:
-              - output: pipelineArtifact
-                artifactName: azureauth-${{ parameters.version }}-${{ config.runtime }}.${{ config.archiveExt }}
-                targetPath: $(Build.SourcesDirectory)/azureauth-${{ parameters.version }}-${{ config.runtime }}.${{ config.archiveExt }}
+            - output: pipelineArtifact
+              artifactName: azureauth-${{ parameters.version }}-packaged
+              targetPath: $(Build.ArtifactStagingDirectory)/azureauth-${{ parameters.version }}-packaged
           steps:
+          - task: PowerShell@2
```
mvanchaa, line 281: We didn't have this step for signing. So does ESRP task creates the `$targetPath` folder if it doesn't exist and in the case of packaging, we need to explicitly create the `$targetPath` folder?

Haard30: Right, also in case of packaging, 1ES PT official version only allows us to upload folder artifacts and not packaged files like zip,tar. So, we had to create a `packaged` folder and put all the files there to upload it for release stage . 😅

.github/workflows/release-azure-pipelines.yml
```diff
@@ -335,10 +348,9 @@ extends:
           os: linux
         templateContext:
           inputs:
-          - ${{ each config in parameters.buildConfigs }}:
-            - input: pipelineArtifact
-              artifactName: azureauth-${{ parameters.version }}-${{ config.runtime }}.${{ config.archiveExt }}
-              targetPath: $(Build.ArtifactStagingDirectory)/azureauth-${{ parameters.version }}-${{ config.runtime }}.${{ config.archiveExt }}
+          - input: pipelineArtifact
+            artifactName: azureauth-${{ parameters.version }}-packaged
```
mvanchaa, line 355: Sorry if this is a lame question. In the `package` job, we are not generating any artifacts with the name `azureauth-${{ parameters.version }}-packaged` no? We are generating `azureauth-${{ parameters.version }}-osx-arm64.tar.gz` / `azureauth-${{ parameters.version }}--win10-x64.zip`. Wouldn't this cause an error as the specified artifact name wouldn't exist?

mvanchaa (approved):

.github/workflows/release-azure-pipelines.yml
Haard30: No that's a good question. So, in the package step, we are uploading `packaged` folder with all the packaged files (line 278). We download that entire folder here and upload individual files.

Haard30: All this because we are not allowed to directly upload compressed files

.github/workflows/release-azure-pipelines.yml
Haard30: Done!

.github/workflows/release-azure-pipelines.yml
mvanchaa: ah okay! I thought artifact name has to be a file name. Got it!

mvanchaa (approved):

.github/workflows/release-azure-pipelines.yml
```diff
@@ -30,25 +30,28 @@ parameters:
...
 pr: none
 
 resources:
   repositories:
-  - repository: CustomPipelineTemplates
+  - repository: OfficePipelineTemplates
     type: git
-    name: 1ESPipelineTemplates/OfficePipelineTemplates
-    ref: refs/tags/release
+    name: OE/OfficePipelineTemplates
```
mijpeterson, line 50: Interesting! This is the first I've seen this path used. Do you have a link to documentation that references this project/repo?

.github/workflows/release-azure-pipelines.yml
```diff
@@ -59,6 +62,9 @@ extends:
         name: Azure-Pipelines-1ESPT-ExDShared
         image: windows-latest
         os: windows
+      # This prevents auto-injected Roslyn task from running the build again.
+      roslyn:
+        copyLogsOnly: true
```
mijpeterson, line 67: So what is the default behavior here if we don't specify this config? Is there a separate build job that gets invoked?

If so, I wonder if that is preferable to having an extra "build" step that runs in sequence, as jobs run in parallel.

mijpeterson (commented):
Just a couple of questions

.github/workflows/release-azure-pipelines.yml
Haard30: Lot of OE projects are using this but I am not able to find the link (pinging the link to oe projects async). The doc does mention `1ESPipelineTemplates/OfficePipelineTemplates` in its example. I think OE reference might be an old one. Let me search support channel to find any relevant information and if the new way is using `1ESPipelineTemplates/OfficePipelineTemplates`, I can revert this

.github/workflows/release-azure-pipelines.yml
Haard30: It would still search for a build task but not use its logs. So, it will have its own build along with build that we do separately. We could have a parallel build or something similar, but I am not sure if there are any benefits to doing that.

mijpeterson: One that I can think of is if it's run in a separate job, that it **might** be faster than adding a build step as part of the existing job. But that sort of depends on how long it takes to restore dependencies in that separate job. And whether or not calling "dotnet package" after the new "dotnet build" would re-build the whole project or simply re-use the build cache from dotnet build.

Haard30: That's an interesting point. Doing a bit of napkin math reveals that, on average 24 mins run of this pipeline, it adds about ~1.5% of overhead (25 seconds for this step). Install dependencies is about 40s on average.

mijpeterson: Interesting! That does seem fairly minor. Is install dependencies + build in a separate job that runs in parallel (Roslyn's default behaviour) faster than the added build step that runs in sequence (25 seconds)?

mijpeterson: Or in other words: is the default 1ES PT behaviour faster, even if it's technically doing more work? (assuming it runs in a separate job)

Haard30: Also, without that config, Roslyn injected builds take 35-40s on average compared to 15s with that config. So, it seems we are not (if at all) impacting the overall run time by anything more than 0.5%.

mijpeterson: Ah got it, if it it's also injecting a sequential step and that step takes longer, this definitely seems like the right way forward :) Thank you for digging into this!

mijpeterson (approved):

.github/workflows/release-azure-pipelines.yml
mijpeterson: Just checking if you had found anything in the support channel!

Haard30: Yes! Updated this to `1ESPipelineTemplates/OfficePipelineTemplates`. From what I found; it seems both repositories are kept up to date but `1ESPipelineTemplates/OfficePipelineTemplates` seems to be the latest way to do this so going with that,

mvanchaa (approved):

PR #82, opened 2022-06-02
Upgrade msal to 4.44

goagain:
# Summary
MSAL 4.44 brings a very exciting change - a C++ wrapped interface to using the Windows Broker natively in MSAL without the extra Microsoft.Identity.Client.Desktop dependency.

That library bloats AzureAuth by nearly 50 MB. Why is that bad? We install AzureAuth, a lot. https://somsubhra.github.io/github-release-stats/?username=AzureAd&repository=microsoft-authentication-cli&page=1&per_page=5 - version 0.3.0 has been downloaded over 100K times already. Making the application small means faster install times which means faster enlistment setup and faster first time usage from within apps.

# Goals:
Upgrade to MSAL 4.44
Remove Microsoft.Identity.Client.Desktop
Ensure all current functionality still works, that we can still use this new version and build and run on Intel and Arm Macs.
# Non-Goals:
Any new functionality
Updating the target framework (We should be able to update to dotnet 6 as well - let's do that as a separate task).

# Update
Upgrade `Microsoft.Identity.Client` from 4.43.1 to 4.44.0
Upgrade `Microsoft.Identity.Client.Extensions.Msal` from 2.19.6 to 2.21.0
Remove `Microsoft.Identity.Client.Desktop` on .NET 4.7.2

# Outcome
Releases don't show a significant size difference.
AzureAuth-win64-msal4.43: 73,105 KB
AzureAuth-win64-msal4.44: 72,988 KB

Executed publish command:
```ps1
dotnet publish .\src\AzureAuth\AzureAuth.csproj --arch=x64 --os=win --self-contained --output=msal4.43 --interactive
```

PR #400, opened 2024-08-21
Upgrade MSAL dependencies to 4.61.3

keyuxuan:
## Changes:

1.  Upgrading MSAL version to `4.61.3`
2. Starting from MSAL version `4.61.0`, usage of `WithBroker` method with no parameter is deprecated, so the method [` clientBuilder.WithBroker();`](https://github.com/AzureAD/microsoft-authentication-cli/blob/7ae2857d79b7859990f0b34ba3ad606ce93876c9/src/MSALWrapper.Benchmark/BrokerBenchmark.cs#L96) used in `BrokerBenchmark` class becomes obsolete. 

      According to the [release log](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.61.0), the alternative to that is to 
reference `Microsoft.Identity.Client.Desktop` when [authenticating with browser](https://aka.ms/msal-net-uses-web-browser) and call `WithWindowsEmbeddedBrowserSupport()`.  
      
     I tried that, but looks like `Microsoft.Identity.Client.Desktop` only support Windows platform, it does not support macos and ubuntu platforms, so the  [`Build and Test`](https://github.com/AzureAD/microsoft-authentication-cli/actions/runs/10495034169/job/29072621189)  failed for those 2 platforms from my earlier commits .  

      Also currently in `Broker` class, we are explicitly only using [`interactive authentication with Windows Broker`](https://github.com/AzureAD/microsoft-authentication-cli/blob/7ae2857d79b7859990f0b34ba3ad606ce93876c9/src/MSALWrapper/AuthFlow/Broker.cs#L160), so the else block (which is used for `ManagedBrokerBenchmark `) in `BrokerBenchmark` class seems unnecessary. So i decided to just remove the else block.

### Benchmark test

`4.61.3`

```

BenchmarkDotNet=v0.13.2, OS=Windows 11 (10.0.22631.4037)
Intel Core i9-10900K CPU 3.70GHz, 1 CPU, 20 logical and 10 physical cores
.NET SDK=8.0.304
  [Host]     : .NET 6.0.33 (6.0.3324.36610), X64 RyuJIT AVX2
  DefaultJob : .NET 6.0.33 (6.0.3324.36610), X64 RyuJIT AVX2
```

|                Method |     Mean |    Error |   StdDev |
|---------------------- |---------:|---------:|---------:|
| NativeBrokerBenchmark | 42.02 ms | 0.329 ms | 0.275 ms |

`4.59.1`

```
BenchmarkDotNet=v0.13.2, OS=Windows 11 (10.0.22631.4037)
Intel Core i9-10900K CPU 3.70GHz, 1 CPU, 20 logical and 10 physical cores
.NET SDK=8.0.304
  [Host]     : .NET 6.0.33 (6.0.3324.36610), X64 RyuJIT AVX2
  DefaultJob : .NET 6.0.33 (6.0.3324.36610), X64 RyuJIT AVX2
```

|                 Method |     Mean |    Error |   StdDev |
|----------------------- |---------:|---------:|---------:|
|  NativeBrokerBenchmark | 44.16 ms | 0.149 ms | 0.132 ms |

### Other

Tested following commands on both mac and windows:
```
azureauth --help
azureauth aad --help
azureauth ado pat --help
azureauth ado token --help
azureauth ado pat scopes --help
azureauth info --help
azureauth ado token
azureauth ado pat
azureauth aad (with different output and auth modes)
```

### DIFF
--- a/src/MSALWrapper.Benchmark/BrokerBenchmark.cs
+++ b/src/MSALWrapper.Benchmark/BrokerBenchmark.cs
@@ -7,7 +7,6 @@ namespace Microsoft.Authentication.MSALWrapper.Benchmark
     using Microsoft.Authentication.MSALWrapper;
     using Microsoft.Identity.Client;
     using Microsoft.Identity.Client.Broker;
-    using Microsoft.Identity.Client.Desktop;
     using Microsoft.Extensions.Logging;
     using Microsoft.Authentication.MSALWrapper.AuthFlow;
     using BenchmarkDotNet.Running;
@@ -88,6 +87,8 @@ namespace Microsoft.Authentication.MSALWrapper.Benchmark
                     enablePiiLogging: false,
                     enableDefaultPlatformLogging: true)
 ;
+            clientBuilder.WithBroker(clientBuilder, new BrokerOptions(BrokerOptions.OperatingSystems.Windows));
+            /*
             if (useNativeBroker)
             {
                 BrokerExtension.WithBroker(clientBuilder, new BrokerOptions(BrokerOptions.OperatingSystems.Windows));
@@ -96,7 +97,7 @@ namespace Microsoft.Authentication.MSALWrapper.Benchmark
             {
                 clientBuilder.WithWindowsEmbeddedBrowserSupport();
             }
-
+            */
             return new PCAWrapper(logger, clientBuilder.Build(), errors, tenantId);
         }
         private void LogMSAL(Identity.Client.LogLevel level, string message, bool containsPii)
--- a/src/MSALWrapper.Benchmark/BrokerBenchmark.cs
+++ b/src/MSALWrapper.Benchmark/BrokerBenchmark.cs
@@ -44,7 +44,6 @@ namespace Microsoft.Authentication.MSALWrapper.Benchmark
         public void WarmUp()
         {
             NativeBrokerBenchmark();
-            ManagedBrokerBenchmark();
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


shalinikhare27:
I am curious that since `Microsoft.Identity.Client.Desktop` is only supported in Windows, did we choose to simply not move to this alternative approach?

mijpeterson:
FWIW, broker/WAM is Windows only. So I think we'd actively want to avoid doing any broker specific logic on Mac and Linux.

src/MSALWrapper.Benchmark/BrokerBenchmark.cs
```diff
@@ -87,14 +73,7 @@ private IPCAWrapper BuildPCAWrapper(ILogger logger, Guid clientId, Guid tenantId
...
 ;
-            if (useNativeBroker)
-            {
-                clientBuilder.WithBroker(new BrokerOptions(BrokerOptions.OperatingSystems.Windows));
-            }
-            else
-            {
-                clientBuilder.WithBroker();
-            }
+            clientBuilder.WithBroker(new BrokerOptions(BrokerOptions.OperatingSystems.Windows));
```
mijpeterson, line 76: Based on the PR description, it sounds like this call is obsolete/deprecated. Is that not true?

keyuxuan: oh mb i did not make it clear, starting from `4.61.0`, usage of `WithBroker` with no parameter is deprecated, it need to be used with the parameter `options`

Haard30 (approved):

src/MSALWrapper.Benchmark/BrokerBenchmark.cs
Haard30: Yes, the managed broker (old broker api where we passed no parameters) is obsolete.

keyuxuan:
> I am curious that since `Microsoft.Identity.Client.Desktop` is only supported in Windows, did we choose to simply not move to this alternative approach?

i dont think we can? because using `Microsoft.Identity.Client.Desktop` would cause [Build and Test](https://github.com/AzureAD/microsoft-authentication-cli/actions/runs/10495034169/job/29072621189) to fail for macos and ubuntu platforms

Haard30:
> I am curious that since `Microsoft.Identity.Client.Desktop` is only supported in Windows, did we choose to simply not move to this alternative approach?

Shalini, were you asking about [this](Microsoft.Identity.Client.Desktop) PR where we moved away from `Microsoft.Identity.Client.Desktop`? If so yes, we moved away from that library earlier and use the alternative approach of newer Broker API. I think we still kept managed broker usage in benchmarking project to compare execution time of both broker apis. But MSAL has decided to deprecate that old API so we can no longer use that directly for benchmarking.

Haard30:
> > I am curious that since `Microsoft.Identity.Client.Desktop` is only supported in Windows, did we choose to simply not move to this alternative approach?
> 
> Shalini, were you asking about [this](https://github.com/AzureAD/microsoft-authentication-cli/pull/323/files) PR where we moved away from `Microsoft.Identity.Client.Desktop`? If so yes, we moved away from that library earlier and use the alternative approach of newer Broker API. I think we still kept managed broker usage in benchmarking project to compare execution time of both broker apis. But MSAL has decided to deprecate that old API so we can no longer use that directly for benchmarking.

Keyu pointed out that we didn't use `Microsoft.Identity.Client.Desktop` in that PR but we had PR open where we were planning to use it but that didn't merge 😅
https://github.com/AzureAD/microsoft-authentication-cli/pull/82

src/MSALWrapper.Benchmark/BrokerBenchmark.cs
mijpeterson: Ah understood! And the only place we used parameterless method was in the benchmark class?

keyuxuan: yes thats the only place!

PR #401, opened 2024-08-23
Update dotnet path in mac publish script

keyuxuan:
The command `./bin/mac/publish` would throw error when i run it at the top level of the folder, for example `/Users/keyuxuan/microsoft-authentication-cli`. The reason is that its using the globally installed `dotnet` instead of the [`dotnet`](https://github.com/AzureAD/microsoft-authentication-cli/blob/main/bin/mac/dotnet) under `bin/mac` folder.

Updating the dotnet to use the absolute path in mac publish script

### DIFF
--- a/bin/mac/publish
+++ b/bin/mac/publish
@@ -42,4 +42,4 @@ esac
 #------------------------------------------
 
 # Publish
-dotnet publish "$GIT_DIR/src/AzureAuth/AzureAuth.csproj" --self-contained true -r "$runtime" -c release -o $DIST $*
+$GIT_DIR/bin/mac/dotnet publish "$GIT_DIR/src/AzureAuth/AzureAuth.csproj" --self-contained true -r "$runtime" -c release -o $DIST $*


mvanchaa (approved):

Haard30 (approved):

rewrlution:
Out of curiosity, do we always install `dotnet` in the `bin/mac` folder in the build machines?

mvanchaa:
> > Out of curiosity, do we always install `dotnet` in the `bin/mac` folder in the build machines?
> 
> good question! based on my understanding, and please correct me if i am wrong:
> 
> i dont think we are really installing `dotnet` in the `bin/mac` folder here, the purpose of [`dotnet`](https://github.com/AzureAD/microsoft-authentication-cli/blob/main/bin/mac/dotnet) script is that it tries to populate an environment variable ADO_TOKEN with an access token using azureauth. And This ADO_TOKEN is being used while doing dotnet publishing.
> 
> So before this PR's change, in the `publish` script, its calling `dotnet publish` directly without that obtaining the token step, thats why there were some errors (which seem to be authentication errors/issues)

To add to what Keyu mentioned, we don't install dotnet in any of the [publish](https://github.com/AzureAD/microsoft-authentication-cli/blob/main/bin/mac/publish) or [dotnet](https://github.com/AzureAD/microsoft-authentication-cli/blob/main/bin/mac/dotnet) scripts under the `bin/mac` folder. In the `dotnet` script, after warming the `ADO_TOKEN`, we invoke the globally installed `dotnet`.

rewrlution:
> > > Out of curiosity, do we always install `dotnet` in the `bin/mac` folder in the build machines?
> > 
> > 
> > good question! based on my understanding, and please correct me if i am wrong:
> > i dont think we are really installing `dotnet` in the `bin/mac` folder here, the purpose of [`dotnet`](https://github.com/AzureAD/microsoft-authentication-cli/blob/main/bin/mac/dotnet) script is that it tries to populate an environment variable ADO_TOKEN with an access token using azureauth. And This ADO_TOKEN is being used while doing dotnet publishing.
> > So before this PR's change, in the `publish` script, its calling `dotnet publish` directly without that obtaining the token step, thats why there were some errors (which seem to be authentication errors/issues)
> 
> To add to what Keyu mentioned, we don't install dotnet in any of the [publish](https://github.com/AzureAD/microsoft-authentication-cli/blob/main/bin/mac/publish) or [dotnet](https://github.com/AzureAD/microsoft-authentication-cli/blob/main/bin/mac/dotnet) scripts under the `bin/mac` folder. In the `dotnet` script, after warming the `ADO_TOKEN`, we invoke the globally installed `dotnet`.

I see. that makes sense! approved!

rewrlution (approved):
Ship it!

PR #402, opened 2024-08-26
Bump System.Formats.Asn1 from version 6.0.0 to 6.0.1

mijpeterson:
System.Formats.Asn1 is a transitive dependency of Microsoft.VisualStudio.Services.Client in the AdoPat project. Unfortunately, the latest stable version of Microsoft.VisualStudio.Services.Client does not upgrade System.Formats.Asn1 to a patched version. As such, I've manually pinned the version of System.Formats.Asn1 to 6.0.1

I've tested generating a PAT and OAuth access token using the "ado pat" and "ado token" subcommands on my Windows machine. They both work as expected.

### DIFF
--- a/src/AdoPat/AdoPat.csproj
+++ b/src/AdoPat/AdoPat.csproj
@@ -14,6 +14,8 @@
     <PackageReference Include="Microsoft.Identity.Client.Extensions.Msal" Version="4.61.3" />
     <PackageReference Include="Microsoft.VisualStudio.Services.Client" Version="19.239.0-preview" />
     <PackageReference Include="System.Data.SqlClient" Version="4.8.6" />
+      
+    <!-- Transitive dependency of Microsoft.VisualStudio.Services.Client temporarily pinned for security reasons. -->
     <PackageReference Include="System.Formats.Asn1" Version="6.0.1" />
   </ItemGroup>
 
--- a/src/AdoPat/AdoPat.csproj
+++ b/src/AdoPat/AdoPat.csproj
@@ -6,7 +6,7 @@
     <RootNamespace>Microsoft.Authentication.AdoPat</RootNamespace>
   </PropertyGroup>
 
-    <ItemGroup>
+  <ItemGroup>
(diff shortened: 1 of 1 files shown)


src/AdoPat/AdoPat.csproj
```diff
@@ -6,14 +6,15 @@
...
-    <ItemGroup>
+  <ItemGroup>
     <!-- Stylecop required items -->
     <AdditionalFiles Include="..\stylecop\stylecop.json" Link="stylecop.json" />
     <Compile Include="..\stylecop\GlobalSuppressions.cs" Link="GlobalSuppressions.cs" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.1.4" />
     <PackageReference Include="Microsoft.Identity.Client.Extensions.Msal" Version="4.61.3" />
     <PackageReference Include="Microsoft.VisualStudio.Services.Client" Version="19.239.0-preview" />
     <PackageReference Include="System.Data.SqlClient" Version="4.8.6" />
+    <PackageReference Include="System.Formats.Asn1" Version="6.0.1" />
```
reillysiemens, line 17: What is `Microsoft.VisualStudio.Services.Client` being used for? Is it a test dependency? If so, is it critical that we pin this transitive dependency? If not, could we include a note here so that it doesn't accidentally stay pinned longer than it needs to be?

reillysiemens: Nvm, we definitely need this package at runtime. It's not a test dependency.

Could we get a comment in here like this?
```xml
<!-- Transitive dependency of Microsoft.VisualStudio.Services.Client temporarily pinned for security reasons. -->
```

mijpeterson: We certainly can! That's a good callout.

mijpeterson: This is done!

keyuxuan (approved):

PR #406, opened 2024-09-12
Adjusting IWA warning message

mijpeterson:
The IWA warning message that's printed out seems to indicate that folks can bypass 2FA/MFA challenges by using Windows Hello or a smart card. However, I believe this may no longer be true, as multiple folks have indicated that they are indeed using those mechanisms for signing in, but are still being presented with the same error. I am able to confirm this myself as well, where even when signing in with Windows Hello, I get the same MFA warning.

With the shift in the determination of when to present MFA challenges, as well as the hardening around MFA within MSFT, it is becoming clear that the guidelines around MFA challenges are ever-changing. As such, I believe providing suggestions around how to bypass these challenges will do more harm than good.

As such, this PR removes this part of the warning message.

### DIFF
--- a/src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
+++ b/src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
@@ -67,7 +67,6 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
                 && ex.Message.StartsWith("AADSTS50076", StringComparison.OrdinalIgnoreCase))
             {
                 this.logger.LogWarning("Warning: IWA failed, 2FA is required.");
-                this.logger.LogWarning("Warning: IWA can pass this requirement if you log into Windows with either a Smart Card or Windows Hello.");
                 throw;
             }
             catch (MsalClientException ex) when (ex.Message.Contains("WS-Trust endpoint not found"))
--- a/src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
+++ b/src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
@@ -66,7 +66,7 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
                 ex.Classification == UiRequiredExceptionClassification.BasicAction
                 && ex.Message.StartsWith("AADSTS50076", StringComparison.OrdinalIgnoreCase))
             {
-                this.logger.LogWarning("Warning: IWA failed, 2FA is required.");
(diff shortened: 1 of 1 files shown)


src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
```diff
@@ -67,7 +67,6 @@ protected override async Task<TokenResult> GetTokenInnerAsync()
                 && ex.Message.StartsWith("AADSTS50076", StringComparison.OrdinalIgnoreCase))
             {
                 this.logger.LogWarning("Warning: IWA failed, 2FA is required.");
```
Haard30, line 69: Should we degrade this to Debug? or remove the 2FA part as well?

shalinikhare27 (dismissed):

src/MSALWrapper/AuthFlow/IntegratedWindowsAuthentication.cs
mijpeterson: The warning message is technically still accurate, in that the MSAL exception does complain saying that an MFA challenge is required. If anything, I'd want to perhaps add some more detail to avoid confusion around people expecting an MFA prompt.

I'm also open to the idea of changing the verbosity level of this message. What's your preference?

Haard30: I would vote for changing the verbosity level as people are less concerned about IWA failing especially with the recent changes.

Haard30: My only concern with 2FA required message is if people are receiving a challenge (Authenticator app/Something else). If not, switching it to debug seems like a safer option.

mijpeterson: From my understanding, IWA will never trigger a 2FA challenge. It's simply stating that one is required, but since IWA is a purely silent flow, it will choose to silently fail instead of prompting the user.

Haard30: ah, that's correct. Well in that case 2FA required does seem a bit inactionable so lowering it to debug makes sense to me.

mijpeterson: done!

Haard30 (approved):

shalinikhare27 (approved):

mvanchaa (approved):

PR #423, opened 2025-04-26
Added support to read mode from environment variables for aad subcommands

mvanchaa:
**Why this change?**

Some users are facing intermittent issues with broker mode while invoking build commands which internally invoke AzureAuth.

If they want to switch to a different mode, they need to change the build scripts and use the changed binaries which is not ideal and is difficult for the end user. 

Hence, added this support, so that the user can set a mode that's not failing while we investigate the problem with the default mode.

### DIFF
--- a/src/AzureAuth.Test/CommandAadTest.cs
+++ b/src/AzureAuth.Test/CommandAadTest.cs
@@ -497,6 +497,19 @@ invalid_key = ""this is not a valid alias key""
             this.logTarget.Logs.Should().ContainMatch($"Invalid value specified for environment variable {EnvVars.AuthMode}*");
         }
 
+        [Test]
+        public void TestEvaluateOptionsWithAuthModeFromEmptyEnvVars()
+        {
+            CommandAad subject = this.serviceProvider.GetService<CommandAad>();
+            subject.Resource = "f0e8d801-3a50-48fd-b2da-6476d6e832a2";
+            subject.Client = "e19f71ed-3b14-448d-9346-9eff9753646b";
+            subject.Tenant = "9f6227ee-3d14-473e-8bed-1281171ef8c9";
+
+            this.envMock.Setup(env => env.Get(EnvVars.AuthMode)).Returns("");
+            subject.EvaluateOptions().Should().BeTrue();
+            subject.AuthModes.Should().Contain(new[] { AuthMode.Default });
+        }
+
         /// <summary>
         /// The root path.
         /// </summary>
--- a/src/AzureAuth.Test/CommandAadTest.cs
+++ b/src/AzureAuth.Test/CommandAadTest.cs
@@ -236,6 +236,7 @@ invalid_key = ""this is not a valid alias key""
 
             // Specify config via env var
             this.envMock.Setup(e => e.Get("AZUREAUTH_CONFIG")).Returns(configFile);
+            this.envMock.Setup(env => env.Get(It.Is<string>(key => key != "AZUREAUTH_CONFIG"))).Returns<string>(key => null);
 
             // Specify a client override on the command line.
             subject.Client = clientOverride;
--- a/src/AzureAuth.Test/CommandAadTest.cs
+++ b/src/AzureAuth.Test/CommandAadTest.cs
@@ -444,6 +444,59 @@ invalid_key = ""this is not a valid alias key""
             subject.TokenFetcherOptions.Should().BeEquivalentTo(expected);
         }
 
+        [Test]
+        public void TestEvaluateOptionsWithAuthModeFromCommandLineOptions()
+        {
+            CommandAad subject = this.serviceProvider.GetService<CommandAad>();
+            subject.Resource = "f0e8d801-3a50-48fd-b2da-6476d6e832a2";
+            subject.Client = "e19f71ed-3b14-448d-9346-9eff9753646b";
+            subject.Tenant = "9f6227ee-3d14-473e-8bed-1281171ef8c9";
+            subject.AuthModes = new List<AuthMode>() { AuthMode.DeviceCode };
+
+            this.envMock.Setup(env => env.Get(EnvVars.AuthMode)).Returns("Web,DeviceCode");
+            subject.EvaluateOptions().Should().BeTrue();
+            subject.AuthModes.Should().Contain(new[] { AuthMode.DeviceCode });
+        }
+
+        [Test]
+        public void TestEvaluateOptionsWithAuthModeFromEnvVar()
+        {
+            CommandAad subject = this.serviceProvider.GetService<CommandAad>();
+            subject.Resource = "f0e8d801-3a50-48fd-b2da-6476d6e832a2";
+            subject.Client = "e19f71ed-3b14-448d-9346-9eff9753646b";
+            subject.Tenant = "9f6227ee-3d14-473e-8bed-1281171ef8c9";
+
+            this.envMock.Setup(env => env.Get("AZUREAUTH_MODE")).Returns("Web,DeviceCode");
+            subject.EvaluateOptions().Should().BeTrue();
+            subject.AuthModes.Should().Contain(new[] { AuthMode.Web, AuthMode.DeviceCode });
+        }
+
+        [Test]
+        public void TestEvaluateOptionsWithNoAuthModeInEnvVarOrOptions()
+        {
+            CommandAad subject = this.serviceProvider.GetService<CommandAad>();
+            subject.Resource = "f0e8d801-3a50-48fd-b2da-6476d6e832a2";
+            subject.Client = "e19f71ed-3b14-448d-9346-9eff9753646b";
+            subject.Tenant = "9f6227ee-3d14-473e-8bed-1281171ef8c9";
+
+            this.envMock.Setup(env => env.Get(It.IsAny<string>())).Returns((string)null);
+            subject.EvaluateOptions().Should().BeTrue();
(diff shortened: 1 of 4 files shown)


keyuxuan (dismissed):

src/AzureAuth/Commands/CommandAad.cs
```diff
@@ -405,5 +418,34 @@ private int GetToken(IPublicClientAuth publicClientAuth)
...
+        public bool TrySetAuthModeFromEnvOrDefault()
+        {
+            var authModesFromEnv = this.env.Get(EnvVars.AuthMode);
+            if (string.IsNullOrEmpty(authModesFromEnv))
+            {
+                this.AuthModes = new[] { AuthMode.Default };
+            }
+
+            var result = new List<AuthMode>();
+            foreach(var val in authModesFromEnv.Split(','))
```
mijpeterson, line 435: Are commas or semi-colons more typical for env variable separation? I think the Path variable separates multiple values with semi-colons.

varshini-elayappen:
@mvanchaa would it be useful to measure how often the default auth mode isn't used? If the number is high, it might help us debug the circumstances in which the default auth mode isn't preferred.

src/AzureAuth/Commands/CommandAad.cs
mvanchaa: oh yes, I thought about this. But I went with the same format that option --mode allows which is comma-separated, so that the input format will be consistent.

src/AzureAuth/Commands/CommandAad.cs
```diff
@@ -60,12 +60,13 @@ public class CommandAad
         /// The help text for the <see cref="ModeOption"/> option.
         /// </summary>
 #if PlatformWindows
-        public const string AuthModeHelperText = @"Authentication mode. Repeated invocations allowed.
-[default: broker, then web]
-[possible values: all, iwa, broker, web, devicecode]";
+        public const string AuthModeHelperText = $"Authentication mode. Repeated invocations allowed.\n" +
```
rewrlution, line 63: From [this stackoverflow post](https://stackoverflow.com/questions/33236592/multiline-c-sharp-interpolated-string-literal):

You can combine string interpolation `$` and verbatim literal `@` together so that you don't need the `\n` newline string.

src/AzureAuth.Test/CommandAadTest.cs
```diff
@@ -444,6 +444,59 @@ public void TestEvaluateOptionsWithoutAliasValidCommandLineOptions()
...
+        [Test]
+        public void TestEvaluateOptionsWithAuthModeFromCommandLineOptions()
+        {
+            CommandAad subject = this.serviceProvider.GetService<CommandAad>();
+            subject.Resource = "f0e8d801-3a50-48fd-b2da-6476d6e832a2";
+            subject.Client = "e19f71ed-3b14-448d-9346-9eff9753646b";
+            subject.Tenant = "9f6227ee-3d14-473e-8bed-1281171ef8c9";
+            subject.AuthModes = new List<AuthMode>() { AuthMode.DeviceCode };
+
+            this.envMock.Setup(env => env.Get(EnvVars.AuthMode)).Returns("Web,DeviceCode");
```
rewrlution, line 456: question, in this test, we returns `Web` and `DeviceCode` mode from the environment variable.

and in line `458`, the final `AuthMode` is `DeviceCode`?

src/AzureAuth/Commands/CommandAad.cs
```diff
@@ -179,7 +186,7 @@ public CommandAad(CommandExecuteEventData eventData, ITelemetryService telemetry
         /// Gets or sets the auth modes.
         /// </summary>
         [Option(ModeOption, AuthModeHelperText, CommandOptionType.MultipleValue)]
-        public IEnumerable<AuthMode> AuthModes { get; set; } = new[] { AuthMode.Default };
+        public IEnumerable<AuthMode> AuthModes { get; set; }
```
rewrlution, line 189: Can you help me understand how the new logic works?

Originally, I was thinking to keep this line, i.e. the default AuthMode value.

Then we can try to parse the value from the environment variable and cli arguments, and override the values according to the precedence.

```ts
// default values
const config = { mode: broker }

// env variables
if (env.mode) mode = env.mode

// cli arguments
const args = argv
if (args.mode) mode = args.mode
```

Looks like we are now removing the default value?

rewrlution (commented):
I left some comments to this PR.

src/AzureAuth.Test/CommandAadTest.cs
mvanchaa: Yes because on line 454 - we specify the mode through options --mode. So that takes precedence over env variables.

src/AzureAuth/Commands/CommandAad.cs
mvanchaa: No, we are not removing the default value. 

```c#
[Option(ModeOption, AuthModeHelperText, CommandOptionType.MultipleValue)]
public IEnumerable<AuthMode> AuthModes { get; set; }
```
This is going to read auth mode from command line option --mode, if specified.
Earlier, if that's not specified, it will be `AuthMode.Default`. 

Now, on line 282, I check if command line option is given. If not, I will set it from env variable. See method `TrySetAuthModeFromEnvOrDefault` on line 426. If env variable is not specified, I set it to default. Else, set it to the modes specified in the env variable.

Let me know if this is not clear.

Manuha:
> @mvanchaa would it be useful to measure how often the default auth mode isn't used? If the number is high, it might help us debug the circumstances in which the default auth mode isn't preferred.

Done.

src/AzureAuth/Commands/CommandAad.cs
mijpeterson: I think that's reasonable :) I don't know much about env variable conventions, but from searching around, it sounds like environment variables that map to directories/paths should use semi-colon delimiters. However, for other environment variables, it appears there isn't really a convention.

So I think the way you have it is great, and your justification makes perfect sense.

src/AzureAuth.Test/CommandAadTest.cs
rewrlution: okay, so the scenario we are testing here is that:

1. env var: web and device code mode
2. cli arg: device code

end result:
device code

mvanchaa: Yes that's right.

src/AzureAuth/Commands/CommandAad.cs
```diff
@@ -60,12 +60,13 @@ public class CommandAad
         /// The help text for the <see cref="ModeOption"/> option.
         /// </summary>
 #if PlatformWindows
-        public const string AuthModeHelperText = @"Authentication mode. Repeated invocations allowed.
-[default: broker, then web]
-[possible values: all, iwa, broker, web, devicecode]";
+        public const string AuthModeHelperText = $@"Authentication mode. Repeated invocations allowed.
+            [default: broker, then web]
```
rewrlution, line 64: I think you might need to remove the trailing spaces.

Otherwise, it might looks like:

```txt
line 1
      line 2
      line 3
```

in the terminal.

mvanchaa: Good catch. But somehow it's not showing the trailing space when I ran it. I will update that anyway

mijpeterson:
More of a general comment for us to think about (and not something to take action on for this PR): I think we're starting to enter the territory where I believe AzureAuth is getting more complex than it should be.

This PR, as well as at least one other we've made in the past, are adding changes to address issues that I believe ultimately stem from a problem with our broker auth flow. I think applying these workarounds to unblock customers is an absolutely valid path forward, but I also think we should start to assess why broker/WAM is consistently failing to auth silently for many of our customers.

[This doc](https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/desktop-mobile/integrated-windows-authentication) even calls out that WAM is meant to be an even more effective way of retrieving tokens silently (over IWA which is not supported for Entra-backed "managed accounts"), and yet we're finding that the local MSAL cache and IWA (for service accounts) continue to work better for many of our customers. And the reality is that these are starting to be flagged as "non-compliant".

I think it'd be a good idea for us to start looking at fixing our broker flow, and re-standardize our customers' auth flow towards the "compliant" solution.

rewrlution (dismissed):
Ship it!

mijpeterson (dismissed):

mvanchaa:
> More of a general comment for us to think about (and not something to take action on for this PR): I think we're starting to enter the territory where I believe AzureAuth is getting more complex than it should be.
> 
> This PR, as well as at least one other we've made in the past, are adding changes to address issues that I believe ultimately stem from a problem with our broker auth flow. I think applying these workarounds to unblock customers is an absolutely valid path forward, but I also think we should start to assess why broker/WAM is consistently failing to auth silently for many of our customers.
> 
> [This doc](https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/desktop-mobile/integrated-windows-authentication) even calls out that WAM is meant to be an even more effective way of retrieving tokens silently (over IWA which is not supported for Entra-backed "managed accounts"), and yet we're finding that the local MSAL cache and IWA (for service accounts) continue to work better for many of our customers. And the reality is that these flows are starting to be flagged as "non-compliant".
> 
> I think it'd be a good idea for us to start looking at fixing our broker flow, and re-standardize our customers' auth flow towards the "compliant" solution.

One hundred percent agreed!

rewrlution (approved):

PR #424, opened 2025-05-01
Extend support to read auth mode from environment variables for ado subcommands.

mvanchaa:
This is continuation of this PR #423 and add support to read auth mode from env vars for ado subcommands.


src/AzureAuth/Commands/Ado/CommandToken.cs
```diff
@@ -98,6 +98,13 @@ public int OnExecute(ILogger<CommandToken> logger, IEnv env, ITelemetryService t
                 return 0;
             }
 
+            // If command line options for mode are not specified, then use the environment variables.
+            this.AuthModes ??= AuthModeHelper.ReadAuthModeFromEnvOrSetDefault(env, eventData, logger);
+            if (!this.AuthModes.Any())
+            {
+                return 1;
```
mijpeterson, line 105: Should we add a logger call here, similar to what you had before?

ie:` this.logger.LogError($"Invalid value specified for environment variable {EnvVars.AuthMode}. Allowed values are: {AuthModeAllowedValues}");`

src/AzureAuth/AuthModeHelper.cs
```diff
@@ -0,0 +1,54 @@
...
+    public static class AuthModeHelper
+    {
+        /// <summary>
+        /// Get the auth modes from the environment or set the default.
+        /// </summary>
+        /// <param name="env">The <see cref="IEnv"/> to use.</param>
+        /// <param name="eventData">Event data to add the auth mode to.</param>
+        /// <param name="logger">The <see cref="ILogger"/> to use.</param>
+        /// <returns>AuthModes.</returns>
+        public static IEnumerable<AuthMode> ReadAuthModeFromEnvOrSetDefault(IEnv env, EventData eventData, ILogger logger)
```
mijpeterson, line 26: Rather than creating a helper and adding a single method here, should we consider adding this to IEnvExtensions instead?

Helper classes always don't fit quite right in OOP, because they don't really represent an entity.

mvanchaa: I was debating between these two options, I will modify it :)

src/AzureAuth/IEnvExtensions.cs
```diff
@@ -22,5 +26,38 @@ public static bool InteractiveAuthDisabled(this IEnv env)
...
                 string.Equals(CorextPositiveValue, env.Get(EnvVars.CorextNonInteractive));
         }
+
+        /// <summary>
+        /// Get the auth modes from the environment or set the default.
+        /// </summary>
+        /// <param name="env">The <see cref="IEnv"/> to use.</param>
+        /// <param name="eventData">Event data to add the auth mode to.</param>
+        /// <returns>AuthModes.</returns>
+        public static IEnumerable<AuthMode> ReadAuthModeFromEnvOrSetDefault(this IEnv env, EventData eventData)
```
mijpeterson, line 36: nit: Is there an alternative to passing in eventData here? It does seem a bit out of place for an environment extension method, but I can understand the desire to centralize telemetry.

mijpeterson: Also, feel free to resolve this comment whenever since I won't be available to review next week. Just wanted to raise a thought 😄

src/AzureAuth/IEnvExtensions.cs
```diff
@@ -22,5 +26,38 @@ public static bool InteractiveAuthDisabled(this IEnv env)
...
+            var result = new List<AuthMode>();
+            foreach (var val in authModesFromEnv.Split(','))
+            {
+                if (Enum.TryParse<AuthMode>(val, ignoreCase: true, out var mode))
+                {
+                    result.Add(mode);
+                }
+                else
+                {
+                    return new List<AuthMode>();
```
rewrlution, line 55: I am wondering what happens when the `TryParse()` returns an error.

Does it mean that we will return an empty list of `AuthMode`?

In this case, should we fall back to the `Auth.Default`?

src/AzureAuth/IEnvExtensions.cs
mvanchaa: That sounds right. I just found a way to put this env variables under the existing env variables we collect for telemetry here https://github.com/AzureAD/microsoft-authentication-cli/blob/a151ef0a7648b89f47d58ced2eaf08cccf77e9c2/src/AzureAuth/Program.cs#L65

mvanchaa: Done

src/AzureAuth/IEnvExtensions.cs
rewrlution: Okay, that makes sense. Thanks for the explanation.

rewrlution (approved):

rewrlution (approved):

keyuxuan (approved):

PR #431, opened 2025-06-11
Attempt to remediate test failures

mijpeterson:
When I made [this PR](https://github.com/AzureAD/microsoft-authentication-cli/pull/430), I realized that the test workflows were failing due to auth errors even though we previously had spent substantial effort to have them dynamically pull in access tokens to be able to connect to the Office feed.

It seems that with the ongoing security hardening, however, that having these actions performed via GitHub workflows is becoming less and less practical. On top of all of that, our ADO instance is moving to an entirely different tenant altogether that will prevent GitHub from even being able to generate tokens via FIC.

With all the writing on the wall, I decided to move our automated workflows over to Azure pipelines since this seems to be the only way to future proof ourselves against further hardening.


.github/actions/get-ado-token/action.yml
```diff
@@ -1,34 +0,0 @@
-name: Get Azure DevOps Access Token
```
mijpeterson, line 1: This file ends up no longer being needed because the pipeline itself has access to the necessary ADO feeds.

.github/workflows/azure-devops-pipelines-cg.yml
```diff
@@ -1,47 +0,0 @@
-# This pipeline will be triggered when either main branch is pushed or 2AM on workdays.
```
mijpeterson, line 1: Replaced by ci.yml

.github/workflows/dotnet-test.yml
```diff
@@ -1,85 +0,0 @@
-name: Build and Test
```
mijpeterson, line 1: Replaced by pr.yml

nuget.config
```diff
@@ -2,16 +2,6 @@
 <configuration>
```
mijpeterson, line 2: Similar to no longer needing the action to generate access tokens, this config should no longer need a token to be provided since the CI environment that previously required this token now has access to the artifact feed.

.azuredevops/release.yml
```diff
@@ -113,14 +113,12 @@ extends:
             inputs:
               command: restore
               feedsToUse: select
-              vstsFeed: $(VSTS_FEED_ID)
```
mijpeterson, line 116: I discovered you can provide the name rather than an ID. This is makes things much simpler since we don't need to hide away an internal ID anymore.

.azuredevops/release.yml
```diff
@@ -113,14 +113,12 @@ extends:
...
               command: restore
               feedsToUse: select
-              vstsFeed: $(VSTS_FEED_ID)
+              vstsFeed: Office
               includeNuGetOrg: false
               arguments: --runtime ${{ config.runtime }}
           # 1ES PT requires explicit build task for Roslyn analysis. Auto-injected Roslyn task will use build logs from this build.
           - task: DotNetCoreCLI@2
             displayName: Build projects
-            env:
```
mijpeterson, line 122: Since the nuget.config file no longer has this ADO_TOKEN token, we can remove this unnecessary environment variable.

mijpeterson (commented):
Some helpful notes for people who might be reviewing.

Haard30 (approved):

aschlesener (approved):

PR #452, opened 2026-04-08
feat: macOS brokered authentication support

dggsax:
## Summary

Adds macOS brokered authentication support via MSAL.NET, NativeInterop, and the macOS Enterprise SSO Extension (deployed through Company Portal).

### What this does

- **Broker auth on macOS** — opt-in via `--mode broker`. Supports interactive broker (first run) and silent broker (subsequent runs via MSAL cache).
- **Company Portal version gating** — requires CP >= 5.2603.0. Warning logged with CP path if version is insufficient, then falls through to next auth flow.
- **Main thread dispatch** — broker interactive calls are dispatched to the main thread via `MacMainThreadScheduler`, as required by MSAL on macOS.
- **Browser fallback** — when both `--mode broker --mode web` are specified, falls back to browser auth if broker fails.
- **NuGet upgrades** — MSAL 4.65.0 → 4.83.1, added NativeInterop v0.20.3.

### Why broker is opt-in on macOS (not the default)

On Windows, the default auth mode is `Broker | Web` — broker is tried first, then web as fallback. **On macOS, the default remains `Web` only, with broker opt-in via `--mode broker`.**

This is intentional because apps configured with broker-required Conditional Access policies (e.g., token protection / compliant device, error 530084) will **hang indefinitely** if web auth is attempted: the browser shows an error page about needing a compliant device but never redirects back to `localhost`, so MSAL's HTTP listener waits forever. Making broker the default would mean that if Company Portal is unavailable and the flow falls through to web, the CLI hangs with no way to recover — a worse experience than a clear error.

Since `--mode` is repeatable (`--mode broker --mode web` tries both in order), users and tooling can explicitly opt into the desired combination. This matches the existing pattern where `--mode broker` on Windows Server silently skips broker (no WAM) and falls through.

### Key design decisions

1. **Broker is opt-in on macOS** (`--mode broker`) — see above for rationale.
2. **Broker silently skipped when unavailable** — matches the existing Windows Server pattern. Warning logged, then falls through to next flow (Web, DeviceCode, etc.) instead of throwing.
3. **No `OperatingSystemAccount` on macOS** — this Windows-only concept does not work on macOS. On macOS, `TryToGetCachedAccountAsync` resolves accounts from the MSAL Keychain cache for silent auth. If multiple accounts are cached (ambiguous), interactive auth is triggered — the safe default for a CLI tool.
4. **SSO Extension check removed** — `app-sso -l` returned empty on test devices yet broker worked fine. The check and associated code were removed entirely as unnecessary.

### Testing

- ✅ All 364 unit tests pass (95 AdoPat + 109 AzureAuth + 160 MSALWrapper)
- ✅ Functional end-to-end testing verified with Work IQ 3P Graph app registration
  - Interactive broker → account picker → token acquired in ~3.6s
  - Silent broker → cached account → token acquired in ~1s
- ✅ New unit tests: `BrokerMacOSTest` (5 tests)
- ✅ Functional test script at `bin/mac/test-macos-broker.sh` (6 tests)

### Files changed

**New files:**
- `src/MSALWrapper.Test/AuthFlow/BrokerMacOSTest.cs` — 5 macOS broker unit tests
- `bin/mac/test-macos-broker.sh` — functional test script

**Modified files:**
- `src/MSALWrapper/MSALWrapper.csproj` — NuGet upgrades
- `src/MSALWrapper/AuthFlow/Broker.cs` — major rewrite for cross-platform
- `src/MSALWrapper/AuthFlow/AuthFlowFactory.cs` — broker availability gating with fallthrough
- `src/MSALWrapper/PlatformUtils.cs` — CP version detection, macOS checks
- `src/MSALWrapper/AuthMode.cs` — broker opt-in on non-Windows (`Default = Web`)
- `src/MSALWrapper/Constants.cs` — macOS broker redirect URI
- `src/AzureAuth/Program.cs` — main thread scheduler pattern
- `src/AzureAuth/AuthModeExtensions.cs` — preserve broker for silent auth in non-interactive mode
- `src/AzureAuth/Commands/CommandAad.cs` — `--mode broker` option, fixed help text (`[default: web]` on non-Windows)
- `docs/usage.md` — macOS broker prerequisites and redirect URI config
- `README.md` — feature matrix updated

### Prerequisites for broker auth

- Company Portal >= 5.2603.0 installed
- Device MDM-compliant
- App registration configured with redirect URI: `msauth.com.msauth.unsignedapp://auth`


bin/mac/swap-cp.sh
ChristopherJMiller, line 1: can this script have a comment at the top describing what it's for testing wise and why it's needed?

bin/mac/test-macos-broker.sh
```diff
@@ -0,0 +1,270 @@
...
+    echo "CP < 2603 or not installed — expecting clear error about Company Portal"
+    echo "Expect: InvalidOperationException with CP version/path info"
+    EXPECTED_EXIT=1
+fi
+run_test "Broker-only (opt-in)" "$EXPECTED_EXIT" \
+    aad --client "$CLIENT" --tenant "$TENANT" \
+    --resource "$RESOURCE" \
+    --mode broker --output json --verbosity "$VERBOSITY"
+
+# ── Test 2: Broker + web combined — COMMENTED OUT ─────────────
```
ChristopherJMiller: for a checked in testing suite is it worth keeping this commented out test?

src/AzureAuth.Test/AuthModeExtensionsTest.cs
```diff
@@ -58,13 +58,15 @@ public void Filterinteraction_Interactive_Auth_Disabled(string envVar)
...
+        [Test]
         public void CombinedAuthMode_Allowed()
         {
             // Arrange
             this.envMock.Setup(e => e.Get(EnvVars.NoUser)).Returns(string.Empty);
             this.envMock.Setup(e => e.Get("Corext_NonInteractive")).Returns(string.Empty);
 
-            var subject = new[] { AuthMode.Web, AuthMode.DeviceCode };
+            // Default on macOS is Web only (broker is opt-in).
+            var subject = new[] { AuthMode.Web };
```
ChristopherJMiller, line 70: should this be an os specific test?

src/MSALWrapper/PlatformUtils.cs
```diff
@@ -39,6 +45,179 @@ public bool IsWindows()
...
+                    this.logger.LogTrace($"Company Portal path: {CompanyPortalAppPath}");
+                    this.logger.LogTrace($"Company Portal version parts: major={parts[0]}, release={parts[1]}{(parts.Length >= 3 ? $", build={parts[2]}" : string.Empty)}");
+
+                    if (!meetsMinimum)
+                    {
+                        this.logger.LogWarning($"macOS broker unavailable: Company Portal {output} (at {CompanyPortalAppPath}) is below minimum required release {MinimumCPRelease}.");
+                        return false;
+                    }
+
+                    // TODO: Re-enable once we confirm SSO Extension registration is required.
```
ChristopherJMiller: we may want to track this in an issue so it's not forgotten about

ChristopherJMiller (changes requested):

src/AzureAuth/Commands/CommandAad.cs
```diff
@@ -65,7 +65,7 @@ public class CommandAad
 [possible values: {AuthModeAllowedValues}]";
 #else
         public const string AuthModeHelperText = $@"Authentication mode. Repeated invocations allowed
-[default: web]
+[default: broker, then web]
```
siddharthpaudwal: just like windows it has condition - 
#if PlatformWindows
[default: broker, then web]

can we add the same for MAC ? 
broker may impact linux machines

src/MSALWrapper/PlatformUtils.cs
dggsax: Yeah, I don't believe this is required, but we can test.

dggsax: Narrator: was not necessary ,deleted the TODO and the underlying code.

dggsax:
Re: [this comment](https://github.com/AzureAD/microsoft-authentication-cli/pull/452#discussion_r3064016046) — good callout. Linux is safe here: `AuthFlowFactory` (line 63-84) only adds the Broker flow when `IsWindows10Or11()` or `IsMacOS() && IsMacOSBrokerAvailable()`. On Linux, `--mode broker` is silently skipped and falls through to the next flow (web, devicecode) — same behavior as Windows Server where broker is unavailable.

The help text already shows `[default: web]` on the `#else` branch (both macOS and Linux), and broker is opt-in only. A Linux user passing `--mode broker` would just get a no-op skip, which is consistent with the existing pattern.

bin/mac/swap-cp.sh
dggsax: Resolved — `swap-cp.sh` has been deleted from the PR entirely.

bin/mac/test-macos-broker.sh
dggsax: Resolved — commented-out web mode tests have been removed. The test script now has 6 active tests only.

src/AzureAuth.Test/AuthModeExtensionsTest.cs
dggsax: Yes — added `[Platform("MacOsX,Linux")]` attributes to the non-Windows tests. These tests are in the `#else` conditional compilation block (which covers both macOS and Linux), and the `[Platform]` attribute ensures NUnit only runs them on those OSes.

ChristopherJMiller (approved):
lgtm

siddharthpaudwal (approved):

PR #453, opened 2026-04-23
feat: macOS brokered authentication support

dggsax:
* feat: add macOS brokered authentication support

Add support for macOS Enterprise SSO Extension brokered authentication, bringing feature parity with Windows WAM broker support.

Key changes:
- Upgrade MSAL 4.65.0 → 4.83.1, add NativeInterop v0.20.3
- Extend Broker auth flow with macOS-specific PCA config, account resolution, and browser fallback
- Add DefaultAccountStore to persist account username for silent auth (OperatingSystemAccount not supported on macOS)
- Add MacMainThreadScheduler message loop in Program.cs
- Extend AuthMode enum and CLI parsing to support broker on macOS
- Update docs with macOS broker prerequisites and redirect URI config

* feat: add Company Portal version check before broker auth

Skip macOS broker auth if Company Portal is not installed or is below version 2603 (which added redirect_uri validation fix for unsigned apps). Falls back to web auth transparently in those cases.

- Add IsMacOSBrokerAvailable() to IPlatformUtils/PlatformUtils
- AuthFlowFactory uses IsMacOSBrokerAvailable() as gatekeeper
- Broker.cs still uses IsMacOS() for runtime behavior (fallback, persist)
- Reads CP version from Info.plist via 'defaults read'

* feat: make broker opt-in on macOS, error on missing CP

On macOS, broker is no longer included in default auth modes. Users must explicitly pass '--mode broker' to use it. If broker is requested but Company Portal >= 5.2603.0 is not installed, AuthFlowFactory throws a clear InvalidOperationException instead of silently falling through to web auth (which hangs for apps with broker-required CA policies like token protection).

- AuthMode.Default on non-Windows: Broker|Web -> Web
- AuthFlowFactory: explicit error when broker requested but CP unavailable
- New test: BrokerRequested_Mac_CP_Unavailable_Throws

* improve: CP diagnostics logging and error messages

- Include CP path in broker unavailable error message
- Add trace-level logging: CP path, raw version output, stderr, parsed version parts (major/release/build)
- Expose CompanyPortalAppPath as public const for error messages
- Update test script to reflect broker-opt-in behavior: new tests for broker+web combined, trace diagnostics, reordered

* fix: dispatch broker interactive calls to main thread on macOS

The macOS broker requires AcquireTokenInteractive to run on the main thread. Program.cs starts the MacMainThreadScheduler message loop on main and dispatches CLI work to Task.Run, but the broker interactive calls were still executing on the background thread.

Now GetTokenInteractive/WithClaims dispatch through MacMainThreadScheduler.RunOnMainThreadAsync when running on macOS with the scheduler active (IsRunning check prevents deadlock in tests).

Also: improved CP diagnostics trace logging and included CP path in the broker unavailable error message.

* feat: add SSO Extension registration pre-flight check

Check 'app-sso -l' for registered Enterprise SSO Extensions before attempting broker auth. If no extensions are registered (MDM profile not applied), gives a clear error instead of a cryptic broker failure.

Designed for easy revert:
- Set AZUREAUTH_SKIP_SSO_CHECK=1 to bypass the check entirely
- If app-sso fails or isn't available, assumes broker may work (non-fatal)
- The check is a single method call in CheckMacOSBrokerAvailable()

* chore: move macOS test scripts to bin/mac/

Move test-macos-broker.sh and swap-cp.sh from repo root to bin/mac/ alongside existing macOS build scripts. Update REPO_ROOT to resolve two levels up from the new location.

* improve: configurable verbosity in test script, disable SSO check

- Add AZUREAUTH_TEST_VERBOSITY env var (default: debug) to control log level across all tests (Test 3 always uses trace)
- Comment out SSO Extension check (confirmed unnecessary for broker)

* docs: add usage examples to test script header

* fix: broker falls through to next auth flow when unavailable on macOS

Match the existing Windows pattern where broker is silently skipped on unsupported platforms (e.g., Windows Server). On macOS, if broker is requested but Company Portal is insufficient, log a warning and continue to the next flow (Web, DeviceCode, etc.) instead of throwing.

This means '--mode broker --mode web' will fall through to web when CP is unavailable, and '--mode broker' alone will try CachedAuth only.

* test: comment out web mode tests, add broker re-prompt after clear

Web auth hangs for this broker-required app (token protection CA policy), so comment out Tests 2 and 5. Add cache clear + broker interactive re-prompt cycle (Tests 6-8) to verify full broker lifecycle: authenticate → clear → re-authenticate.

* fix: address PR review comments

- Delete swap-cp.sh (reviewer: no need for checked-in CP swap utility)
- Remove commented-out web mode tests from test script (reviewer: keep checked-in tests clean)
- Add [Platform] attributes to non-Windows AuthModeExtensions tests (reviewer: make OS-specificity explicit)
- Fix help text: non-Windows default is 'web' not 'broker, then web' (reviewer: broker may confuse Linux users; --mode is repeatable so --mode broker explicitly opts in)

* docs: update usage.md to reflect broker opt-in on macOS

- Broker is opt-in via --mode broker, not automatic
- Removed SSO Extension prerequisite (confirmed unnecessary)
- Added CP version requirement (>= 5.2603.0)
- Added example CLI invocations
- Clarified fallthrough behavior when broker unavailable

* Remove unused SSO Extension check code

The IsSSOExtensionRegistered method and its TODO were confirmed unnecessary during testing — broker auth works without SSO Extension registration showing in app-sso output. Removes dead code and the AZUREAUTH_SKIP_SSO_CHECK env var that was only needed for this check.

* Fix async deadlock in PersistDefaultAccount

Changed PersistDefaultAccount from sync .Wait()/.Result to async/await. The previous pattern (calling task.Wait() on an async method inside an async context) is a classic .NET deadlock risk, especially dangerous with the macOS MacMainThreadScheduler SynchronizationContext.

* Remove DefaultAccountStore — rely on MSAL cache instead

DefaultAccountStore persisted account usernames (PII) to plaintext JSON files in ~/.azureauth/ with world-readable permissions. This was only needed to disambiguate multi-account scenarios (TryToGetCachedAccountAsync returns null when >1 accounts are cached).

Without it, multi-account users get an extra broker prompt — which is the safe default for a CLI tool and avoids writing PII to disk.

Changes:
- Delete DefaultAccountStore.cs and DefaultAccountStoreTest.cs
- Simplify Broker.ResolveAccountAsync (no more persisted username lookup)
- Remove PersistDefaultAccountAsync from Broker.cs
- Simplify BrokerMacOSTest.cs (remove store setup/teardown)

---------

### DIFF
--- a/src/AzureAuth/Commands/CommandAad.cs
+++ b/src/AzureAuth/Commands/CommandAad.cs
@@ -90,7 +90,7 @@ namespace Microsoft.Authentication.AzureAuth.Commands
 #if PlatformWindows
         public const string AuthModeAllowedValues = "all, iwa, broker, web, devicecode";
 #else
-        public const string AuthModeAllowedValues = "all, web, devicecode";
+        public const string AuthModeAllowedValues = "all, broker, web, devicecode";
 #endif
 
         private const string ResourceOption = "--resource";
--- a/src/AzureAuth/Program.cs
+++ b/src/AzureAuth/Program.cs
@@ -6,6 +6,7 @@ namespace Microsoft.Authentication.AzureAuth
     using System;
     using System.Runtime.InteropServices;
     using System.Text;
+    using System.Threading.Tasks;
 
     using McMaster.Extensions.CommandLineUtils;
 
@@ -13,6 +14,7 @@ namespace Microsoft.Authentication.AzureAuth
     using Microsoft.Authentication.MSALWrapper;
     using Microsoft.Extensions.DependencyInjection;
     using Microsoft.Extensions.Logging;
+    using Microsoft.Identity.Client.Utils;
     using Microsoft.Office.Lasso;
     using Microsoft.Office.Lasso.Telemetry;
 
@@ -22,6 +24,34 @@ namespace Microsoft.Authentication.AzureAuth
     public class Program
     {
         private static void Main(string[] args)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                // macOS brokered auth requires interactive calls on the main thread.
+                // Start the MSAL message loop on main, dispatch CLI work to a background thread.
+                var scheduler = MacMainThreadScheduler.Instance();
+                var cliTask = Task.Run(() =>
+                {
+                    try
+                    {
+                        MainInner(args);
+                    }
+                    finally
+                    {
+                        scheduler.Stop();
+                    }
+                });
+
+                scheduler.StartMessageLoop();
+                cliTask.GetAwaiter().GetResult();
+            }
+            else
+            {
+                MainInner(args);
+            }
+        }
+
+        private static void MainInner(string[] args)
         {
             // Use UTF-8 output encoding.
             // This will impact the NLog Console Target as well as any other Console usage.
--- a/src/MSALWrapper/AuthFlow/AuthFlowFactory.cs
+++ b/src/MSALWrapper/AuthFlow/AuthFlowFactory.cs
@@ -41,7 +41,10 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
 
             // We skip CachedAuth if Broker is present in authMode on windows 10 or 11, since Broker 
             // already tries CachedAuth with its PCAWrapper object built using withBroker(options).
-            if (!(authMode.IsBroker() && platformUtils.IsWindows10Or11()))
+            // The same applies on macOS when the broker is available.
+            // If broker is requested but unavailable, CachedAuth is still added as a first-pass attempt.
+            bool brokerWillRun = authMode.IsBroker() && (platformUtils.IsWindows10Or11() || platformUtils.IsMacOSBrokerAvailable());
+            if (!brokerWillRun)
             {
                 flows.Add(new CachedAuth(logger, authParams, preferredDomain, pcaWrapper));
             }
@@ -53,12 +56,31 @@ namespace Microsoft.Authentication.MSALWrapper.AuthFlow
                 flows.Add(new IntegratedWindowsAuthentication(logger, authParams, preferredDomain, pcaWrapper));
             }
 
-            // This check silently fails on winserver if broker has been requested.
-            // Future: Consider making AuthMode platform aware at Runtime.
+            // Broker is silently skipped when unavailable on the current platform
+            // (e.g., Windows Server, macOS without Company Portal). The executor
+            // continues to the next flow in the list (Web, DeviceCode, etc.).
             // https://github.com/AzureAD/microsoft-authentication-cli/issues/55
-            if (authMode.IsBroker() && platformUtils.IsWindows10Or11())
+            if (authMode.IsBroker())
             {
-                flows.Add(new Broker(logger, authParams, preferredDomain: preferredDomain, pcaWrapper: pcaWrapper, promptHint: promptHint));
+                if (platformUtils.IsWindows10Or11())
+                {
+                    flows.Add(new Broker(logger, authParams, preferredDomain: preferredDomain, pcaWrapper: pcaWrapper, promptHint: promptHint, platformUtils: platformUtils));
+                }
+                else if (platformUtils.IsMacOS())
+                {
+                    if (platformUtils.IsMacOSBrokerAvailable())
+                    {
+                        flows.Add(new Broker(logger, authParams, preferredDomain: preferredDomain, pcaWrapper: pcaWrapper, promptHint: promptHint, platformUtils: platformUtils));
+                    }
+                    else
+                    {
+                        logger.LogWarning(
+                            "Broker authentication was requested but is not available on this machine. " +
+                            "macOS broker requires Company Portal version 5.2603.0 or later " +
+                            $"(checked: {PlatformUtils.CompanyPortalAppPath}). " +
+                            "Skipping broker and falling through to next auth flow.");
+                    }
+                }
             }
 
             if (authMode.IsWeb())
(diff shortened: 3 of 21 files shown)


ChristopherJMiller (dismissed):
lgtm

siddharthpaudwal (dismissed):

src/AzureAuth/Commands/CommandAad.cs
```diff
@@ -90,7 +90,7 @@ public class CommandAad
 #if PlatformWindows
         public const string AuthModeAllowedValues = "all, iwa, broker, web, devicecode";
 #else
-        public const string AuthModeAllowedValues = "all, web, devicecode";
+        public const string AuthModeAllowedValues = "all, broker, web, devicecode";
```
kyle-rader-msft, line 93: IWA should really be dropped from Windows at this point, which means finally the auth modes between mac and windows are actually the same. Should be a follow up change

src/AzureAuth/Program.cs
```diff
@@ -22,6 +24,34 @@ namespace Microsoft.Authentication.AzureAuth
     public class Program
     {
         private static void Main(string[] args)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
```
kyle-rader-msft, line 28: Is there a reason not to simply also do this on Windows?

src/MSALWrapper/AuthFlow/AuthFlowFactory.cs
```diff
@@ -53,12 +56,31 @@ public static IEnumerable<IAuthFlow> Create(
...
+                    flows.Add(new Broker(logger, authParams, preferredDomain: preferredDomain, pcaWrapper: pcaWrapper, promptHint: promptHint, platformUtils: platformUtils));
+                }
+                else if (platformUtils.IsMacOS())
+                {
+                    if (platformUtils.IsMacOSBrokerAvailable())
+                    {
+                        flows.Add(new Broker(logger, authParams, preferredDomain: preferredDomain, pcaWrapper: pcaWrapper, promptHint: promptHint, platformUtils: platformUtils));
+                    }
+                    else
+                    {
```
kyle-rader-msft: These could be collapsed so the flows.add is called once in a top block that checks if win10Or11, or (is mac, & mac broker enabled), with the Log warning in a single else if block.

src/MSALWrapper/AuthFlow/Broker.cs
```diff
@@ -105,46 +112,143 @@ protected override async Task<TokenResult> GetTokenInnerAsync()
...
+
+            return await TaskExecutor.CompleteWithin(
+                this.logger,
+                this.interactiveAuthTimeout,
+                $"{this.Name} browser fallback",
+                (CancellationToken cancellationToken) => fallbackPca
+                    .WithPromptHint(this.promptHint)
+                    .GetTokenInteractiveAsync(this.scopes, account, cancellationToken),
+                this.errors).ConfigureAwait(false);
+        }
```
kyle-rader-msft: We shouldn't need to have a fallback method in this way, we already have the notion of AuthFlows - and the fall back to web shoudl already happen by the nature of broker coming first as an auth flow, and web coming after that.

kyle-rader-msft (changes requested):

ChristopherJMiller (dismissed):

src/AzureAuth/Commands/CommandAad.cs
dggsax: Agreed — IWA removal should be a follow-up. Filed as a mental TODO for now.

src/AzureAuth/Program.cs
dggsax: Windows WAM handles threading internally — it doesn't require main-thread dispatch like the macOS broker does. The `MacMainThreadScheduler` exists specifically because macOS's NSRunLoop-based message loop requires interactive broker calls on the main thread. On Windows, `AcquireTokenInteractive` works fine from any thread, so wrapping it in a scheduler would add complexity for no functional benefit.

src/MSALWrapper/AuthFlow/AuthFlowFactory.cs
dggsax: Done — collapsed into a single `if (IsWindows10Or11() || IsMacOSBrokerAvailable())` with one `flows.Add` call. The warning is in a single `else if (IsMacOS())` block. See latest commit.

src/MSALWrapper/AuthFlow/Broker.cs
dggsax: Good call — removed `FallbackToBrowserAuthAsync` entirely. The `AuthFlowExecutor` already handles this: if Broker returns with errors/null, the executor moves to the next flow (Web). Also removed the broad `catch (Exception) when (IsMacOS())` block — MSAL exceptions are caught by `AuthFlowBase.GetTokenAsync()` and flow through the executor naturally. See latest commit.

src/AzureAuth/Commands/CommandAad.cs
dggsax: Tracked in #454

src/MSALWrapper/AuthMode.cs
```diff
@@ -49,13 +49,20 @@ public enum AuthMode : short
...
+
         /// <summary>
         /// All auth modes.
         /// </summary>
-        All = Web | DeviceCode,
+        All = Broker | Web | DeviceCode,
 
         /// <summary>
-        /// Default auth mode.
+        /// Default auth mode. On macOS, broker is opt-in via --mode broker because
```
kyle-rader-msft, line 63: Depending on how this rollout goes, we might want to include `broker | web` as the default for Mac in the future.

kyle-rader-msft (approved):

kyle-rader-msft (dismissed):
Also please update the CHANGELOG.md

src/MSALWrapper/AuthMode.cs
dggsax: Agreed — tracked in #455. Once we have confidence from the rollout, we can flip the default.

siddharthpaudwal (approved):

ChristopherJMiller (approved):

PR #462, opened 2026-07-24
Add Linux brokered authentication via Microsoft Identity Broker

fintelia:
This implementation parallels the macOS one. I've tested it on Ubuntu 26.04.

**LLM disclosure:** I used copilot to generate this patch.


fintelia:
Note for anyone wanting to test this: you may have to set the `AZUREAUTH_MODE=broker` environment variable to get applications to use the broker authentication rather than defaulting to browser or device code.

siddharthpaudwal (approved):

kyle-rader-msft (approved):

